PR #19, opened 2020-06-20
OS Separation

### LINKED ISSUE #18: Separate the uC/OS III code from our code.
As we start to work with uC/OS III, I think that it would be a good idea to separate the OS from our application and driver code. This provides the following benefits:

- Multiple projects can use the same OS, without having to repeat the effort of porting the code.
- New projects will easily be able to easily import the OS.
- Any changes that we make to the OS-level code can be reflected in all projects with one commit, rather than requiring several projects to be updated individually.
- We separate the logical components of our applications from our OS, which will help with dependency management, and will make any future transitions easier to manage. This also means that changes can be made to the OS concurrently with changes to some of the applications, and no messy merges will be required.
- If any changes are made to uC/OS III at the Micrium level (unlikely but possible), we will be able to reflect those changes easier in our code base.

The only downside that I can think of: changing the OS code for one application would change it for another, which may not be warranted. This should be relatively minor though, since we use the same processors on most of our boards. Even if this becomes an issue, setting one project to track a different branch/commit of the OS repo would allow us to protect compatibility and would offer an easy route for future synchronization (i.e. a merge), should we decide that the same changes would be helpful in another project.

The next steps as far as I see it:

1. Separate the OS code into a new repository.
2. Remove the OS code from the Controls repository, and reconfigure the build to use the new OS repo. This will require also moving bsp_cpu to the new repository. (See the OSDemo branch if you don't know what this is).
3. Test.

If this is done correctly, I believe that it should not impact how code is developed for our team, unless that development changing the configuration internal to the OS. In that case, the only difference would be that the change takes place on a separate repository.

I'd like to hear what others think of this plan. I can try to carry this out myself, but I'll wait until there's been some conversation.

chaseblock:
This separates the OS from the rest of the code. It will now be tracked in two independent repositories.

This pull request currently includes a demonstration of the OS's functionality as a `main` function. If this should be kept blank for now, I can revert the main.c file without much effort.

The huge amount of files that changed is due to renaming a folder.

This also adds:

1. A new build target to flash the physical board.
2. Some UART stuff (mainly for the demo mentioned above)
3. A better `make clean` that doesn't error if the place is already clean.

Closes #18


chaseblock:
Looks good to me. Runs on the STM32F413 on the Nucleo board just fine.

One that that tripped me up while I was testing this was the `make flash` target. This relies on the object file that is built by `make stm32f413` (which makes sense), but this means that it will try to build this file using the default options before flashing, so if you're flashing a test, you need to do it via `make flash TEST=TEST_NAME`. Everyone else may already know this, but I thought I'd mention it just in case.

Anyways, @JimothyGreene feel free to hit the approve button whenever you're ready.

PR #20, opened 2020-06-25
ADC simulator

ErickCortez98:
Added pedals to the simulation GUI, a Python script that generates ADC data from the simulated pedals and writes them into a csv file, a C file that processes the raw ADC data from the csv file, converts it a percentage and provides the values when requested, and a test file, Test_BSP_ADC.c, that displays in the terminal the press percentage of the accelerator and brake.


JimothyGreene:
I'm may be about to piss you off, so just a warning.

My stupid brain forgot to separate ADC and the Pedals in my head. So we want to make the ADC module not specific to the pedals quite yet, since the Pedal stuff will be part of the Driver software. That being said, I want you to do less with this module. I'll leave comments on each file explaining what API I want and how I want it to operate. But if you have more questions, I'm on Slack.

Since some of this software does do pedal stuff, maybe keep the code somewhere safe for now, and I can have you both work on the Pedal drivers once we start working on them.

BSP/Inc/BSP_ADC.h
```diff
@@ -0,0 +1,13 @@
...
+#define __BSP_ADC_H
+
+//function that returns the percentage pressed of the accelerator pedal
+float ADC_Percent_Accel_Read();
+
+//function that returns the percentage pressed of the accelerator pedal
+float ADC_Percent_Brake_Read();
+
+// Initialize the ADC module
+void ADC_Init(void);
```
JimothyGreene, line 11: The functions we want to expose through this API are as follows:

`BSP_ADC_Init()`
`BSP_ADC_Accel_GetMillivoltage()`
`BSP_ADC_Brake_GetMillivoltage()`

The two bottom functions should return the value in millivolts for each respective ADC.

BSP/Simulator/Hardware/Pedals.py
```diff
@@ -0,0 +1,31 @@
+import os
+import csv
```
JimothyGreene, line 2: I would call this module `ADC.py` just to keep the naming parallel to the C stuff

BSP/Simulator/Src/BSP_ADC.c
```diff
@@ -0,0 +1,59 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include "BSP_ADC.h"
+#include "common.h"
+#include "config.h"
```
JimothyGreene, line 6: Most of the #includes should be in the header file

BSP/Simulator/Src/BSP_ADC.c
```diff
@@ -0,0 +1,59 @@
...
+//I'm assuming 3.3V corresponds to a 100% pressed pedal
+
+#define FILE_NAME "BSP/Simulator/Hardware/Data/Pedals.csv"
+
+int ADCvalueAccel, ADCValueBrake;
+float ADCPercentageAccel, ADCPercentageBrake;
+
+void ADC_Init(void) {
+    // Checking if file exists
+    if(access(FILE_NAME, F_OK) == -1){
```
JimothyGreene, line 25: I would change this to a != 0 just to be more inclusive. Honestly, I'm not sure what all values it will return, but that just makes it more inclusive.

JimothyGreene:
Also, for the documentation, take a look at the `simulator_switches` branch to see how I documented each function. The header file shows what it does, the source file shows how it does it, and the test file explains how it was tested.

BSP/Inc/BSP_ADC.h
JimothyGreene: With whatever new functions you want to add and expose through the API, make sure to use the ```BSP_ADC_``` prefix to the functions.

BSP/Inc/BSP_ADC.h
```diff
@@ -0,0 +1,82 @@
+#ifndef __BSP_ADC_H
+#define __BSP_ADC_H
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+#include "common.h"
+#include "config.h"
+
+enum ADCHardware {Accelerator, Brake};
```
JimothyGreene, line 10: I would make this a typedef and call it something like ```pedal_t``` to make it a type that we can reference later as well.

BSP/Inc/BSP_ADC.h
```diff
@@ -0,0 +1,82 @@
...
+ * @return  None
+ */ 
+void BSP_ADC_Init(void);
+
+/**
+ * @brief   Provides the number of active ADC channels 
+ * @param   None
+ * @return  numberOfADCValues, the number of ADC channels connected to the ADC providing raw ADC data
+ */ 
+int BSP_ADC_Number_Values();
```
JimothyGreene, line 24: We shouldn't need this function considering that our typedef will define explicitly which pieces of hardware are connected to the ADC and therefore how many channels we are using.

BSP/Inc/BSP_ADC.h
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * @brief   Provides pointer to array with the ADC values of csv file
+ * @param   None
+ * @return  rawADCValues, pointer to array that contains the ADC values in the array
+ */ 
+int* BSP_ADC_Read_Values();
```
JimothyGreene, line 31: I don't want to be able to get an array of unlabeled values. It's better if I explicitly say which channel I want to get the data for, and then receive that channel's data specifically (especially since we made that enum). So, this function shouldn't be needed either.

BSP/Inc/BSP_ADC.h
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * @brief   Provides the ADC value of the channel at the specified index
+ * @param   indexOfValue, index of value in the array of all ADC values obtained with read_ADC_Values
+ * @return  rawADCValue
+ */ 
+int BSP_ADC_Get_Value(int indexOfValue);
```
JimothyGreene, line 38: I'd like to keep this function. However, the return types should all be specific types of integers. In this case, since the ADC is 12 bits, we only need a 16-bit integer as the return type. 

Also, the parameter should now be the enum (that you should make a typedef as mentioned above) since we want to use that now to reference the specific devices.

BSP/Inc/BSP_ADC.h
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * @brief   Provides the ADC value in millivolts of the channel at the specified index
+ * @param   indexOfValue, index of value in the array of all ADC values obtained with read_ADC_Values
+ * @return  ADC value in millivolts
+ */ 
+float BSP_ADC_Get_Millivoltage(int indexOfValue);
```
JimothyGreene, line 45: I also want to keep this function. However, we want to avoid using floating point numbers. Since we are returning a millivoltage, you should just be able to return the integer portion of that number, since it will be precise to 3 decimal places as a voltage. 

Also, same as above for the paramater

BSP/Inc/BSP_ADC.h
```diff
@@ -0,0 +1,82 @@
...
+ * @return  Millivolts of Accelerator ADC
+ */ 
+float BSP_ADC_Accel_GetMillivoltage();
+
+/**
+ * @brief   Getter function that returns the Brake ADC value in mV
+ * @param   None
+ * @return  Millivolts of Brake ADC
+ */ 
+float BSP_ADC_Brake_GetMillivoltage();
```
JimothyGreene, line 80: We don't need any of these functions anymore. Now that we have the enum, we don't need specific functions for brake and accel. And we don't want the percentage functions anymore either.

BSP/Simulator/Src/BSP_ADC.c
```diff
@@ -0,0 +1,215 @@
...
+C file that reads the raw ADC data from the CSV file, 
+converts it back to a distance pressed (percentage for now), and stores both.
+We should be able to request both the raw ADC data,
+the distance (percentage for now) and millivolts from this library.
+ */
+
+//NOTE: ADC precision is 12 bits (4096), range 3.3V
+//I'm assuming 3.3V corresponds to a 100% pressed pedal
+
+#define FILE_NAME "BSP/Simulator/Hardware/Data/Pedals.csv"
```
JimothyGreene, line 13: Use the macro defined in bsp.h

Look at BSP_Switches.c to see how I used it.

BSP/Simulator/Src/BSP_ADC.c
```diff
@@ -0,0 +1,215 @@
...
+#define MAX_CHANNELS 10
```
JimothyGreene, line 14: I would define this in the header file

BSP/Simulator/Src/BSP_ADC.c
```diff
@@ -0,0 +1,215 @@
...
+ * @brief   Helper function which returns a 
+ *          percentage of the parameter when    
+ *          compared to the 4096 possible range 
+ *          of the ADC
+ * @param   raw ADC value
+ * @return  percentage of ADC value 
+ */ 
+float convert_ADC_to_Percent(int ADCvalue) {
+    return (((float)ADCvalue) / 4096) * 100;
+}
```
JimothyGreene, line 41: This isn't necessary anymore since we are scrapping the percentage functions

BSP/Simulator/Src/BSP_ADC.c
```diff
@@ -0,0 +1,215 @@
...
+
+/**
+ * @brief   Helper function which returns a 
+ *          milivoltage based on the raw ADC value
+ *          compared to the 3.3 volt range of the 
+ *          pedals' potentiometers
+ * @param   raw ADC value
+ * @return  milivolts 
+ */ 
+float convert_ADC_to_Voltage(int ADCvalue){
```
JimothyGreene, line 51: No floating point and more specific int type

BSP/Simulator/Src/BSP_ADC.c
```diff
@@ -0,0 +1,215 @@
...
+    //counting how many ADC Values the file has
+    for(char c = getc(filePointer); c!= EOF; c = getc(filePointer)){
+        if(c == '\n' || c == ','){
+            numberOfADCValues++;
+        }
+    }
+    //close the file 
+    fclose(filePointer);
+    return numberOfADCValues;
+}
```
JimothyGreene, line 77: See comments in header file

BSP/Simulator/Src/BSP_ADC.c
```diff
@@ -0,0 +1,215 @@
...
+    }
+    //reading the ADC values in the csv file and storing them in the rawADCValues array
+    for(int i = 0; i < numberOfValues; i++){
+        fscanf(filePointer,"%d,", &(rawADCValues[i]));
+    }
+    //closing the file
+    fclose(filePointer);
+    //returning array
+    return rawADCValues;
+}
```
JimothyGreene, line 106: See comments in header file

BSP/Simulator/Src/BSP_ADC.c
```diff
@@ -0,0 +1,215 @@
...
+
+/**
+ * @brief   Function that returns a specific
+ *          raw ADC value of a channel, the index of
+ *          the channel is provided as a parameter 
+ * @param   indexOfValue, index of value in the array of all ADC values obtained with read_ADC_Values
+ * @return  rawADCValue
+ */ 
+int BSP_ADC_Get_Value(int indexOfValue){
```
JimothyGreene, line 115: See comments in header

BSP/Simulator/Src/BSP_ADC.c
```diff
@@ -0,0 +1,215 @@
...
+}
+
+/**
+ * @brief   Function that returns a specific
+ *          raw ADC value in millivolts of a channel, the index of
+ *          the channel is provided as a parameter 
+ * @param   indexOfValue, index of value in the array of all ADC values obtained with read_ADC_Values
+ * @return  ADC value in millivolts
+ */ 
+float BSP_ADC_Get_Millivoltage(int indexOfValue){
```
JimothyGreene, line 133: See comments in header

BSP/Simulator/Src/BSP_ADC.c
```diff
@@ -0,0 +1,215 @@
...
+//** FOLLOWING FUNCTIONS ARE SPECIFIC FOR ACCELERATOR AND BRAKE ADC INFO **///
+
+/**
+ * @brief   Getter function that reads the Accelerator ADC,
+ *          which we know it's at index 0 of the csv file 
+ *          value, converts it to a percentage,  
+ *          and returns it
+ * @param   None
+ * @return  Percentage of Accelerator ADC
+ */
```
JimothyGreene, line 163: See header file

BSP/Simulator/Src/BSP_ADC.c
```diff
@@ -0,0 +1,215 @@
...
+ * @param   None
+ * @return  Percentage of Accelerator ADC
+ */ 
+float BSP_ADC_Accel_GetPercentage() {
+     //reading raw ADC values from csv file
+    int rawADCValue = BSP_ADC_Get_Value(Accelerator);
+    //convert ADC value to a percentage using a helper function and return value
+    return convert_ADC_to_Percent(rawADCValue);
+}
+
```
JimothyGreene, line 170: See header

BSP/Simulator/Src/BSP_ADC.c
```diff
@@ -0,0 +1,215 @@
...
+ * @param   None
+ * @return  Percentage of Brake ADC
+ */ 
+float BSP_ADC_Brake_GetPercentage() {
+    //reading raw ADC values from csv file
+    int rawADCValue = BSP_ADC_Get_Value(Brake);
+    //convert ADC value to a percentage using a helper function and return value
+    return convert_ADC_to_Percent(rawADCValue);
+}
+
```
JimothyGreene, line 185: See header

BSP/Simulator/Src/BSP_ADC.c
```diff
@@ -0,0 +1,215 @@
...
+ *          and returns it
+ * @param   None
+ * @return  Millivolts of Accelerator ADC
+ */ 
+float BSP_ADC_Accel_GetMillivoltage(){
+    //reading raw ADC values from csv file
+    int rawADCValue = BSP_ADC_Get_Value(Accelerator);
+    //convert ADC value to milivolts using a helper function and return value
+    return convert_ADC_to_Voltage(rawADCValue);
+}
```
JimothyGreene, line 199: See header

BSP/Simulator/Src/BSP_ADC.c
```diff
@@ -0,0 +1,215 @@
...
+ * @param   None
+ * @return  Millivolts of Brake ADC
+ */ 
+float BSP_ADC_Brake_GetMillivoltage(){
+    //reading raw ADC values from csv file
+    int rawADCValue = BSP_ADC_Get_Value(Brake);
+    //convert ADC value to milivolts using a helper function and return value
+    return convert_ADC_to_Voltage(rawADCValue);
+}
+
```
JimothyGreene, line 215: See header

Tests/Test_BSP_ADC.c
```diff
@@ -0,0 +1,53 @@
...
+ * 
+ * Uncomment the "individual tests for each function" to determine the output of individual functions
+ * of the ADC module
+ */ 
+
+#include "common.h"
+#include "config.h"
+#include "BSP_ADC.h"
+
+int main() {
```
JimothyGreene, line 17: Update this once you update the other files

BSP/Simulator/Src/BSP_ADC.c
```diff
@@ -0,0 +1,129 @@
...
+ *          raw ADC value of a channel, the index of
+ *          the channel is provided as a parameter 
+ * @param   indexOfValue, index of value in the array of all ADC values obtained with read_ADC_Values
+ * @return  rawADCValue
+ */ 
+int16_t BSP_ADC_Get_Value(pedal_t hardwareDevice){
+    //checking if indexOfValue is out of bounds
+    if(hardwareDevice > number_Values() - 1){
+        hardwareDevice = number_Values() - 1;
+    }
```
JimothyGreene, line 107: We don't want this to happen. If another function passes an incorrect argument, it should throw an error so that we know something is wrong.

BSP/Simulator/Src/BSP_ADC.c
```diff
@@ -0,0 +1,129 @@
...
+    //reading all values and storing them in an int array
+    int* rawADCValues = read_ADC_Values();
+    int16_t rawADCValue = *(rawADCValues + hardwareDevice);
```
JimothyGreene, line 110: You can just index it like this: rawADCValues[hardwareDevice]

BSP/Simulator/Src/BSP_ADC.c
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * @brief   Function that returns a specific
+ *          raw ADC value in millivolts of a channel, the index of
+ *          the channel is provided as a parameter 
+ * @param   indexOfValue, index of value in the array of all ADC values obtained with read_ADC_Values
+ * @return  ADC value in millivolts
+ */ 
+int BSP_ADC_Get_Millivoltage(pedal_t hardwareDevice){
+    //getting rawADCValue at the specified index
+    int rawADCValue = BSP_ADC_Get_Value(hardwareDevice);
```
JimothyGreene, line 124: int16_t

BSP/Simulator/Src/BSP_ADC.c
```diff
@@ -0,0 +1,129 @@
...
+    //counting how many ADC Values the file has
+    for(char c = getc(filePointer); c!= EOF; c = getc(filePointer)){
+        if(c == '\n' || c == ','){
+            numberOfADCValues++;
+        }
+    }
+    //close the file 
+    fclose(filePointer);
+    return numberOfADCValues;
+}
```
JimothyGreene, line 65: I don't think this function is necessary since we can read from a file just until the file ends. So we don't need to know in advance how many entries there are

BSP/Simulator/Src/BSP_ADC.c
```diff
@@ -0,0 +1,129 @@
...
+    //array that will store the ADC values read by the function, it has a maximum capacity of 10
+    static int rawADCValues[MAX_CHANNELS];
+    //number of values inside the csv file
+    int numberOfValues = number_Values();
+    //opening file in read mode
+    FILE *filePointer = fopen(FILE_NAME, "r");
+    //verifying the number of values in the csv file is less than 10
+    if(numberOfValues > MAX_CHANNELS){
+        numberOfValues = MAX_CHANNELS;
+    }
```
JimothyGreene, line 85: We don't really need to verify this. If there are more values, then just don't read them. There shouldn't be more than 10 anyway, and if there are, then that's a problem with the Python side of the simulator

BSP/Simulator/Src/BSP_ADC.c
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * @brief   Helper function that reads all of the ADC values 
+ *          from a file with raw values provided 
+ *          by the ADC and stores all values in local 
+ *          array of max capacity 10
+ * @param   none
+ * @return  rawADCValues, pointer to array that contains the ADC values in the array
+ */ 
+int* read_ADC_Values() {
```
JimothyGreene, line 75: I'd prefer to not have a static int* returned. Instead, you can have the function accept an int* as a parameter (a buffer array), and fill that array. That way, the variable stays local and impermanent.

BSP/Simulator/Src/BSP_ADC.c
```diff
@@ -0,0 +1,129 @@
...
+}
+
+/**
+ * @brief   Function that returns a specific
+ *          raw ADC value in millivolts of a channel, the index of
+ *          the channel is provided as a parameter 
+ * @param   indexOfValue, index of value in the array of all ADC values obtained with read_ADC_Values
+ * @return  ADC value in millivolts
+ */ 
+int BSP_ADC_Get_Millivoltage(pedal_t hardwareDevice){
```
JimothyGreene, line 122: Return a specific size integer here

BSP/Simulator/Hardware/Pedals.py
```diff
@@ -0,0 +1,31 @@
...
+    if accel_value == 4096: accel_value = 4095
+    if brake_value == 4096: brake_value = 4095
+    states = [accel_value, brake_value]
+    with open(file, "w") as csvfile:
+        csvreader = csv.writer(csvfile)
+        csvreader.writerow(states)
+
+
+if __name__ == '__main__':
+    set_pedals(0.7, 0.0)
```
JimothyGreene, line 31: This can be taken out now that we have the whole setup to test it with

BSP/Simulator/Src/BSP_ADC.c
```diff
@@ -0,0 +1,100 @@
...
+/**
+ * @brief   Helper function which returns a 
+ *          milivoltage based on the raw ADC value
+ *          compared to the 3.3 volt range of the 
+ *          pedals' potentiometers
+ * @param   raw ADC value
+ * @return  milivolts 
+ */ 
+int16_t convert_ADC_to_Voltage(int16_t ADCvalue){
+    return (((float)ADCvalue / 4096) * 3300);
```
chaseblock, line 41: Casting to a `float` is taking an unnecessary performance hit, since the function ultimately returns an int16_t. What would be better is something along the lines of what's shown below.
```suggestion
    return (uint16_t) (((uint32_t)ADCvalue * 3300) / 4096);
```
The cast to `uint32_t` ensures that the multiplication does not overflow (and since the registers are 32-bit, shouldn't be different in terms of performance), and moving the division to the last step avoid early truncation, without having to involve the FPU. The final cast to uint16_t is optional, since it should be implicitly done anyways, but it's never a bad idea to be explicit about things.

Also, I don't like magic numbers.

BSP/Simulator/Src/BSP_ADC.c
```diff
@@ -0,0 +1,100 @@
...
+ *          from a file with raw values provided 
+ *          by the ADC and stores all values in local 
+ *          array of max capacity 10
+ * @param   none
+ * @return  rawADCValues, pointer to array that contains the ADC values in the array
+ */ 
+void read_ADC_Values(int16_t *rawADCValues) {
+    //opening file in read mode
+    FILE *filePointer = fopen(FILE_NAME, "r");
+    //TODO: Correct the following function, read all values till the end of file
```
chaseblock, line 55: Has this TODO been done?

BSP/Simulator/Src/BSP_ADC.c
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * @brief   Helper function which returns a 
+ *          milivoltage based on the raw ADC value
+ *          compared to the 3.3 volt range of the 
+ *          pedals' potentiometers
+ * @param   raw ADC value
+ * @return  milivolts 
+ */ 
+int16_t convert_ADC_to_Voltage(int16_t ADCvalue){
```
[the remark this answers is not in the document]
JimothyGreene: Agreed

BSP/Simulator/Src/BSP_ADC.c
JimothyGreene: Yeah we want to avoid using floating point numbers as much as possible. Since this is in mV, we don't need to have a fraction portion of our number for the precision we want. 

As for the magic numbers, I would go ahead and put some #defines in the header file to explicitly say what these numbers are

BSP/Simulator/Src/BSP_ADC.c
JimothyGreene: I don't think so. To use `fscanf()` and read until end of file you should have a while loop like this:

```
while(fscanf(filePointer, "%d,", &(rawADCValues[i])) == 1) {
    // Do stuff
}
```
This will break when the scanner expects 1 conversion but doesn't see anything to convert

PR #22, opened 2020-07-03
Timer branch two

rafidh611:


BSP/Inc/BSP_Timer.h
```diff
@@ -0,0 +1,5 @@
+#include<stdio.h>
+#include<stdlib.h>
```
JimothyGreene, line 2: Spaces between the include and the library

BSP/Inc/BSP_Timer.h
```diff
@@ -0,0 +1,5 @@
...
+
+int update_Time(int row);
+int writeReload(int reload);
```
JimothyGreene, line 5: All functions should start with ```BSP_Timer_``` to make sure they are explicit. Also, make sure this API is common to both the simulator and the real board. 

We would want these functions probably:
```
BSP_Timer_Init()
BSP_Timer_Update()
```

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -0,0 +1,44 @@
+#include<stdio.h>
+#include<stdlib.h>
+#include "Timer.h"
+
+
+int update_Time(int row)
```
JimothyGreene, line 6: We want to use specific int types that use the minimum required space for whatever variable you're using.

JimothyGreene: Also, this function should update all timers. And its primary purpose should be to call whatever function the timer was initialized to call. The whole point is to have a timer that calls some function every so often (like an interrupt).

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -0,0 +1,44 @@
...
+ int subt = reload - current ;
+
+ return subt; 
+    
+    
+    
+
+}
+
+int writeReload(int reload)
```
JimothyGreene, line 39: This functionality should happen in the ```BSP_Timer_Init()``` function. That function should also set some static global variable(s) that are function pointers.

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -0,0 +1,44 @@
...
+{
+  FILE *CANdata = fopen("CAN.csv","w");
+  fprintf(CANdata,"0x%x,%d",reload,reload);
+  fclose(CANdata);
```
JimothyGreene, line 43: This shouldn't be CAN data anymore

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -27,7 +27,7 @@ int update_Time(int row)
 
     }
  
- int subt = reload - current ;
+ int8_t subt = reload - current ;
```
JimothyGreene, line 30: This function shouldn't be returning anything. Its function should be to write the new reload value to the CSV file like a real timer would update the CURRENT register. It should also call upon whatever function it was initialized to if the timer reaches 0.

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -36,9 +36,9 @@ int update_Time(int row)
 
 }
 
-int writeReload(int reload)
+int8_t BSP_Timer_Init(int8_t reload)
```
JimothyGreene, line 39: This function shouldn't return anything. Also, it should set some sort of static global variable in this file that is the function pointer to whatever function needs to be called on interrupt.

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -36,9 +36,9 @@ int update_Time(int row)
...
 {
-  FILE *CANdata = fopen("CAN.csv","w");
-  fprintf(CANdata,"0x%x,%d",reload,reload);
-  fclose(CANdata);
+  FILE *TIMERdata = fopen("Timer.csv","w");
+  fprintf(TIMERdata,"0x%x",reload);
```
JimothyGreene, line 42: When writing to the file, it should write 2 numbers (the CURRENT value and the RELOAD value)

JimothyGreene:
Closing this until both sides of the Timer sim are together.

PR #25, opened 2020-07-23
Timer simulation

rafidh611:
This is the code we have so far. We still have to run some tests.


BSP/Simulator/Hardware/simulate.py
```diff
@@ -15,6 +16,35 @@ def update_buttons():
...
+		if currentValue > 0: # Systick stops at 0
+			with open(simulatorFile, 'w') as csvfile:
+				writer = csv.writer(csvfile, delimiter=',')
+
+				writer.writerow([currentValue - 1, reloadValue])
+
+	except FileNotFoundError:
+		print("File " + simulatorFile + " not found")
+	
+	window.after(1, update_timer)
```
JimothyGreene, line 47: Most of this functionality should be implemented in a separate Timer.py file. Take a look at the Switches.py and Pedals.py files to see the formatting for that. For the exceptions, we want to create the file if they are not already created (again, see the other modules for examples). In terms of the Blank CSV situation, if there is a blank file (meaning nothing has been initialized), just set the reloadValue and currentValue to some maximum value (like 0xFFFFFF)

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -0,0 +1,55 @@
+#include<stdio.h>
+#include<stdlib.h>
```
JimothyGreene, line 2: These includes are in `BSP_Timer.h` so you don't need to include them here too

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -0,0 +1,55 @@
+#include<stdio.h>
+#include<stdlib.h>
+#include "BSP_Timer.h"
+
+
+// function keeps updating time and will interrupt at the right time to go to specefic functions
```
JimothyGreene, line 6: See the Style Guide for the comment structure here and on other functions

BSP/Inc/BSP_Timer.h
```diff
@@ -0,0 +1,7 @@
+#include<stdio.h>
+#include<stdlib.h>
+
+void (*function1);    //function pointers
+void (*function2);  
+void BSP_Update_Time(int8_t row);
+void BSP_Timer_Init(int8_t reload, void *func1, void *func2);
```
JimothyGreene, line 7: See the Style Guide for the comment structure in header files with function prototypes

BSP/Simulator/Hardware/simulate.py
```diff
@@ -15,6 +16,35 @@ def update_buttons():
...
+
+			except IndexError: # If the csv is blank. Not sure what to do in this case so I just print for debugging.
+				print("Blank CSV")
+
+
+		if currentValue > 0: # Systick stops at 0
+			with open(simulatorFile, 'w') as csvfile:
+				writer = csv.writer(csvfile, delimiter=',')
+
+				writer.writerow([currentValue - 1, reloadValue])
```
JimothyGreene, line 42: When currentValue goes below 0, make sure that you rewrite the reloadValue into the currentValue register

Tests/Test_BSP_Timer.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+
+    BSP_Timer_Init(1000,functOnepoiner,functTwopoiner);
+
+    while(1){
+      count++;
+      BSP_Update_Time(count);
+    }
+    
+}
```
JimothyGreene, line 33: We don't want to be using .cpp files since that requires a different compiler

BSP/Simulator/Hardware/simulate.py
```diff
@@ -16,35 +15,6 @@ def update_buttons():
             button.config(relief=tk.RAISED)
     window.after(1, update_buttons)
 
-def update_timer():
```
JimothyGreene: I would keep this function here, and have it call the function from the Timer.py file and also window.after()

BSP/Simulator/Hardware/Timer.py
```diff
@@ -0,0 +1,37 @@
+import csv
+import os
+
+# Path of file
+file = "Data/Timer.csv"
```
JimothyGreene, line 5: Make this the full file path like in Switches.py since the script will be executed from the root directory using a symbolic link.

BSP/Simulator/Hardware/Timer.py
```diff
@@ -0,0 +1,37 @@
...
+		reader = csv.reader(csvfile, delimiter=',')
+		
+		try:
+			for row in reader: # Should be 2 rows
+				current_values.append(int(row[0]))
+				reload_values.append(int(row[1]))
+
+		except IndexError: # If the csv is blank.
+			current_values.extend([0x100000, 0x1000000]) # 0xFFFFFF + 1 since we are decrementing below
+			reload_values.extend([0xFFFFFF, 0xFFFFFF])
```
JimothyGreene, line 28: This try/except isn't actually catching anything. If the file is blank, then the `reader` object has now `row`, so the loop will not even execute. The IndexError is happening in the next block when you try to write to the file. That's where you should catch the exception.

BSP/Simulator/Hardware/Timer.py
```diff
@@ -18,20 +18,20 @@ def update():
...
-				writer.writerow([reload_values[timer], reload_values[timer]])
+
+			try:
+				if current_values[timer] > 0: # If timer is still running
+					writer.writerow([current_values[timer] - 1, reload_values[timer]])
+				else: # When timer hits 0, reload
+					writer.writerow([reload_values[timer], reload_values[timer]])
+
+			except IndexError: # If a row is blank
+				writer.writerow([0xFFFFFF, 0xFFFFFF])
```
JimothyGreene, line 37: This is working great now

BSP/Simulator/Hardware/Timer.py
```diff
@@ -2,7 +2,7 @@
 import os
 
 # Path of file
-file = "Data/Timer.csv"
+file = "BSP/Simulator/Hardware/Data/Timer.csv"
```
JimothyGreene, line 5: Perfect

BSP/Inc/BSP_Timer.h
```diff
@@ -0,0 +1,8 @@
+#include<stdio.h>
+
+#include<stdlib.h>
+
+
+void BSP_Timer_Update();
+void BSP_Timer_Init(int Timer1Reload,int Timer2Reload, void *func1, void *func2);
+
```
JimothyGreene, line 8: Documentation, header guard. Take a look at BSP_Switches.h for format

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -0,0 +1,73 @@
+#include "BSP_Timer.h"
+
+static void (*function1);
+static void (*function2);
```
JimothyGreene, line 4: Give better names. Maybe something like timer1Handler and timer2Handler

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -0,0 +1,73 @@
...
+ 
+
+     if(current1==0)
+      {
+          (*function1);//sendDashboard func called
+      }
+      if(current2==0)
+      {
+         (*function2); //sendMC func called
+      }
```
JimothyGreene, line 50: This doesn't call the function. It just dereferences the pointer

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -0,0 +1,73 @@
...
+ * @param   none
+
+ * @return  none
+
+ */  
+
+void BSP_Timer_Update()
+{
+    FILE *file = fopen("Timer.csv", "r");
+    char arr[300];
```
JimothyGreene, line 19: Why 300?

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -0,0 +1,73 @@
...
+
+    int current2;
+
+
+    while(fgets(arr,sizeof(arr),file))
+    {
+        counter++;
+        if(counter == row)
+        {
+            fscanf(file,"%x , %x",&current1,&current2);
```
JimothyGreene, line 33: The format is as follows:

2 lines in the CSV file

CURRENT1, RELOAD1
CURRENT2, RELOAD2

You won't need a row variable. Just read all rows (2) and store the values properly

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -0,0 +1,73 @@
...
+ * @param   2 reload values for two timers, Two function pointers 
+
+ * @return  none
+
+ */  
+
+void BSP_Timer_Init(int Timer1Reload,int Timer2Reload,void *func1,void *func2)
+{
+  FILE *TimerData = fopen("Timer.csv","w");
+  fprintf(TimerData,"0x%x , 0x%x",Timer1Reload,Timer2Reload);
```
JimothyGreene, line 69: Each timer is its own line. You should write the RELOAD values to the CURRENT register to initialize.

Tests/Test_BSP_Timer.c
```diff
@@ -0,0 +1,38 @@
...
+void functionTwo()
+{
+  printf("Second Function worked");
+}
+
+
+int main()
+{
+    void (*functOnepointer)= &functionOne;
+    void (*functTwopointer)= &functionTwo;
```
JimothyGreene, line 28: You don't have to make these variables. You can just pass `functionOne` and `functionTwo` to the Init function

BSP/Inc/BSP_Timer.h
```diff
@@ -0,0 +1,22 @@
...
+/** 
+ * @brief   Updates the time by reading the CSV file and calling relevant functions at the right time 
+ * @param   none
+ * @return  none
+ */  
+void BSP_Timer_Update();
+
+/** 
+ * @brief   Intializes the Timer
+ * @param   2 reload values for two timers, Two function pointers
```
JimothyGreene, line 18: Good, but I would prefer it if you just split these up. The tools in most IDEs look for it split up. You can look at the simulator_can branch for an example of how that's done with a lot of parameters.

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -0,0 +1,76 @@
+#include "BSP_Timer.h"
+
+static void (*Timer1Handler)(int);
+static void (*Timer2Handler)(int);
```
JimothyGreene, line 4: Could you pass void instead of int? If so, that would be preferable.

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -0,0 +1,76 @@
...
+    int reload;
+
+    int i=0;
+
+    int index=0;
+
+
+    while(i<2)
+    {
+        fscanf(file,"%d , %d",&current,&reload);
```
JimothyGreene, line 32: If you're storing these values into `arr`, you don't need to have the two intermediate variables here. Also, try to use a more descriptive name than `arr`

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -0,0 +1,76 @@
...
+        arr[index]=current;
+        arr[index+1]=reload;
+        index=+2;
```
JimothyGreene, line 35: Is this supposed to be += 2?

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -0,0 +1,76 @@
...
+ * @param   none
+
+ * @return  none
+
+ */  
+
+void BSP_Timer_Update()
+{
+   FILE *file = fopen("Timer.csv", "r");
+    int arr[4];
```
JimothyGreene, line 19: A better approach might be to make a 2d array of size 2x2. Just makes more logical sense to me.

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -0,0 +1,76 @@
...
+
+ * @return  none
+
+ */  
+
+void BSP_Timer_Init(int Timer1Reload,int Timer2Reload,void *func1,void *func2)
+{
+  FILE *TimerData = fopen("Timer.csv","w");
+  fprintf(TimerData,"%d , %d",Timer1Reload,Timer1Reload);
+  fprintf(TimerData,"\n%d , %d",Timer2Reload,Timer2Reload);
```
JimothyGreene, line 72: You can omit spaces between the commas when writing to the CSV file. Just to use as little space as possible.

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -0,0 +1,76 @@
...
+
+ * @param   2 reload values for two timers, Two function pointers 
+
+ * @return  none
+
+ */  
+
+void BSP_Timer_Init(int Timer1Reload,int Timer2Reload,void *func1,void *func2)
+{
+  FILE *TimerData = fopen("Timer.csv","w");
```
JimothyGreene, line 70: Add file locks when opening. You can take a look at the simulator_can branch or the main simulator branch in BSP_Switches.c, BSP_UART.c, or BSP_ADC.c

BSP/Inc/BSP_Timer.h
```diff
@@ -0,0 +1,22 @@
+/**
+ * Header file for the library to interact
+ * with the timers
+ */
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/file.h>
```
JimothyGreene, line 7: Here you can just include "common.h" and "bsp.h"

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -0,0 +1,62 @@
...
+    int timeCurrent[2];
+    int current;
+    int reload;
+    int i=0;
+    int index=0;
+    
+    while(i<2)
+    {
+        fscanf(file,"%d , %d",&current,&reload);
+        timeCurrent[index]=current;
```
JimothyGreene, line 26: Instead of this, you could get rid of `current` altogether and simply pass (timeCurrent+index) where you have &current

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -0,0 +1,62 @@
...
+    flock(fno, LOCK_EX);
+    int timeCurrent[2];
+    int current;
+    int reload;
+    int i=0;
+    int index=0;
+    
+    while(i<2)
+    {
+        fscanf(file,"%d , %d",&current,&reload);
```
JimothyGreene, line 25: This approach could be made cleaner if you did something like was done in BSP_CAN.c in the main simulator branch. Take a look at that and see if that would work here. It's just a bit messy with all the counters and such.

BSP/Simulator/Src/BSP_Timer.c
rafidh611: When I try to pass the array with the index value into the fscanf  function, it gives me an error when the code is compiling. It says that the right format is not being passed into the function. It works so far when I pass the address of a normal int variable

BSP/Simulator/Src/BSP_Timer.c
rafidh611: In BSP_CAN.c, when they use the scan function, they are only passing in address of a normal int variable. My scan function also only works if I pass in only the address of an int variable.

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -0,0 +1,60 @@
...
+    int i=0;
+    int index=0;
+    
+    while(i<2)
+    {
+        fscanf(file,"%d , %d",timeCurrent+index,&reload);
+        index++;
+        i++;
+        
+    }
```
JimothyGreene, line 28: This can be turned into a for loop since it uses a counter variable

BSP/Simulator/Src/BSP_Timer.c
JimothyGreene: Still this. Just use the same format you used in the other function

BSP/Simulator/Src/BSP_Timer.c
JimothyGreene: Have you tried that? So it would be defined like:

`static void (*timer1Handler)(void)`
and the same for the second handler

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -51,10 +48,13 @@ void BSP_Timer_Update()
 
 void BSP_Timer_Init(int Timer1Reload,int Timer2Reload,void *func1,void *func2)
 {
-  FILE *TimerData = fopen("Timer.csv","w");
-  fprintf(TimerData,"%d,%d",Timer1Reload,Timer1Reload);
-  fprintf(TimerData,"\n%d,%d",Timer2Reload,Timer2Reload);
-  fclose(TimerData);
+  FILE *file = fopen("Timer.csv","w");
```
JimothyGreene, line 51: The file path here is incorrect. See how the other modules did it by using the macro defined in bsp.h

This incorrect path is causing this to not work since the Python file is updating the correct file and this one isn't.

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -51,10 +48,13 @@ void BSP_Timer_Update()
 
 void BSP_Timer_Init(int Timer1Reload,int Timer2Reload,void *func1,void *func2)
```
JimothyGreene, line 49: Now that I'm looking at this again, I would recommend using an enum approach to this. Very similar to how BSP_CAN does it (it also has 2 lines like this has 2 timers). So the Init function would take the enum of which timer, the reload for that timer, and the handler function. It would require you to call the Init function twice to initialize both timers, but that also allows us to initialize each timer in different places so they aren't tied to one another.

BSP/Inc/BSP_Timer.h
```diff
@@ -0,0 +1,24 @@
+/**
+ * Header file for the library to interact
+ * with the timers
+ */
+#include "common.h"
+#include <bsp.h>
+
+typedef enum {Time1=0, Time2=1} Time_t;
```
JimothyGreene, line 8: See style guide for the naming conventions/capitalization of enums

BSP/Inc/BSP_Timer.h
JimothyGreene: Still this

BSP/Inc/BSP_Timer.h
```diff
@@ -0,0 +1,24 @@
...
+ * @return  none
+ */  
+void BSP_Timer_Update();
+
+/** 
+ * @brief   Intializes the Timer
+ * @param   2 reload values for two timers, Two function pointers 
+ * @return  none
+ */  
+void BSP_Timer_Init(int TimerReload,void *func1, Timer_t time);
```
JimothyGreene, line 23: Does this compile? This uses `Timer_t` and the enum is typedef as `Time_t`

BSP/Simulator/Src/BSP_Timer.c
```diff
@@ -0,0 +1,78 @@
...
+
+void BSP_Timer_Init(int TimerReload,void *func,Time_t time)
+{
+  if(time==Time1)
+  {
+    FILE *file = fopen(FILE_NAME,"w");
+    int fno = fileno(file);
+    flock(fno, LOCK_EX);
+    fprintf(file,"%d,%d",TimerReload,TimerReload);
+    fprintf(file,"\n%d,%d",TempValue,TempValue);
```
JimothyGreene, line 59: This will overwrite the value that may have been already initialized here. You should instead first read the values that are currently written to the file, write new values to the timer that was specified, and write the value that was read to the other timer. See BSP_CAN for that approach.

Tests/Test_BSP_Timer.c
```diff
@@ -0,0 +1,34 @@
...
+{
+  printf("Second Function worked\n");
+}
+
+
+int main()
+{
+
+    BSP_Timer_Init(1000,functionOne,0);
+    BSP_Timer_Init(100,functionTwo,1);
```
JimothyGreene, line 29: The point of creating an enum is to be able to call functions using the names given to them, not the number value.

BSP/Inc/BSP_Timer.h
JimothyGreene: Also, you don't need to specify every value. Once you specify that the first value is 0, the rest will increment from there.

BSP/Simulator/Src/BSP_Timer.c
rafidh611: Since this will happen, I created a variable call TempValue that stores the value of the previous Reload value and prints out the value again on the CSV so nothing is lost. It worked  when I ran it.

JimothyGreene: I see what you've done now after looking closer. The only problem with that is that this approach assumes that the 2 timers are always going to be initialized immediately one after the other. If one is initialized, more code is run, then the other is initialized, the CURRENT register of the first initialized timer will be changed back to its original value, which will mess with the interrupt frequency during initialization. I would prefer if you read the values and rewrote the line that is not being changed.

PR #29, opened 2020-07-27
CAN Simulation

### LINKED ISSUE #2: CAN Simulation
Simulate sending and receiving messages across the CANBus

Requirements:
1. Python script that can read from the CSV file and return the CAN_ID and CAN_MESSAGE
2. Python simulation engine addition to display the current/last sent message & ID
3. C file for both TX and RX of CAN messages. Since we will be using 2 CAN buses, you will need to create functions for both CAN1 and CAN2 (since those are the two CAN pin pairs we are using on the new board design).

JimothyGreene:
Simulates both CAN buses connected to the Controls system. CAN1 is the CAN line connected to the rest of the car's systems. This is simulated simply by creating a line in a file where messages can be sent by the Controls system and will be displayed on the GUI. As of right now, this does not allow us to send messages from the GUI (to mimic sending messages from other systems), but that is not incredibly pertinent to our system's operations at this point. That can be added in the future. CAN2 is the CAN line connected to the motor controller. This is simulated in a similar fashion in addition to the logic for the motor controller. The MotorController.py script expects a certain message ID every 250ms, and will go into neutral if it does not receive that message. It also simulates a linear acceleration of 0.5 m/s per 250ms (in either direction depending on the relationship between the desired velocity and current velocity. Closes #2


BSP/Inc/BSP_CAN.h
```diff
@@ -0,0 +1,44 @@
+/**
+ * Header file for the library to interact
+ * with both CAN lines in the car
+ */
+
+#ifndef __BSP_CAN_H
+#define __BSP_CAN_H
+
+#include "common.h"
+#include <bsp.h>
```
chaseblock, line 10: For consistency, it might be better to use `#include "bsp.h"` to distinguish our code from system headers.

JimothyGreene: I'll go back through and change that in all the parts of the simulator once everything is merged in

PR #30, opened 2020-07-29
Contactors Simulation

### LINKED ISSUE #17: Contactor Simulation
Simulate the contactors for the motor and the array. The Controls board is responsible for connecting these systems to the rest of the car.

1. C file that initializes the contactors (both motor and array). This file should also be able to set the state of each contactor and read the state of each contactor.

2. Python file that reads the state of each contactor.

3. Addition to the simulation GUI that displays the state of each contactor and updates periodically like the button states and such.

sugamarora99:
Created C file that can read the state of both contactors or set their states. Created Python module that reads contactor states and added this output to the GUI. Closes #17

### DIFF
--- a/BSP/Inc/BSP_Contactors.h
+++ b/BSP/Inc/BSP_Contactors.h
@@ -5,8 +5,35 @@
 #include "config.h"
 #include <bsp.h>
 
+typedef enum {MOTOR=0, ARRAY} ContactorType_t;
+/**
+ * @brief   Confirms that the CSV file
+ *          has been created and creates
+ *          one if not
+ * @param   None
+ * @return  None
+ */ 
 void BSP_Contactors_Init(void);
-void read_Contactor_States(int16_t *contactorStates);
-void BSP_Contactor_Set(int16_t motorContactorState, int16_t arrayContactorState);
+
+/**
+ * @brief   Reads contactor states from a file 
+ *          and returns the current state of 
+ *          a specified contactor
+ * @param   contactor specifies the contactor for 
+ *          which the user would like to know its state (MOTOR/ARRAY)
+ * @return  The contactor's state (ON/OFF)
+ */ 
+state_t BSP_Contactors_Get(ContactorType_t contactor);
+
+/**
+ * @brief   Sets the state of a specified contactor
+ *          by updating csv file
+ * @param   contactor specifies the contactor for 
+ *          which the user would like to set its state (MOTOR/ARRAY)
+ * @param   state specifies the state that 
+ *          the user would like to set (ON/OFF)
+ * @return  The contactor's state (ON/OFF)
+ */ 
+void BSP_Contactor_Set(ContactorType_t contactor, state_t state);
 
 #endif
--- a/BSP/Inc/BSP_Contactors.h
+++ b/BSP/Inc/BSP_Contactors.h
@@ -1,5 +1,5 @@
-#ifndef __BSP_SWITCHES_H
-#define __BSP_SWITCHES_H
+#ifndef __BSP_CONTACTORS_H
+#define __BSP_CONTACTORS_H
 
 #include "common.h"
 #include "config.h"
--- a/BSP/Inc/BSP_Contactors.h
+++ b/BSP/Inc/BSP_Contactors.h
@@ -34,6 +34,6 @@ state_t BSP_Contactors_Get(ContactorType_t contactor);
  *          the user would like to set (ON/OFF)
  * @return  The contactor's state (ON/OFF)
  */ 
-void BSP_Contactor_Set(ContactorType_t contactor, state_t state);
+void BSP_Contactors_Set(ContactorType_t contactor, state_t state);
 
 #endif
--- a/BSP/Inc/BSP_Contactors.h
+++ b/BSP/Inc/BSP_Contactors.h
@@ -6,7 +6,7 @@
 #include <bsp.h>
 
 void BSP_Contactors_Init(void);
-static void read_Contactor_States(int16_t *contactorStates);
-static void BSP_Contactor_Set(int16_t motorContactorState, int16_t arrayContactorState);
+void read_Contactor_States(int16_t *contactorStates);
+void BSP_Contactor_Set(int16_t motorContactorState, int16_t arrayContactorState);
 
 #endif
--- a/BSP/Inc/BSP_Contactors.h
+++ b/BSP/Inc/BSP_Contactors.h
@@ -1,0 +1,12 @@
+#ifndef __BSP_SWITCHES_H
+#define __BSP_SWITCHES_H
+
+#include "common.h"
+#include "config.h"
+#include <bsp.h>
+
+void BSP_Contactors_Init(void);
+static void read_Contactor_States(int16_t *contactorStates);
+static void BSP_Contactor_Set(int16_t motorContactorState, int16_t arrayContactorState);
+
+#endif
--- a/BSP/Simulator/Hardware/Contactor.py
+++ b/BSP/Simulator/Hardware/Contactor.py
@@ -1,0 +1,20 @@
+import csv
+import os
+import fcntl
+
+
+file = "BSP/Simulator/Hardware/Data/Contactor.csv"
+
+#Reads file and returns 1 or 0, 1 is on, 0 is off
+def Read_Contactor():
+    returnNothing = ["", ""]
+    if os.path.isfile(file) == True:    #only execute the following code if the .csv file has been created
+        os.makedirs(os.path.dirname(file), exist_ok=True)
+        with open(file, 'r') as csvfile:
+            fcntl.flock(csvfile.fileno(), fcntl.LOCK_EX)    # Lock file before reading
+            csvreader = csv.reader(csvfile) 
+            Contactor_State = next(csvreader)
+            fcntl.flock(csvfile.fileno(), fcntl.LOCK_UN)    # Unlock file after reading
+            return Contactor_State
+    else:
+        return returnNothing #csv file hasn't been made, there is no data to read
--- a/BSP/Simulator/Hardware/Contactor.py
+++ b/BSP/Simulator/Hardware/Contactor.py
@@ -3,7 +3,7 @@ import os
 import fcntl
 
 
-file = "BSP/Simulator/Hardware/Data/Contactor.csv"
+file = "BSP/Simulator/Hardware/Data/Contactors.csv"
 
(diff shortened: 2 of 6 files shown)


BSP/Inc/BSP_Contactors.h
```diff
@@ -0,0 +1,12 @@
+#ifndef __BSP_CONTACTORS_H
+#define __BSP_CONTACTORS_H
+
+#include "common.h"
+#include "config.h"
+#include <bsp.h>
+
+void BSP_Contactors_Init(void);
+void read_Contactor_States(int16_t *contactorStates);
```
JimothyGreene, line 9: Be consistent with the naming. So I would call this BSP_Contactors_Get() since it's standard to have setter and getter functions.

BSP/Inc/BSP_Contactors.h
```diff
@@ -0,0 +1,12 @@
...
+void BSP_Contactor_Set(int16_t motorContactorState, int16_t arrayContactorState);
```
JimothyGreene, line 10: 2 things.

1. Be consistent with the plurality of the naming (just for ease of use in the future). So call this BSP_Contactors_Set 
2. So that we don't have to pass both states every time, I would have it take an enum as a parameter (similar to how it's done in BSP_CAN with the 2 different CAN lines). That way you can have it take the enum and a state and it will set the appropriate contactor state and not touch the other one.

BSP/Inc/BSP_Contactors.h
JimothyGreene: Also, as the comment below this says, I would include the enum as a parameter here also and make the function actually return the state of that contactor instead of getting both into a buffer array.

BSP/Inc/BSP_Contactors.h
```diff
@@ -0,0 +1,12 @@
...
+
```
JimothyGreene, line 11: Overall, this file needs documentation (see the style guide or other modules for examples). Also, the parameter types don't need to be int16_t since a contactor state is only ever a 0 or 1. In fact, I created an enum in config.h that defines an ON and OFF state for the switches, and that can be used here as well. It just makes it easier to work with.

BSP/Simulator/Hardware/Contactor.py
```diff
@@ -0,0 +1,27 @@
+import csv
+import os
+import fcntl
+
+
+file = "BSP/Simulator/Hardware/Data/Contactors.csv"
+
+#Reads file and returns 1 or 0, 1 is on, 0 is off
+def Read_Contactor():
```
JimothyGreene, line 9: Naming convention. Each of the modules has been calling their respective reading functions just `read()`. Easier to call in the simulator.py file

BSP/Simulator/Hardware/Contactor.py
```diff
@@ -0,0 +1,27 @@
+import csv
+import os
+import fcntl
+
+
+file = "BSP/Simulator/Hardware/Data/Contactors.csv"
+
+#Reads file and returns 1 or 0, 1 is on, 0 is off
```
JimothyGreene, line 8: Follow the Docstring documentation style. See other modules for examples.

BSP/Simulator/Hardware/Contactor.py
```diff
@@ -0,0 +1,27 @@
...
+def Read_Contactor():
+    returnNothing = ["", ""]
+    if os.path.isfile(file) == True:    #only execute the following code if the .csv file has been created
```
JimothyGreene, line 11: This is unnecessary. Take a look at how the other modules do it. They run line 12 every time (which creates the folder if it doesn't exist) and checks if the filepath is valid and opens the file in write mode otherwise (this takes away the need for the overarching if/else block with repetitive code.

BSP/Simulator/Hardware/Contactor.py
```diff
@@ -0,0 +1,27 @@
...
+        os.makedirs(os.path.dirname(file), exist_ok=True)
+        with open(file, 'r') as csvfile:
+            fcntl.flock(csvfile.fileno(), fcntl.LOCK_EX)    # Lock file before reading
+            try:    #the following code is to prevent simulator from crashing if the csv is blank
+                csvreader = csv.reader(csvfile)
+                Contactor_State = next(csvreader)
```
JimothyGreene, line 17: Naming convention for Python variables are lowercase and snake case

BSP/Simulator/Hardware/simulate.py
```diff
@@ -3,16 +3,20 @@
 from functools import partial
 
 import Switches
+import Contactor
 
 
 def update_buttons():
     """Periodically update the display state of buttons"""
     switches = Switches.read()
+    contactorState = Contactor.Read_Contactor()
```
JimothyGreene, line 12: This function is for updating the buttons. You should make your own update function so that it's more modular

BSP/Simulator/Hardware/simulate.py
```diff
@@ -3,16 +3,20 @@
...
     for i, button in enumerate(buttons):
         if switches & 1<<i:
             button.config(relief=tk.SUNKEN)
         else:
             button.config(relief=tk.RAISED)
+    contactor_status = tk.Label(master=window, text="Motor Contactor:" + contactorState[0]+"\nArray Contactor:" + contactorState[1], font=("Helvetica", 8))
+    contactor_status.place(x=325, y= 10)
```
JimothyGreene, line 19: This is repeated code

BSP/Simulator/Hardware/simulate.py
```diff
@@ -45,6 +49,11 @@ def update_buttons():
     button.grid(row=i//max(button_frame_rows), column=i%len(button_frame_columns), sticky='nsew')
     buttons.append(button)
 
+#Creates text that displays Contactor states
+contactor_status = tk.Label(master=window, text="Motor Contactor: \nArray Contactor:", font=("Helvetica", 8))
+contactor_status.place(x=325, y= 10)
+
+
```
JimothyGreene, line 56: Create the frame in the same way that the other modules did it. Check out how it was done for CAN. That'll be most similar to this.

BSP/Simulator/Src/BSP_Contactors.c
```diff
@@ -0,0 +1,48 @@
+#include "BSP_Contactors.h"
+#include <stdio.h>
```
JimothyGreene, line 2: This was already included in common.h

BSP/Simulator/Src/BSP_Contactors.c
```diff
@@ -0,0 +1,48 @@
...
+    int fno = fileno(filePointer);
+    //flock(fno, LOCK_EX);
+    for(int i = 0; i < 10; i++){
+        // If the file doesn't contain any more values, stop early
+        if (fscanf(filePointer,"%hd,", &(contactorStates[i])) <= 0) break;
+    }
+    // Closing the file
+    flock(fno, LOCK_UN);
+    fclose(filePointer);
+}
```
JimothyGreene, line 42: See comments in header file

BSP/Simulator/Src/BSP_Contactors.c
```diff
@@ -0,0 +1,48 @@
...
+    // Closing the file
+    flock(fno, LOCK_UN);
+    fclose(filePointer);
+} 
+
+void BSP_Contactor_Set(int16_t motorContactorState, int16_t arrayContactorState) {
+    FILE *filePointer = fopen(FILE_NAME, "w");
+    fprintf(filePointer, "%d, %d", motorContactorState, arrayContactorState);
+    fclose(filePointer);
+}
```
JimothyGreene, line 48: See comments in header file

Tests/Test_BSP_Contactors.c
```diff
@@ -0,0 +1,12 @@
...
+#include "config.h"
+#include <bsp.h>
+
+int main() {
+    BSP_Contactors_Init();
+    BSP_Contactor_Set(1,0);
+    int16_t states[5];
+    read_Contactor_States(states);
+    printf("%d%d", states[0], states[1]);
+    exit(0);
```
JimothyGreene, line 11: I would prefer it if this test was built to interact with the Python side. Something in an infinite loop that continuously asks for you to change the states or something. Based on user input.

Tests/Test_BSP_Contactors.c
```diff
@@ -0,0 +1,12 @@
+#include "common.h"
```
JimothyGreene, line 1: After making the changes to this file, leave documentation on how this test should be used.

JimothyGreene:
Python stuff looking better. Take a look at the comments on the C stuff and make sure to respond to the comments when you fix one of the issues so I can resolve them.

BSP/Inc/BSP_Contactors.h
cnflood: Changed to BSP_Contactors_Get()

BSP/Inc/BSP_Contactors.h
cnflood: Changed function name to BSP_Contactors_Set().
Its inputs are now enums for both contactor type and state.

BSP/Inc/BSP_Contactors.h
cnflood: Added documentation for all functions in BSP_Contactors.h. Implemented enum from config.h into the set function.

BSP/Simulator/Src/BSP_Contactors.c
cnflood: Removed for being redundant: #include <stdio.h>

BSP/Simulator/Src/BSP_Contactors.c
cnflood: See changes outlined in my response to your header file comments.

BSP/Simulator/Src/BSP_Contactors.c
cnflood: See changes outlined in my response to your header file comments.

Tests/Test_BSP_Contactors.c
cnflood: Test now has an infinite loop that accepts user input in the command line. I tested it and it worked without error for me. I wasn't quite sure if this test file was what you're looking for, so let me know if I need to make any changes to it so that it's more useful/helpful!

Tests/Test_BSP_Contactors.c
cnflood: Documented what is being tested and how it should be used.

PR #33, opened 2020-08-17
Motor Controller Driver

### LINKED ISSUE #10: Motor Controller Driver
**Summary**: This module interacts with the Tritium motor controller to send and receive the correct messages periodically in order to drive the motor.

**Requirements**:
1. Function to initialize the motor controller that essentially calls BSP_CAN_Init with CAN2 as the parameter since CAN2 is the line that is connected to the motor controller
2. Function to write the MOTOR DRIVE command to the bus. The required ID and message contents are in the OneNote on the Interfaces page.
3. Function to read the VELOCITY command from the bus. This should only return a SUCCESS if the ID on the bus is the expected ID (also on the OneNote), and should write velocity in both m/s and rpm to 2 separate buffer values passed as parameters. 

**Purpose**: This allows us to abstract the IDs and message structure required by the Tritium so that we can simply call these functions to get and send information. The send function will be set on a periodic interrupt using the Timers, and the read function will be called during the main loop of execution or in one of the main threads (whatever we decide on when making the architecture)

sugamarora99:
Closes #10

### DIFF
--- a/Drivers/Inc/MotorController.h
+++ b/Drivers/Inc/MotorController.h
@@ -28,7 +28,7 @@ void MotorController_Drive(uint32_t newVelocity, uint32_t motorCurrent);
 /**
  * @brief   Reads most recent command from CAN2 bus
  * @param   message the buffer in which the info for the CAN message will be stored
- * @return  true if a message is read
+ * @return  SUCCESS if a message is read
  */ 
 error_t MotorController_Read(CANbuff *message);
 
--- a/Drivers/Inc/MotorController.h
+++ b/Drivers/Inc/MotorController.h
@@ -8,21 +8,22 @@
  * @param   None
  * @return  None
  */ 
-void Motor_Controller_Init();
+void Motor_Controller_Init(void);
 
 /**
  * @brief   Sends MOTOR DRIVE command on CAN2
- * @param   rpmBuff the car's velocity in rpm
- * @param   mpsBuff the car's velocity in m/s
+ * @param   newVelocity desired motor velocity setpoint in m/s
+ * @param   motorCurrent desired motor current setpoint as a percentage of max current setting
  * @return  None
  */ 
-void Motor_Controller_Drive(uint32_t *rpmBuff, uint32_t *mpsBuff);
+void Motor_Controller_Drive(uint32_t newVelocity, uint32_t motorCurrent);
 
 /**
  * @brief   Reads VELOCITY command from CAN2
- * @param   None
+ * @param   rpmBuff the car's velocity in rpm
+ * @param   mpsBuff the car's velocity in m/s
  * @return  true if the ID on the bus is the expected ID for VELOCITY
  */ 
-bool Motor_Controller_ReadVelocity();
+bool Motor_Controller_ReadVelocity(uint32_t *rpmBuff, uint32_t *mpsBuff);
 
 #endif
--- a/Drivers/Inc/MotorController.h
+++ b/Drivers/Inc/MotorController.h
@@ -26,11 +26,10 @@ void MotorController_Init(void);
 void MotorController_Drive(uint32_t newVelocity, uint32_t motorCurrent);
 
 /**
- * @brief   Reads desired command from CAN2
- * @param   message the car's information regarding the desired command
- * @param   expectedID the id for the command the user wants to check CAN2 for
- * @return  true if the ID on the bus is the expected ID 
+ * @brief   Reads most recent command from CAN2 bus
+ * @param   message the buffer in which the info for the CAN message will be stored
+ * @return  true if a message is read
  */ 
-bool MotorController_Read(CANbuff *message);
+error_t MotorController_Read(CANbuff *message);
 
 #endif
--- a/Drivers/Inc/MotorController.h
+++ b/Drivers/Inc/MotorController.h
@@ -1,0 +1,28 @@
+#ifndef __MOTORCONTROLLER_H
+#define __MOTORCONTROLLER_H
+
+#include <bsp.h>
+
+/**
+ * @brief   Initializes the motor controller
+ * @param   None
+ * @return  None
+ */ 
+void Motor_Controller_Init();
+
+/**
+ * @brief   Sends MOTOR DRIVE command on CAN2
+ * @param   rpmBuff the car's velocity in rpm
+ * @param   mpsBuff the car's velocity in m/s
+ * @return  None
+ */ 
+void Motor_Controller_Drive(uint32_t *rpmBuff, uint32_t *mpsBuff);
+
+/**
+ * @brief   Reads VELOCITY command from CAN2
+ * @param   None
+ * @return  true if the ID on the bus is the expected ID for VELOCITY
+ */ 
+bool Motor_Controller_ReadVelocity();
+
+#endif
--- a/Drivers/Inc/MotorController.h
+++ b/Drivers/Inc/MotorController.h
@@ -4,6 +4,13 @@
 #include <bsp.h>
 #include "CAN.h"
 
+typedef struct 
+{
+    uint32_t id;
+    uint32_t firstNum;
+    uint32_t secondNum;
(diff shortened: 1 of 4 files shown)


Drivers/Inc/MotorController.h
```diff
@@ -0,0 +1,30 @@
+#ifndef __MOTORCONTROLLER_H
+#define __MOTORCONTROLLER_H
+
+#include <bsp.h>
+#include "CAN.h"
```
JimothyGreene, line 5: Why is this include needed?

Drivers/Src/MotorController.c
```diff
@@ -0,0 +1,68 @@
...
+    }
+    int i = 0;
+    while(index < 8){
+        data[index] = (motorCurrent >> (8 * (4-i-1))) & 0xFF;
+        index++;
+        i++;
+    }
+    
+    BSP_CAN_Write(CAN_2, MOTOR_DRIVE, data, 8);
+}
```
JimothyGreene, line 30: Both of these look very good

Drivers/Src/MotorController.c
```diff
@@ -0,0 +1,68 @@
...
+                if(k != length-1){
+                    mpsSum = mpsSum << 8;
+                }
+            }
+        }
+        *rpmBuff = rpmSum;
+        *mpsBuff = mpsSum;
+        return SUCCESS;
+    }
+    return FAILURE;
```
JimothyGreene, line 67: This looks very good. The only thing now is that after we discussed our threads and after I've looked at the MotorController CAN protocol a bit more, I'd like to change this function slightly. Instead of this being for the Velocity command only, I'd like this to be applicable to all informational messages the MC sends to us. All of these can be found on the Google Drive on the CAN Bus IDs spreadsheet. 

That being said, the best way I see to do this is to create a struct that is used as the buffer here. The struct can hold 3 values:
1. The ID of the received message
2. The first number of the message (bits [63:32])
3. The second number of the message (bits [31:0])

This way, it can be applied for all informational commands. The general structure of this function that you have can still be kept since all of the messages are structured the same way (split into 2 32-bit pieces).

Drivers/Inc/MotorController.h
```diff
@@ -0,0 +1,30 @@
...
+ */ 
+void MotorController_Init(void);
+
+/**
+ * @brief   Sends MOTOR DRIVE command on CAN2
+ * @param   newVelocity desired motor velocity setpoint in m/s
+ * @param   motorCurrent desired motor current setpoint as a percentage of max current setting
+ * @return  None
+ */ 
+void MotorController_Drive(uint32_t newVelocity, uint32_t motorCurrent);
```
chaseblock, line 20: I think I've had this conversation with someone before, but I can't quite remember the answer from then. Which of these parameters takes priority? If we want some velocity _v_ and some current _i_, but in order to maintain _v_ we need to supply a current _i<sub>1</sub>_ > _i_, then what will the current be?

Drivers/Src/MotorController.c
```diff
@@ -0,0 +1,68 @@
+#include "MotorController.h"
+
+
+#define MOTOR_DRIVE 0x221
+#define VELOCITY_ID 0x243
+#define SUCCESS     1
+#define FAILURE     0
```
chaseblock, line 7: These seem pretty generic. Are there global error codes available outside of the MotorController.c file that could be used instead (something common to all drivers/apps)?

JimothyGreene: Nothing yet, but we should add something in `config.h` that can be used globally. I'll make sure that happens.

Drivers/Inc/MotorController.h
JimothyGreene: There are 2 modes that the Tritium operates under. 
1. Torque control. During this mode, we always have to send a desired velocity that is unobtainable (some very large value). And the motor current is simply proportional to the percentage of the accelerator pressed. In that case, only the motor current is relevant.
2. Velocity control. During this mode, the tritium will use the maximum available current to try and achieve the desired velocity. The way the documentation reads, it seems like the current set point will take priority.

Drivers/Inc/MotorController.h
JimothyGreene: Still this

Drivers/Src/MotorController.c
```diff
@@ -0,0 +1,71 @@
...
+    while(index < 8){
+        data[index] = (motorCurrent >> (8 * (4-i-1))) & 0xFF;
+        index++;
+        i++;
+    }
+    
+    BSP_CAN_Write(CAN_2, MOTOR_DRIVE, data, 8);
+}
+
+bool MotorController_Read(CANbuff *message, uint32_t expectedID){
```
JimothyGreene, line 34: I would take out the expected ID here. We want to simply get whatever message is being sent on the Motor Controller CAN bus and pack it into a struct. That way, we can put it into our FIFO eventually and forward the messages to Data Acquisition.

Drivers/Src/MotorController.c
```diff
@@ -0,0 +1,71 @@
...
+                    firstSum = firstSum << 8;
+                }
+            }
+        }else{
+            //get first number (bits 0-31)
+            for(int j = 0; j < 4; j++){
+                firstSum += data[j] & 0xFF;
+                if(j != 3){
+                    firstSum = firstSum << 8;
+                }
```
JimothyGreene, line 55: This is a little repetitive here. Try to consolidate this so the same code doesn't need to be written. Honestly, none of the messages we need to receive have less than the 64 bits, so I'm not sure we need to account for a message of half the length.

Drivers/Inc/MotorController.h
```diff
@@ -26,11 +26,10 @@ void MotorController_Init(void);
 void MotorController_Drive(uint32_t newVelocity, uint32_t motorCurrent);
 
 /**
- * @brief   Reads desired command from CAN2
- * @param   message the car's information regarding the desired command
- * @param   expectedID the id for the command the user wants to check CAN2 for
- * @return  true if the ID on the bus is the expected ID 
+ * @brief   Reads most recent command from CAN2 bus
+ * @param   message the buffer in which the info for the CAN message will be stored
+ * @return  true if a message is read
```
JimothyGreene, line 31: This returns an error_t now

Drivers/Src/MotorController.c
```diff
@@ -31,36 +31,26 @@ void MotorController_Drive(uint32_t newVelocity, uint32_t motorCurrent){
...
-        if(length < 5){
-            for(int i = 0; i < length; i++){
-                firstSum += data[i] & 0xFF;
-                if(i != length-1){
-                    firstSum = firstSum << 8;
-                }
+        //get first number (bits 0-31)
+        for(int j = 0; j < 4; j++){
+            firstSum += data[j] & 0xFF;
+            if(j != 3){
```
JimothyGreene, line 45: Some magic numbers floating around here. Try to define these values. Something like defining the 8 (size of data array) as a MAX_CAN_LEN or smth and then making the other values relative to that. Easier to read and change if necessary.

Drivers/Src/MotorController.c
```diff
@@ -6,7 +6,7 @@
 #define SUCCESS     1
 #define FAILURE     0
```
JimothyGreene: These can be removed now also. Sorry didn't catch this earlier

sugamarora99: Fixed

JimothyGreene (changes requested):
I'm getting a segmentation fault when testing this. I can run it once, it gives incorrect ID and data values, and then every time I try to run it after that, I get segfaults. If you're not getting the same thing, I'll look into it, but see if you get it too.

JimothyGreene (changes requested):
Everything seems to work now. Just put the documentation on everything and clean up the code wherever you can and I'll approve this.

JimothyGreene (approved):

PR #34, opened 2020-08-21
Pedal driver

ErickCortez98:
Driver file that provides a function to initialize the pedals and a function that reads the percentage pressed of either pedal as a percentage of the total distance.

### DIFF
--- a/Drivers/Src/Pedals.c
+++ b/Drivers/Src/Pedals.c
@@ -5,6 +5,7 @@
  */
 
 #include "Pedals.h"
+#include "BSP_ADC.h"
 
 /**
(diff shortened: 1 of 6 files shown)


JimothyGreene:
Closes #34

Drivers/Src/Pedals.c
```diff
@@ -0,0 +1,35 @@
+/**
+ * C file that initializes the accelerator and brake pedals
+ * and reads the position of the pedals to see how far the 
+ * pedal has been pressed
+ */
+
+#include "Pedals.h"
+#include "BSP_ADC.h"
```
JimothyGreene, line 8: This doesn't need to be included again

PR #35, opened 2020-08-23
Canbus driver

### LINKED ISSUE #9: CAN Driver
**Summary**: This will provide a layer of abstraction on top of our BSP_CAN code. Our system needs to send and receive specific CAN messages to and from other systems every so often. This module will provide functions for each of those instances. For now, the specific IDs are not defined. As we discuss more of the system integration stuff, we'll decide on the proper IDs and message structures, but this module should be independent of those specific values.

**Requirements**: 

1. Function to initialize the CAN bus that should call BSP_CAN_Init with the CAN1 as the parameter since CAN1 is the bus used for the entire car.
2. (This structure is taken from BPS, so feel free to look at their code for their CANbus Driver). Function to send CAN messages with an enumerated type to describe the different IDs that we need to send messages to from our system. For now, that is probably just the Telemetry system, but that may change in the future. This function should accept that ID and a data packet. This packet can be in the form of a Union, so we can send differing types of data. The data should be converted into a Byte array since our BSP_CAN_Write() function accepts an array pointer and a length of that array. This can be done with a static private function. It will similar to the BPS code but not just to 4 bytes. 
3. Function to read a message from the CAN bus. This will essentially take an expected ID and buffer value as parameters and call BSP_CAN_Read to get back an ID and message. If the ID matches the expected ID, the function should return some SUCCESS value and fill the buffer value with the message. Otherwise, it should return some FAIL value. This will allow us to easily determine if an expected message has been sent.

**Purpose**: This will allow us to easily send CAN messages that are pertinent to our system and the way our CAN bus is setup instead of the broad BSP_CAN code that we've written. Our 2 main CAN responsibilities are to send Motor Controller/Motor information to Telemetry, and to periodically receive a message from BPS telling us whether or not the motor can be safely turned on.

SoroushF79:
Closes #9


Drivers/Inc/CAN.h
```diff
@@ -0,0 +1,54 @@
...
+	TRIP = 0x02,
+    ALL_CLEAR = 0x101,
+    CONTACTOR_STATE = 0x102,
+    CURRENT_DATA = 0x103,
+    VOLT_DATA = 0x104,
+    TEMP_DATA = 0x105,
+    SOC_DATA = 0x106,
+    WDOG_TRIGGERED = 0x107,
+    CAN_ERROR = 0x108
+} CANId_t;
```
JimothyGreene, line 16: Our relevant IDs should be on the CAN Bus IDs spreadsheet on the google drive. We send the Car State command and should be receiving the Motor Disable command. Also, we'll be receiving messages on the other CAN bus (CAN2) that we'll need to forward onto this bus using the same IDs. All of the IDs and command names/structures can be found on the Drive spreadsheet.

Drivers/Inc/CAN.h
```diff
@@ -0,0 +1,54 @@
...
+int CANbus_Send(CANId_t id, CANPayload_t payload);
+
+/**
+ * @brief   Checks if the CAN ID matches with expected ID and then copies message to given buffer array
+ * @param   CAN line bus
+ * @param   CAN message ID
+ * @param   pointer to buffer array to store message
+ * @return  0 if ID matches and 1 if it doesn't
+ */
+int CANbus_Read(CAN_t bus, uint32_t id, uint8_t* buffer);
```
JimothyGreene, line 52: I would change these function names from using `CANbus` to just `CAN`

Drivers/Src/CAN.c
```diff
@@ -0,0 +1,120 @@
...
+
+static void floatToBytes(float val, uint8_t* bytes_array, uint8_t bytes);
+
+/**
+ * @brief   Initializes the CAN system
+ * @param   None
+ * @return  None
+ */
+
+// Not sure if we want CAN1 to be an argument to CAN_Init or not
```
JimothyGreene, line 11: This initialization should always initialize CAN1, so it's good as it is right now

Drivers/Src/CAN.c
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * @brief   Transmits data onto the CANbus
+ * @param   id : CAN id of the message
+ * @param 	payload : the data that will be sent.
+ * @return  0 if data wasn't sent, otherwise it was sent.
+ */
+int CANbus_Send(CANId_t id, CANPayload_t payload) {
+
+	uint8_t bytes = 4; // Not sure how we want to decide the number of wanted bytes. An extra paramter to this function?
```
JimothyGreene, line 24: This can be part of the payload struct

Drivers/Src/CAN.c
```diff
@@ -0,0 +1,120 @@
...
+		case SOC_DATA:
+			uint8_t txdata[bytes];
+			floatToBytes(payload.data.f, &txdata[0]);
+			return BSP_CAN_Write(id, txdata, bytes);
+
+		case WDOG_TRIGGERED:
+			return BSP_CAN_Write(id, &payload.data.b, 1);
+
+		case CAN_ERROR:
+			return BSP_CAN_Write(id, &payload.data.b, 1);
```
JimothyGreene, line 59: Same thing as the header file with the IDs. Since this is for the send function, we only need it to work for the Car State command and all the Motor Controller information commands right now.

Drivers/Src/CAN.c
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * @brief   Checks if the CAN ID matches with expected ID and then copies message to given buffer array
+ * @param   CAN line bus
+ * @param   CAN message ID
+ * @param   pointer to buffer array to store message
+ * @return  0 if ID matches and 1 if it doesn't
+ */
+
+int CANbus_Read(CAN_t bus, uint32_t Expected_ID, uint8_t* buffer)
```
JimothyGreene, line 102: I think I want to change this from taking an expected ID. Since we'll have the CAN IDs listed in the header file, we can just check if it matches any of the IDs that we are supposed to be **receiving**, and return a fail if it doesn't. For right now, this should only be the motor disable command from BPS.

Drivers/Inc/CAN.h
```diff
@@ -0,0 +1,48 @@
+#ifndef CAN_H__
+#define CAN_H__
+
+typedef enum {
+	MOTOR_DRIVE = 0x221,
+    MOTOR_POWER = 0x222,
+    RESET = 0x223,
+    VELOCITY = 0x243
```
JimothyGreene, line 8: These are the IDs that we're sending to the Motor Controller. The IDs that we'll be sending to the other car systems can be found here: https://docs.google.com/spreadsheets/d/11YWoMVZw8BFr8kyO4DIz0g-aIU_vVa0d-WioSRq85TI/edit#gid=0
under the Controls section.
Most of them are the forwarded messages that we receive from the motor controller.

Drivers/Src/CAN.c
```diff
@@ -0,0 +1,95 @@
...
+
+/**
+ * @brief   Checks if the CAN ID matches with expected ID and then copies message to given buffer array
+ * @param   CAN line bus
+ * @param   CAN message ID
+ * @param   pointer to buffer array to store message
+ * @return  0 if ID matches and 1 if it doesn't
+ */
+
+int CAN_Read(CAN_t bus, uint32_t Expected_ID, uint8_t* buffer)
```
JimothyGreene, line 77: I actually want to remove the `Expected_ID` part of this. We'll really only be reading a message from BPS called Motor Disable, so you can just account for just this message rn.

JimothyGreene (changes requested):
Functionality seems like it'll work, but I'm going to hold off on testing rn. Need to fix the CAN IDs. This bus is for the main car, which is separate from the motor controller. I left a link to the CAN Bus IDs spreadsheet in the comments.

Drivers/Inc/CAN.h
```diff
@@ -29,21 +36,25 @@ typedef struct {
...
  * @brief   Checks if the CAN ID matches with expected ID and then copies message to given buffer array
  * @param   CAN line bus
  * @param   pointer to buffer array to store message
  * @return  0 if ID matches and 1 if it doesn't
  */
+<<<<<<< Updated upstream
 int CANbus_Read(CAN_t bus, uint8_t* buffer);
+=======
+int CAN_Read(CAN_t bus, uint32_t id, uint8_t* buffer);
+>>>>>>> Stashed changes
```
JimothyGreene: Fix this merge conflict

Drivers/Src/CAN.c
```diff
@@ -0,0 +1,98 @@
...
+		case MC_BUS:
+		case VELOCITY:
+		case MC_PHASE_CURRENT:
+		case VOLTAGE_VEC:
+		case CURRENT_VEC:
+		case BACKEMF:
+		case TEMPERATURE:
+		case ODOMETER_AMPHOURS:
+		case CAR_STATE:
+			return BSP_CAN_WRITE(bus,id, &payload.data.b, payload.bytes)
```
JimothyGreene, line 34: CAR_STATE can use the payload.data.b
Everything else should use the payload.data.d (64 bits) for the message

Drivers/Src/CAN.c
```diff
@@ -0,0 +1,98 @@
...
+
+	// Reverse the array. I don't know why this has to be done, to be honest, but that's what
+	// they did in floatTo4Bytes in BPS/Drivers/Src/CANbus.c . This is a generalized version of
+	// what was done in that file so that we can reverse an array of any length.
+	for(i=0; i<bytes/2; i++){
+		temp = bytes_array[i];
+		bytes_array[i] = bytes_array[bytes-1-i]
+		bytes_array[bytes-1-i] = temp 
+	}
+}
```
JimothyGreene, line 70: I don't think this is being used anymore. If not, take this out and remove the float field from the union

JimothyGreene (changes requested):
Almost there. Few more things then I'll do some functionality testing.

Drivers/Inc/CAN.h
rafidh611: What is a merge conflict? Like something is wrong in the code?

JimothyGreene: The whole <<<< and >>>> stuff. Means there are 2 versions of that file tracked by Git and it just combined them. You have to essentially choose the right one and delete the other one.

JimothyGreene: I think it's good now.

Drivers/Src/CAN.c
```diff
@@ -0,0 +1,68 @@
...
+	switch (id) {
+		case MC_BUS:
+		case VELOCITY:
+		case MC_PHASE_CURRENT:
+		case VOLTAGE_VEC:
+		case CURRENT_VEC:
+		case BACKEMF:
+		case TEMPERATURE:
+		case ODOMETER_AMPHOURS:
+        case MOTOR_DISABLE:
```
JimothyGreene, line 32: Don't need to send this MOTOR_DISABLE message, so it can be removed from this function

JimothyGreene (changes requested):
One small comment left. Also the flipped message issue.

PR #36, opened 2020-08-29
Displayboard

ErickCortez98:
- Added a UART enum UART_t with two UARTS
- Added extra UART_t parameter to both read and write functions of UART file
- Each function now reads or writes to the first or second line of the csv file depending on which UART we passed as a parameter to the function
- Added test file that prints random data in both UARTs to the terminal
- Modified documentation of the header file and c file
- Deleted the UART2 file and modified any file that made use of it by now using the UART file and its modified methods


BSP/Inc/bsp.h
```diff
@@ -18,6 +18,7 @@
 #define CONTACTORS_CSV "Contactors.csv"
 #define PEDALS_CSV "Pedals.csv"
 #define UART_CSV "UART.csv"
+#define UART2_CSV "UART2.csv"
```
JimothyGreene, line 21: This doesn't need to be defined anymore

BSP/Inc/BSP_UART.h
```diff
@@ -9,25 +9,32 @@
 #include "common.h"
 #include <bsp.h>
 
+typedef enum {UART_1, UART_2, NUM_UART} UART_t;
 /**
  * @brief   Initializes the UART peripheral
  */
 void BSP_UART_Init(void);
```
JimothyGreene, line 16: Header file needs to be generic, so just have this accept a `UART_t` as a parameter. It won't change anything for simulator, but the initialization will be different for the real board depending on that argument.

JimothyGreene (changes requested):
All working. Great job. I want to add a part of the simulator GUI that allows us to enter text as if the Controls system is sending a message over UART. Basically just a little part of the GUI that has a text box. When you send that information, it should write it to the UART CSV file (for now, we'll call the Display UART UART_1 and this new UART UART_2, but that'll probably change later). Then, on the C side, if we read the UART_2 line, we should be able to read that message.

BSP/Simulator/Hardware/UART.py
```diff
@@ -0,0 +1,42 @@
+import csv
+import os
+import fcntl
+
+# path to the file
+file = "BSP/Simulator/Hardware/Data/UART.csv"
+
+
+def write(msg, uart_num):
```
JimothyGreene, line 9: This module would only be used for UART_2 right now since UART_1 is the display, so it can just assume 2 and not take a parameter.

JimothyGreene (changes requested):
This is super awesome thanks. One small comment that I left, so address that. And then if you could update the Test_BSP_UART to print the last sent message from UART_2 instead of using it the same way as UART_1. I looked in the CSV file so it works, but this is just so that there's a test file for it.

JimothyGreene (approved):
Amazing thank you. I'll pull this in now. Great work on this!

PR #41, opened 2020-09-06
Update submodules to use http instead of ssh.

chaseblock:


chaseblock:
From what I can tell, this is required in order to get the github actions to pull in the submodules properly (as it does not possess an ssh key to do so).

JimothyGreene (approved):

PR #46, opened 2020-09-12
Lights simulation: Fixes #32 

### LINKED ISSUE #32: Lights Simulation
Simulation for the LEDs (both internal and external are controlled by software now)

**Summary**: This module simulates the internal and external LEDs as they are controlled by our system.

**Requirements**:
1. Python module to read current state of lights to be provided to the simulator GUI
2. Addition to GUI that displays the lights and their states. Colors are appreciated
3. C module that should be able to set the state of each light as well as pseudo initialize the lights
4. C Test file to show that we can set the states of each light. Should ideally take user input.

babusid:
- Created Lights.C which has Read and Update functions for all Lights.
- Created Lights.H which exposes Lights.C to the main simulator.
- Wrote Test file for Lights which switches all lights on and off on a loop and prints it to the console.
- Updated simulate.py and lights.py to include all ten lights built into the Lights C module.


Tests/Test_BSP_Lights.c
```diff
@@ -0,0 +1,32 @@
+#include "common.h"
+#include "config.h"
+#include <bsp.h>
+
+int main() {
+    BSP_Lights_Init();
+    printf("M_CNCTR\tRSVD_LED\tCTRL_FAULT\tRIGHT_BLINK\tHeadlightPWR\tA_CNCTR\tBPS_FAULT\tLEFT_BLINK\tBPS_PWR\tBrakeLight\n");
+    while(1) {
+        printf("%d\t%d\t%d\t%d\t%d\t%d\t%d\t%d\t%d\t%d\r",
```
Cam0Cow, line 9: The values aren't lined up under their respective labels, they should probably be more spaced out.

Cam0Cow (changes requested):

BSP/Inc/BSP_Lights.h
```diff
@@ -0,0 +1,33 @@
...
+* @return  returns state enum which indicates ON/OFF
+*/ 
+state_t BSP_Lights_Read(LIGHT_t LightChannel);
+
+/**
+* @brief   Switch light between ON/OFF states
+* @param   LightChannel Which light to switch
+* @return  void
+*/ 
+void BSP_Lights_Switch(LIGHT_t LightChannel);
```
JimothyGreene, line 29: We don't always want to toggle the light, so I would make this function called `BSP_Lights_Set()` and also take in a `state_t` parameter. That way we don't need to know whether or not the lights are on to turn them on or off to turn them off. Mainly it'll just relieve us of one step in the application layer.

BSP/Simulator/Src/BSP_Lights.c
```diff
@@ -0,0 +1,98 @@
...
+
+
+}
+
+/**
+* @brief   Switch the selected light between ON/OFF
+* @param   LightChannel Which Light you want to switch
+* @return  void
+*/ 
+void BSP_Lights_Switch(LIGHT_t LightChannel){
```
JimothyGreene, line 61: See comments in header file

JimothyGreene (changes requested):
One main functionality change. See in line comments.

BSP/Simulator/Src/BSP_Lights.c
```diff
@@ -0,0 +1,98 @@
...
+    FILE* fp = fopen(FILE_NAME, "r");
+    //will this error case ever be reached? Init function won't fire if file is unavailable
+    if (!fp) {
+        printf("Lights not available\n\r");
+        return 2;
```
ErickCortez98, line 66: Returning value inside a void function. 
```suggestion
        return;
```

ErickCortez98 (changes requested):
Small details, see comments.

JimothyGreene (approved):
Looks all good

PR #47, opened 2020-09-12
BSP_CAN.c changed to clear file upon read

AleNeff:


JimothyGreene (changes requested):
Currently, this will clear the entire CSV file, which means both lines (both CAN bus entries). Instead, I'd like only the line in the file that was read from to be cleared and the other to stay intact.

AleNeff:
Should be good to go now with correct implementation

Cam0Cow (changes requested):
Reading from CAN1 while CAN2 is an empty line (or if there's no line for CAN2 at all) causes the BSP_CAN_Read function to overwrite CAN2 with 3 or 4 random characters (and those characters aren't consistent between executions)

BSP/Simulator/Src/BSP_CAN.c
```diff
@@ -124,6 +126,26 @@ uint8_t BSP_CAN_Read(CAN_t bus, uint32_t* id, uint8_t* data) {
         data[i] = (fullData >> (8 * (len-i-1))) & 0xFF;
```
ErickCortez98, line 126: Initialize _len_ variable to 0. 
It seems to happen that since the line we are reading form the csv file in testCAN3() is empty, _len_ doesn't get initialized to 0 with sscanf from the char array currentCAN but a random number. This causes conflicts with this for loop running a random number of times (depending on what value _len_ got randomly initizialized to) because the _data_ array's size might be less that _len_.

ErickCortez98 (dismissed):
Initialize variable _len_ to 0 inside the BSP_CAN_Read function. Details in comment.

JimothyGreene (approved):
Working properly. Looks good.

PR #64, opened 2020-10-03
Added class that holds all the global variables and constants of the car state to clean a little bit the code

ErickCortez98:
Added class at the bottom of the file named CarState that holds all the variables that used to be globals now as class attributes. The variables and globals are accessed in the form CarState._variable_name_


JimothyGreene (approved):
Looks good to me. Much cleaner now.

Cam0Cow (approved):

PR #65, opened 2020-10-03
simulate.py can update CAN with empty list

SoroushF79:


BSP/Simulator/Hardware/simulate.py
```diff
@@ -46,8 +46,13 @@ def update_CAN():
     """Periodically update the display state of the CAN bus"""
     global id_text, message_text
     can = CAN.read()
+    
+    if not can: # If can is empty list
+        can.extend(["Empty", "Empty"])
+
```
JimothyGreene, line 52: Can you implement this logic in the `CAN.py` file instead? I want the `simulate.py` module to be as unaware as possible as to what's going on behind the scenes.

JimothyGreene (changes requested):
Logic looks good. Just move it to be more specified.

JimothyGreene (approved):
Looks good and works as far as I can tell

ErickCortez98 (approved):
Working well on my side too

PR #66, opened 2020-10-10
Bsp unified tests

Skebob:


JimothyGreene (changes requested):
Update this by merging with master first. There are some bugs in here that are caused by issues that I believe were resolved by future commits.

JimothyGreene (approved):
It all looks good to me. We've found one bug but it's not with this test. I'll make an issue for it.

Cam0Cow (approved):

PR #69, opened 2020-10-10
Integrated Driver Testfile

babusid:
Tests all drivers in their current state in one test file.

### DIFF
--- a/Tests/Test_IntegratedDriverTest.c
+++ b/Tests/Test_IntegratedDriverTest.c
@@ -8,6 +8,7 @@
 
 int main() {
     //CAN Test
+    printf("Testing CAN module:");
     CANbus_Init();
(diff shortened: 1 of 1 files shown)


JimothyGreene (changes requested):
This should ideally run in some sort of infinite loop so it can just run while we test things on the simulator UI and interactions between drivers.

JimothyGreene (changes requested):
Looks like it's working pretty well. Just try to format the output a bit nicer. Some more spacing so that the title of the column lines up with its data.

ErickCortez98 (changes requested):
Seems to be working great, but it would maybe be useful to print out some comments specifying which module is being tested at each part of the process as well as maybe which part in the GUI can be used to modify or visualize the outputs

JimothyGreene (approved):
Works well for me now. I think we'll be integrating this one into `master`

Cam0Cow (approved):

PR #71, opened 2020-10-10
Script to make simulator launch easier

rafidh611:


JimothyGreene (changes requested):
Can you rename the file to just be called `simulate`? That way we just have to run `./simulate` to run it.

JimothyGreene (changes requested):
Working well. You can delete the simulate.py symbolic link that's in the root Controls folder as part of this PR as well. Since that's no longer needed.

JimothyGreene (approved):
Looking good and works well for me

ErickCortez98 (approved):
Working well for me. 
Note: I got issues when trying to run it under the WSL directory, after changing the Controls directory to Windows the script worked.

PR #72, opened 2020-10-10
SPI Simulator

### LINKED ISSUE #63: SPI Simulation
Now we're going to using SPI to communicate with the steering wheel (for internal lights and switches), so we should add a BSP module for the SPI line. This should also move the Switches and Lights interfaces up to Driver level.

JimothyGreene:
This module adds a hardware simulation for the Shift Register as well as a BSP interface to SPI. This will be used by the Switches and Lights to communicate with the Minion board on the steering wheel. Closes #63


BSP/Simulator/Src/BSP_SPI.c
```diff
@@ -0,0 +1,87 @@
...
+
+    // Lock file
+    int fno = fileno(fp);
+    flock(fno, LOCK_EX);
+
+    // Read file
+    char csv[128];
+    uint32_t fullData;
+    fgets(csv, 128, fp);
+    sscanf(csv, "%x", &fullData);
```
Cam0Cow, line 77: The write function writes to the CSV in decimal and reads in hex. Is this intended?

BSP/Simulator/Src/BSP_SPI.c
```diff
@@ -0,0 +1,87 @@
...
+
+    // Split hex data into bytes
+    for (uint8_t i = 0; i < rxLen; i++) {
+        rxBuf[i] = (fullData >> (8 * (rxLen-i-1))) & 0xFF;
```
Cam0Cow, line 81: This reads in reverse order from what was written to the csv.

Tests/Test_BSP_SPI.c
```diff
@@ -0,0 +1,14 @@
...
+int main(void) {
+    BSP_SPI_Init();
+    uint8_t txData[] = {0x41, 0x12, 0x00};
+    uint8_t txLen = 3;
+    BSP_SPI_Write(txData, txLen);
+    uint8_t rxData[1];
+    uint8_t rxLen = 1;
+    usleep(1000000);
+    BSP_SPI_Read(rxData, rxLen);
+    printf("Received Data: %x\n\r", rxData[0]);
```
Cam0Cow, line 12: What exactly is the steering wheel going to return to our board? Should it even return anything at all?

JimothyGreene: If we want to read the status of the switches, we have to send it a command through SPI and then we'll get the data back as a response.

BSP/Simulator/Src/BSP_SPI.c
JimothyGreene: No. I'll add a change in to make it all hex

BSP/Simulator/Src/BSP_SPI.c
JimothyGreene: When I'm testing it, it reads in the correct order. Is there a case that's making it come up in reverse?

BSP/Simulator/Hardware/ShiftRegister.py
```diff
@@ -0,0 +1,153 @@
...
+        0x12: 'GPIOA',
+        0x13: 'GPIOB',
+        0x14: 'OLATA',
+        0x15: 'OLATB'
+    }
+
+    REGISTER_ADDR = [REGISTER_ADDR_0, REGISTER_ADDR_1]
+
+    def write_register(self, register_addr, data):
+        register = self.REGISTER_ADDR[self.REGISTER_DATA['IOCON'] & 0x80][register_addr]
```
Cam0Cow, line 90: This should probably be (self.REGISTER_DATA['IOCON'] & 0x80) >> 7, otherwise you'll get an index of 128

BSP/Simulator/Hardware/ShiftRegister.py
```diff
@@ -0,0 +1,153 @@
...
+                    newData += data & (0x01 << i)
+                else:
+                    newData += oldData & (0x01 << i)
+            self.REGISTER_DATA[f'GPIO{register[-1]}'] = newData
+        else:
+            self.REGISTER_DATA[register] = data
+
+
+    def read_register(self, register_addr):
+        return self.REGISTER_DATA[self.REGISTER_ADDR[self.REGISTER_DATA['IOCON'] & 0x80][register_addr]]
```
Cam0Cow, line 107: Same idea here, you want to access index 1 when IOCON has the high bit set

Cam0Cow (approved):
Works as intended

ErickCortez98 (approved):

PR #74, opened 2020-10-11
Ale neff large scale test

AleNeff:
This Test isn't done but I'm opening a draft pull request because it's close -- the only thing left (I believe) is somehow testing Motor Controller functionality


JimothyGreene (changes requested):
Add a part of this to test the `CANbus` module as well. That part can probably just write a message and then read it back to make sure it's all working properly. Also, try and find a better method of clearing the screen instead of running the `clear` command each time. If you take a look at the BSP full test (I think there's a PR still open but it might get merged in soon), there's a method on there that works well.

AleNeff:
I looked at the BSP method but when I branch into there and run that test
it doesn’t actually work, it just spams to the terminal over & over without
clearing the lines before?

AleNeff:
I've added an option at the start to select what will be tested (Either the CAN or everything else). The CAN test is implemented (it's basically a copy of the pre-existing CAN test) but I'm not 100% sure the CAN module works as it's supposed to.

JimothyGreene (changes requested):
The Everything Else option seems to only test the pedals and some of the switches

AleNeff:
The stuff in the terminal isn't really the test, it's more the interaction with the actual simulator. I'll ask about what else it should do in the meeting.

JimothyGreene:
I think we're going to merge in the Driver test from @babusid 's PR. Closing this one now.

PR #76, opened 2020-10-17
Display update

### LINKED ISSUE #38: New method for Display communication
`snprintf()` is not the best method, so we want to change this to something a little quicker. Need to figure out a better solution and then implement on Controls master board and Gecko.

Cam0Cow:
Updated the display code on the board and in the simulator to support a new method to pass data to the gecko board. A hex string is now passed to the gecko, eliminating the need to parse a float value. Closes #38

### DIFF
--- a/BSP/Simulator/Hardware/Display.py
+++ b/BSP/Simulator/Hardware/Display.py
@@ -3,6 +3,7 @@
 import csv
 import os
 import fcntl
+import struct
 
 # Path of file
 file = "BSP/Simulator/Hardware/Data/UART.csv"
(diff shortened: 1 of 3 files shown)


JimothyGreene (changes requested):
Working great, so nothing to say on that front. There are some compiler warnings with the signedness of certain values, so take a look and try to fix those up. Also, try to add a few comments in the `Display.c` to explain a little more about what each part is doing.

JimothyGreene (approved):
Perfect thanks. All looks great!

ErickCortez98 (approved):

PR #79, opened 2020-10-17
Ale cur vel fix

AleNeff:
Switched placement of current bits and velocity bits for csv file to be read correctly by simulator


JimothyGreene (changes requested):
Overall, this doesn't seem like it switched anything. You may have switched them and then switched back? I'm not sure. All I'm seeing is that you removed the incrementing of the `index` variable.

AleNeff:
Yeah I think I accidentally reverted the change, newest commit & push fixed it.

JimothyGreene (approved):
Looks good to me. Should follow the correct format now.

Cam0Cow (approved):

PR #87, opened 2020-11-07
Bps simulator enhacements

SoroushF79:
Simulator now has checkbox for charging. TEST_CAN leads to stack smashing.

### DIFF
--- a/BSP/Simulator/Hardware/simulate.py
+++ b/BSP/Simulator/Hardware/simulate.py
@@ -23,6 +23,7 @@ CONTACTOR_FREQ = 500
 DISPLAY_FREQ = 500
 LIGHTS_FREQ = 500
 PRECHARGE_FREQ = 500
+MOTOR_DISABLE_FREQ = 500
 
 
 def update_contactors():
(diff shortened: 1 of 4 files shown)


SoroushF79:
I forgot that the id shows up in decimal not hex but I couldn't find the problem for that

BSP/Simulator/Hardware/simulate.py
```diff
@@ -87,6 +89,9 @@ def update_lights():
         light.set(f"{lights[i]}: {(lights_state >> i) & 0x01}")
     window.after(LIGHTS_FREQ, update_lights)
 
+def update_MotorDisable():
+    """Sends MOTOR_DISABLE message when checkbox is checked/unchecked"""
+    MotorMessage.sendMotorDisable(chargingBool.get())
```
JimothyGreene, line 94: Make sure that you add a call to the `charging_frame.after()` method here so it calls this method periodically. Right now it's only calling this method when the checkmark is pressed. We want the BPS message to be continuously broadcasted on a periodic interval all the time. The only thing that changes is the data packet.

JimothyGreene (changes requested):
See comments

JimothyGreene (changes requested):
Looking better now. Can you get it to display the ID in hex? You can use the Python built-in method `hex()` to get the hex string of an integer

JimothyGreene (approved):
Looks good to me

Cam0Cow (approved):

PR #90, opened 2020-11-21
Error Handling in Simulator

### LINKED ISSUE #78: Add exception handling in simulator
example: read() function in MotorController.py
-Exceptions need to be used throughout to avoid modules in the UI hanging during simulation

JimothyGreene:
Up until this point, some of our simulator modules were raising exceptions when certain pieces of data weren't formatted correctly or there was a blank line in a CSV file. Now, instead of breaking that part of the simulator when this error arises, the exception is handled gracefully, and execution of the simulator still continues.

Note: This PR also introduces a small update to the `Test_BSP.c` file to account for the new Display data format.

Closes #78

### DIFF
--- a/BSP/Simulator/Hardware/CAN.py
+++ b/BSP/Simulator/Hardware/CAN.py
@@ -16,24 +16,26 @@ def read():
     # Creates file if it doesn't exist
     os.makedirs(os.path.dirname(file), exist_ok=True)
     if not os.path.exists(file):
-        with open(file, 'w'):
+        with open(file, "w"):
             pass
     message = []
-    with open(file, 'r') as csvfile:
-        fcntl.flock(csvfile.fileno(), fcntl.LOCK_EX)
-        csvreader = csv.reader(csvfile)
-        for row in csvreader:
-            message.append(row)
-        fcntl.flock(csvfile.fileno(), fcntl.LOCK_UN)
-    message = [0, 0] if message == [] else message[0]
-
-    if not message: # If can is empty list
-        message.extend(["Empty", "Empty"])
-    return message
+    try:
+        with open(file, "r") as csvfile:
+            fcntl.flock(csvfile.fileno(), fcntl.LOCK_EX)
+            csvreader = csv.reader(csvfile)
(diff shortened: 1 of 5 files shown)


BSP/Simulator/Hardware/CAN.py
```diff
@@ -16,54 +16,59 @@ def read():
...
+            fcntl.flock(csvfile.fileno(), fcntl.LOCK_EX)
+            csvreader = csv.reader(csvfile)
+            for row in csvreader:
+                message.append(row)
+            fcntl.flock(csvfile.fileno(), fcntl.LOCK_UN)
+        try:
+            return [message[0][0], message[0][1]]
+        except IndexError:
+            return ["", ""]
+    except Exception:
```
ErickCortez98, line 33: Do you think it would be maybe better to handle this exception (and the other ones involving IO) as an IOError? Just for style purposes, because I think that's the only type of exception that this could cause.

Cam0Cow (approved):
Doesn't seem to break any component as far as I can tell, however stack smashing is still an issue.

BSP/Simulator/Hardware/CAN.py
JimothyGreene: I was mainly just trying to catch anything that could get dropped in here. I think it should be fine this way if you're okay with it.

Tests/Test_BSP.c
```diff
@@ -86,11 +86,11 @@ int main() {
...
-        printf("ID: 0x%x\nData: ", id);
-        for (uint8_t i = 0; i < len; i++) {
-            printf("0x%x ", rxData[i]);
-        }
-        printf("\n");
+        // printf("ID: 0x%x\nData: ", id);
+        // for (uint8_t i = 0; i < len; i++) {
+        //     printf("0x%x ", rxData[i]);
+        // }
+        // printf("\n");
```
ErickCortez98, line 93: This was removed to be able to fit all the info in the terminal, right?

ErickCortez98 (dismissed):
Works great, very good style fixes

PR #91, opened 2020-11-21
GPIO Precharge

SoroushF79:
Layer Precharge on currently unwritten GPIO functions


BSP/Simulator/Hardware/PreCharge.py
```diff
@@ -3,10 +3,13 @@
...
 import fcntl
+import GPIO
  
-file = "BSP/Simulator/Hardware/Data/PreCharge.csv";
-
+file = "BSP/Simulator/Hardware/Data/GPIO.csv";
 
+def write(data, port=2):
+    GPIO_write(data, port)
+
```
JimothyGreene, line 12: There won't be a `GPIO.py`, so this write function should write the data to the correct line of the `GPIO.csv` itself

BSP/Simulator/Src/BSP_Precharge.c
```diff
@@ -14,43 +14,15 @@
...
-        printf("Precharge file is not found\n");
-        return;
-    }
-
-    int fno = fileno(fp);
-    flock(fno,LOCK_EX);
-
-    int data = 0;
-    fscanf(fp,"%d", &data);
+    int data = BSP_GPIO_Read();
```
JimothyGreene, line 17: You'll end up passing in a parameter here. It'll probably be an enum, so don't worry about it for now. Just a heads up

JimothyGreene (changes requested):
Few minor changes for this prototyped version

Drivers/Src/Precharge.c
```diff
@@ -0,0 +1,53 @@
+/* Copyright (c) 2020 UT Longhorn Racing Solar */
+
+#include "Precharge.h"
+
+#define FILE_NAME DATA_PATH(PRECHARGE_CSV)
+#define PRECHARGE_PORT PORTC
+#define MOTOR_BIT 3
```
ErickCortez98, line 7: These could be part of the header file

ErickCortez98 (changes requested):

JimothyGreene (changes requested):
The Python simulation side doesn't read from the correct port (line) in the GPIO.csv file. It should be reading from line 3 (port C). Also, we don't need a write function since we'll never change the state of the precharge boards from the simulator. I would take a look at the Contactors code that was merged into master recently. It should be nearly identical on both the Python and C sides.

JimothyGreene (approved):
All working!

ErickCortez98 (approved):

PR #92, opened 2020-11-28
BSP GPIO

JimothyGreene:
Simulation for BSP GPIO. This does not change any of the Python code, since those changes will be handled by the issues regarding moving modules up to the Driver level. This creates a layer to access GPIO pins.


Tests/Test_BSP_GPIO.c
```diff
@@ -0,0 +1,42 @@
...
+    char port;
+    char rw;
+    uint16_t data = 0;
+
+    while (1) {
+        printf("Choose Port (A-D): ");
+        port = getchar();
+        getchar();
+
+        if (port - 'A' >= NUM_PORTS) {
```
SoroushF79, line 19: Shouldn't this be a while loop? Why would you continue if they enter an invalid port?

PR #94, opened 2020-11-28
Lights drivers

### LINKED ISSUE #82: BSP_Lights -> Drivers
Now that we have the lights attached to SPI in our board design, our software stack should reflect that. The SPI module should be nearing completion, and we need to move the Lights functionality up to the Drivers level so it can rely on the SPI module instead of acting on its own.

The header/source files should be part of the Drivers now, and they should call on the BSP_SPI_Read/Write functions to execute their tasks. Additionally, the Python files should now be calling on the ShiftRegister to read/write values across SPI instead of writing to their own individual CSV files.

JimothyGreene:
Moves the Lights code to a Driver level to call upon both GPIO and SPI functions depending on the specific light. The Python simulation for the lights now also relies on the GPIO and SPI to read the current states of the lights. This was tested using the original lights test by changing the function names. The same random changing seems to work as it did before.

Closes #82


Cam0Cow (approved):

SoroushF79 (approved):

BSP/Simulator/Src/BSP_SPI.c
```diff
@@ -10,9 +10,7 @@
 #define FILE_NAME DATA_PATH(SPI_CSV)
```
ErickCortez98, line 10: The file name inside the perror function inside init is not SPI_CSV

PR #95, opened 2020-11-28
Contactors Drivers

### LINKED ISSUE #85: BSP_Contactors -> Driver
Once the GPIO simulation is completed, the Contactor functionality should be moved up to the driver level so that we call upon the GPIO functions individually. The Contactors are external pieces of hardware, so it makes the most sense to have them labeled as Drivers and call GPIO functionality which is board specific.

JimothyGreene:
Moves the contactors logic to the driver level to rely on the GPIO layer in the BSP

Closes #85


Drivers/Src/Contactors.c
```diff
@@ -0,0 +1,66 @@
...
+
+#include "Contactors.h"
+
+/**
+ * @brief   Initializes contactors to be used
+ *          in connection with the Motor and Array
+ * @param   contactor the contactor to initialize
+ * @return  None
+ */ 
+void Contactors_Init(contactor_t contactor) {
```
Cam0Cow, line 15: Is there any reason for this function to take an argument?

BSP/Simulator/Hardware/Contactor.py
```diff
@@ -14,15 +17,19 @@ def read():
...
-        return contactor_states 
+    states = []
+    with open(file, "r") as csvfile:
+        fcntl.flock(csvfile.fileno(), fcntl.LOCK_EX)  # Lock file before reading
+        csvreader = csv.reader(csvfile)
+        for row in csvreader:
+            states.append(row)
+        fcntl.flock(csvfile.fileno(), fcntl.LOCK_UN)
+    try:
+        states = int(states[2][0])
```
ErickCortez98, line 31: So the state of both contactors is going to be just glued to one another as a single integer (which we later separate and return individually by left shifting), right? Just to clarify for my own understanding

Drivers/Src/Contactors.c
JimothyGreene: In the event that we would want to initialize one contactor but not the other, I left this here. It likely won't be used in that way, but it keeps it consistent with the other drivers as well.

BSP/Simulator/Hardware/Contactor.py
JimothyGreene: They're connected through GPIO, so both contactors are pins on one port. In that sense, their states are accessed/modified with one single integer where each contactor is represented by an individual bit.

Drivers/Src/Contactors.c
ErickCortez98: ^^
Both contactors are in the same port, so is it because once we actually interface with the actual board we'll also set up each pin in the port individually?

JimothyGreene: Yeah we'll end up initializing each pin as an output pin, so we could end up initializing them separately.

Drivers/Src/Contactors.c
```diff
@@ -0,0 +1,66 @@
...
+/**
+ * @brief   Reads contactor states from a file 
+ *          and returns the current state of 
+ *          a specified contactor
+ * @param   contactor specifies the contactor for 
+ *          which the user would like to know its state (MOTOR/ARRAY)
+ * @return  The contactor's state (ON/OFF)
+ */ 
+State Contactors_Get(contactor_t contactor) {
+    uint16_t data = BSP_GPIO_Read(CONTACTORS_PORT);
```
ErickCortez98, line 28: do you think it would be better to use an 8-bit int?

Drivers/Inc/Contactors.h
```diff
@@ -1,20 +1,25 @@
...
+#define __CONTACTORS_H
 
 #include "common.h"
 #include "config.h"
-#include <bsp.h>
+#include "BSP_GPIO.h"
+
+#define CONTACTORS_PORT PORTC
+#define MOTOR_CNCTR_PIN 2
+#define ARRAY_CNCTR_PIN 1
```
ErickCortez98, line 12: by the way, what's going to be in pin 0 of the port?

ErickCortez98 (dismissed):

SoroushF79 (approved):

PR #96, opened 2020-12-06
Change the Display message to a better format.

chaseblock:
* Half the message size
* Less time spent processing the message
* Less code


chaseblock:
We're using a different display now, right? If that's the case, this PR should be closed.

JimothyGreene:
> We're using a different display now, right? If that's the case, this PR should be closed.

Yeah agreed. Once we get the display we'll revisit this.

PR #97, opened 2020-12-12
SwitchDriver

babusid:
**Python Side**:
- Switches.py modified to read shiftregister data directly rather than through function
- Switches.py Interfaces with GPIO.CSV to write ignition states

**C Side**:
- Switches.C driver created
- Read function reads from SPI for all switches other than Ignition 1 and 2 (these 2 are read from GPIO), and Regen (which doesn't have an SPI bit corresponding to it yet)


Drivers/Src/Switches.c
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+    // If we are trying to get the state of the ignition switches
+    else{ 
+        int ignStates = BSP_GPIO_Read(PORTA);
+        if(sw == IGN_1){
+            return ignStates & 0x1;
+        }else{
+            return (ignStates & 0x2) >> 1;
+        }
```
JimothyGreene, line 68: IGN_1 is on PA1 and IGN_2 is on PA0. These should be reversed.

Drivers/Src/Switches.c
```diff
@@ -0,0 +1,72 @@
...
+ */ 
+State Switches_Read(switches_t sw){
+    uint8_t query[3]={SPI_OPCODE_R,SPI_GPIOA,0x00}; //query GPIOA
+    uint8_t SwitchReadData[4] = {0};
+    BSP_SPI_Write(query,3);
+
+    // If we are not trying to get the state of the ignition switches
+    if(sw != IGN_1 && sw != IGN_2){
+        //for loop is for addressing bug where register address is returned as the data of that register
+        for (uint8_t i = 0; i <= 1; i++)
```
JimothyGreene, line 47: How does this fix that issue?

Drivers/Src/Switches.c
```diff
@@ -0,0 +1,72 @@
+#include "Switches.h"
+#include <time.h>
```
JimothyGreene, line 2: Is this needed?

JimothyGreene (changes requested):
Few small changes before I test it further

Drivers/Inc/Switches.h
```diff
@@ -0,0 +1,27 @@
+#ifndef SWITCHES_H
+#define SWITCHES_H
+#include "BSP_SPI.h"
+#include "BSP_GPIO.h"
+#include "common.h"
+#include "config.h"
+
+
+typedef enum {CRUZ_st=0, CRUZ_EN, HZD_SQ, FR_SW, Headlight_SW, LEFT_SW, RIGHT_SW, REGEN_SW, IGN_1, IGN_2, REV_SW} switches_t;
```
chaseblock, line 9: CRUZ_st and Headlight_SW are screwing up the capitalization pattern.

Drivers/Inc/Switches.h
```diff
@@ -0,0 +1,27 @@
...
+ */ 
+void Switches_Init(void);
+
+/**
+ * @brief   Reads the current state of 
+ *          the specified switch
+ * @param   sw the switch to read
+ * @return  State of the switch (ON/OFF)
+ */ 
+State Switches_Read(switches_t sw);
```
chaseblock, line 25: Is this really the right return value for this? Some of our switches are three-state and the `State` type only has `ON` and `OFF`.

Drivers/Src/Switches.c
```diff
@@ -0,0 +1,89 @@
...
+    }
+    // If we are trying to get the state of the ignition switches
+    else{ 
+        int ignStates = BSP_GPIO_Read(PORTA);
+        if(sw == IGN_1){
+            return (ignStates & 0x2) >> 1;
+        }else{
+            return (ignStates & 0x1);
+        }
+    }
```
chaseblock, line 86: This seems like a great time for a switch statement to make clear exactly which switches are handled in which case. It results in more typing, but I think it's more readable. Up to y'all though.

```suggestion
    switch(sw) {
    case CRUZ_SW:
    case CRUZ_EN:
    case HZD_SQ: 
    case FR_SW: 
    case HEADLIGHT_SW: 
    case LEFT_SW: 
    case RIGHT_SW: 
    case REGEN_SW:
        BSP_SPI_Write(query,3);
        do{
            BSP_SPI_Read(SwitchReadData,2);
        }while(SwitchReadData[0] == SPI_GPIOA);
        if (SwitchReadData[1] & (1 << sw)) {
            return ON;
        } else {
            return OFF;
        }
    
    case REV_SW:
        query[1] = SPI_GPIOB;
        BSP_SPI_Write(query,3);
        do{
            BSP_SPI_Read(SwitchReadData,2);
        }while(SwitchReadData[0] == SPI_GPIOB);
        if (SwitchReadData[1] & (1 << 7)) {
            return ON;
        } else {
            return OFF;
        }
        
    case IGN_1: {
        int ignStates = BSP_GPIO_Read(PORTA);
        return (ignStates & 0x2) >> 1;
    }
    case IGN_2: {
        int ignStates = BSP_GPIO_Read(PORTA);
        return (ignStates & 0x1);
    }
        
    default:
        // Shouldn't happen
        return -1;
    }
```

Tests/Test_Lights.c
```diff
@@ -17,11 +18,11 @@ int main() {
                 Lights_Read(LEFT_BLINK),
                 Lights_Read(BPS_PWR),
                 Lights_Read(BrakeLight));
-        fflush(stdout);
-        
+        fflush(stdout);  
         light_t light = rand() % 10;
         State state = Lights_Read(light);
         Lights_Set(light, !state);
+        usleep(10000);
```
chaseblock, line 25: Just a note: this test does not compile for the STM32F4 target, since we can't rely on POSIX compliance.

chaseblock (commented):
Just some thoughts.

Drivers/Inc/Switches.h
chaseblock: I guess the real question is a which layer we need to worry about the difference between two switches and a three-state switch, if ever.

Tests/Test_Lights.c
JimothyGreene: The tests will be target-specific. Once we have to test on the STM32 it'll be a slightly different test

Drivers/Inc/Switches.h
JimothyGreene: We have 2 values in the enum for the 3-state switches. So we have a `RIGHT_SW` and `LEFT_SW` to handle for that.

JimothyGreene (dismissed):
Seems good to me. @Cam0Cow just make sure the test works properly for you and that you're happy with it.

Cam0Cow (approved):
I've added the missing test case for the Regen switch, everything else seems to work exactly as needed.

JimothyGreene (approved):
Looks good to me

PR #98, opened 2020-12-12
SPI Struct

ErickCortez98:
Added a struct to BSP_SPI.h that contains three data fields: opcode, address, and data. This in order to hold the messages sent through SPI as structs of this type instead of arrays.


BSP/Inc/BSP_SPI.h
```diff
@@ -11,6 +11,13 @@
 #include "common.h"
 #include "bsp.h"
 
+typedef struct _spi_message {
+    int8_t opcode;
+    int8_t requestedPort;
+    int8_t data;
+} spi_message;
```
chaseblock, line 18: It's not immediately obvious to me what these fields are, so maybe better names are in order. Is the first field referring to read/write? If so, maybe a boolean value `write` would be better. The `requestedPort` may be better as an enum value, rather than a plain `int8_t`. Maybe a name like `cs` (for chip select) would be more in line with typical jargon.

Will we ever have to worry about having multiple data bytes in a single frame? If so, then we may want to provide a `uint8_t *data_buf` rather than a plain integer.

chaseblock: Ok I see now that this refers to an operation and register selection specifically for the chip that we are using, which makes more sense to me now. Just caught me off guard to see it in the BSP file rather than a driver layer file.

JimothyGreene (commented):
@chaseblock makes a good point about having these definitions in the BSP layer. It might be worthwhile just for separation of the SPI logic with the specific GPIO expander logic to move this `spi_message` struct as well as the #defines below to the Driver level. Maybe in some sort of `GPIOExpander.h` that both the `Lights` and `Switches` drivers can use.

JimothyGreene (approved):
Looks good to me

Cam0Cow (approved):

PR #117, opened 2021-01-30
Rtos/activate array

### LINKED ISSUE #114: [RTOS] Activate Array Task
This RTOS task is responsible for turning on the array precharge board before closing the main contactors. The state diagrams and some pseudocode can be found on the Google Drive (Electrical Team > Controls > New Design > Software > RTOS), but don't take either resource as absolute truth. This is the time to reason about this application code, so try to think about what exactly each module should be doing.

Cam0Cow:
Implements the tasks that activates the Array. Closes #114


Apps/Src/ActivateArray.c
```diff
@@ -0,0 +1,25 @@
+/* Copyright (c) 2020 UT Longhorn Racing Solar */
+
+#include "ActivateArray.h"
+
+void Task_ArrayConnection(void *p_arg) {
```
chaseblock, line 5: What if the driver turns the key back (into the off position) after turning it on?

Apps/Src/ActivateArray.c
```diff
@@ -0,0 +1,25 @@
...
+
+    (void) p_arg;
+
+    OS_ERR err;
+    CPU_TS  ts;
+
+    // Wait until driver turned the key to the right place
+    OSSemPend(&ActivateArray_Sem4, 0, OS_OPT_PEND_BLOCKING, &ts, &err);
```
chaseblock, line 13: What's setting this semaphore? Why do we need a separate task to enable the array, when the task that is reading from the switch to set this semaphore could instead just be calling an enableArray() or disableArray() function?

Cam0Cow: I'm not quite sure what the original rationale was. As we have them right now, the activate array task and activate motor task turn on the precharge board, wait 5 seconds, and then turn on their respective contactor. Since the key will be turned quickly through both of these positions, we can't really have a blocking call to enableArray() (and a similar blocking call to enableMotor() ).

As for turning the array off, I think we can simply have some disableArray() function, since there's no need to wait for the precharge board. This function hasn't been written yet, however.

JimothyGreene: From what I remember, to answer Chase's first question, the Switches reading task will set the semaphore once the key has been turned appropriately. The main rationale for a separate task, as Roie said, was to avoid blocking other functionality during the wait time. If that's not something that you think is necessary @chaseblock, it would be much simpler to put just a function call instead of a task. 

My opinion on this is that as long as this task isn't causing problems related to context switching (which I doubt it will since it'll be called very infrequently), I think it's fine in its current implementation. However, I would recommend adding the functionality to switch off the array with this as well. I recently changed the name of the task in the design to `ArrayConnection` mainly to imply that the task handles both directions of connection. I'll leave it up to you @Cam0Cow how exactly you want to implement it though.

Apps/Src/ArrayConnection.c
```diff
@@ -0,0 +1,60 @@
...
+        (OS_TICK)NULL,
+        (void*)NULL,
+        (OS_OPT)(OS_OPT_TASK_STK_CLR),
+        (OS_ERR*)&err
+    );
+    // TODO: error handling
+
+    while (1) {
+        // Wait until some change needs to be made to the array state
+        OSSemPend(&ArrayConnectionChange_Sem4, 0, OS_OPT_PEND_BLOCKING, &ts, &err);
```
JimothyGreene, line 46: Same comment here as on #127

Apps/Src/ArrayConnection.c
```diff
@@ -0,0 +1,61 @@
+/* Copyright (c) 2020 UT Longhorn Racing Solar */
+
+#include "ArrayConnection.h"
+#include "Contactors.h"
+
+static void arrayStartup(OS_ERR *err) {
+    Precharge_Write(ARRAY_PRECHARGE, ON); // Turn on the array precharge
+
+    OSTimeDlyHMSM(0, 0, 5, 0, OS_OPT_TIME_HMSM_STRICT, &err); // wait 5 seconds
```
JimothyGreene, line 9: Add this 5 seconds as a macro in the `config.h` file. After looking at the Tritium docs at Pickle, we'll likely have to change this. It'll just make it clearer what this number means.

JimothyGreene (commented):
Seems good to me. Just make the one change to make that a macro. Small change but just keeps it consistent with design.

Config/Inc/config.h
```diff
@@ -12,10 +12,13 @@ typedef enum {OFF = 0, ON} State;
 
 // Wheel diameter in meters
 #define WHEEL_DIAMETER 0.4f
+
 // Any sort of gear reduction from the motor to the wheels
 // Defined as motor velocity / wheel velocity
 #define GEAR_REDUCTION 1.0f
 
+#define PRECHARGE_ARRAY_DELAY(err) (OSTimeDlyHMSM(0, 0, 5, 0, OS_OPT_TIME_HMSM_STRICT, &err))
```
JimothyGreene, line 20: Oh I think I said it wrong. I mainly just meant make the number 5 into the macro. For the number of seconds we need to delay. That's the only thing we need to change really

JimothyGreene (approved):

PR #119, opened 2021-01-30
ReadPedals task

ErickCortez98:
Implements the task ReadPedals


Apps/Inc/Tasks.h
```diff
@@ -133,5 +134,7 @@ extern OS_SEM SendCarCAN_Sem4;
  * Global Variables
  */
 // TODO: Put all global state variables here
+int8_t brake_percentage;
+int8_t accelerator_percentage;
```
chaseblock, line 138: Variables should not be declared in the .h file. It will inevitably lead to a multiply defined symbol error down the line.

Also, do these really need to be global? Or can we get away with an getter function?

Apps/Src/RTOSPedals.c
```diff
@@ -0,0 +1,36 @@
...
+    
+    OS_ERR err;
+
+    while(1){
+        accelerator_percentage = Pedals_Read(ACCELERATOR);
+        brake_percentage = Pedals_Read(BRAKE);
+
+        // Post to semaphores (velocityChange, lightsChange)
+        OSSemPost (&VelocityChange_Sem4, OS_OPT_POST_ALL, &err);
+        OSSemPost (&LightsChange_Sem4, OS_OPT_POST_ALL, &err);
```
chaseblock, line 31: If you're planning on just looping infinitely and posting to these semaphores every time, we may as well just have the velocity and lights threads run without blocking semaphores and with some delay in the loop that still makes them responsive.

This would also probably be a bit better just so a change doesn't sneak by that task when someone forgets to post to a semaphore in the future.

Apps/Src/RTOSPedals.c
```diff
@@ -0,0 +1,36 @@
...
+* @brief   Read both pedal percentage values and set the global variables
+* @param   None
+* @return  None
+*/ 
+void Task_ReadPedals(void *p_arg){
+    Pedals_Values_Init();
+    
+    OS_ERR err;
+
+    while(1){
```
chaseblock, line 25: Add a delay in this loop. Doesn't have to be long, but we don't need to be wasting all our cycles on reading from pedals.

Apps/Inc/Tasks.h
ErickCortez98: Getter functions sound better indeed

JimothyGreene: This structure was copied over from what BPS did with state variables that needed to be accessed by multiple threads but only modified by one. Putting getter functions in a file that really just defines the task thread seems a little excessive to me. We could avoid potential multiply defined symbol errors by creating some sort of car state struct to contain all this information. Just my opinion on this

Apps/Src/RTOSPedals.c
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+* @brief   Initialize both pedals to 0%
+* @param   None
+* @return  None
+*/ 
+static void Pedals_Values_Init(void){
+    accelerator_percentage = 0;
+    brake_percentage = 0;
+}
```
JimothyGreene, line 17: I don't think you'd need a function to initialize this. Since the variables are global, they'll be initialized to 0 by default

Apps/Inc/Tasks.h
ErickCortez98: A struct with the state of the pedals was added in the latest push. There are other two structs that store states, one for the lights and another one for the switches. Hence, we could maybe have a general state struct that will store these three child structs: pedals, lights, and switches, and whichever more state structs we keep adding along the way/?

Apps/Src/RTOSPedals.c
JimothyGreene: The idea with these is to have it only post the semaphores when the pedal position changes

Apps/Inc/StateTypes.h
```diff
@@ -0,0 +1,38 @@
...
+    uint8_t RSVD_LED;
+} LightStates;
+
+typedef struct PedalStates_t{
+    uint8_t ACC_PERCENT;
+    uint8_t BRAKE_PERCENT;
+} PedalStates_t;
+
+
+
```
JimothyGreene, line 38: I think this file is made obsolete by the most recent PR #130

Apps/Src/Tasks.c
```diff
@@ -62,3 +62,4 @@ OS_SEM SendCarCAN_Sem4;
  * Global Variables
  */
 // TODO: Put all global state variables here
+PedalStates_t pedals;
```
JimothyGreene, line 65: Restructure to fit with #130

Apps/Src/RTOSPedals.c
```diff
@@ -0,0 +1,31 @@
...
+    while(1){
+        pedals.ACC_PERCENT = Pedals_Read(ACCELERATOR);
+        pedals.BRAKE_PERCENT = Pedals_Read(BRAKE);
+        // The velocity and lights threads can run without blocking
+        // semaphores, the delay added an the end of this loop will allow
+        // them to run
+
+        // Post to semaphores (velocityChange, lightsChange) 
+        // OSSemPost (&VelocityChange_Sem4, OS_OPT_POST_ALL, &err);
+        // OSSemPost (&LightsChange_Sem4, OS_OPT_POST_ALL, &err);
```
JimothyGreene, line 23: Here, I would keep track of the previous value that this task read and compare if the new value is different from the previous value for either brake or accel. If either is different, then post the semaphores.

JimothyGreene (changes requested):

Apps/Inc/RTOSPedals.h
```diff
@@ -0,0 +1,21 @@
...
+#include "Pedals.h"
+#include "Tasks.h"
+#include "os.h"
+
+/**
+* @brief   Read both pedal percentage values
+* @param   None
+* @return  None
+*/ 
+void Task_ReadPedals(void *p_arg);
```
JimothyGreene, line 18: This isn't needed. Prototype is already defined in `Tasks.h`

Apps/Src/RTOSPedals.c
```diff
@@ -0,0 +1,40 @@
+/* Copyright (c) 2020 UT Longhorn Racing Solar */
+
+#include "RTOSPedals.h"
+
+uint8_t PreviousAccel;
+uint8_t PreviousBrake;
```
JimothyGreene, line 6: These don't need to be global. They can be locally defined like `err` before the while loop

Apps/Src/RTOSPedals.c
```diff
@@ -0,0 +1,40 @@
...
+        
+        // The velocity and lights threads can run without blocking
+        // semaphores, the delay added an the end of this loop will allow
+        // them to run
+
+        // Post to semaphores if there is a change in newly read accel or brake compared to prev values
+        if(carStates->AccelPedalPercent != PreviousAccel || carStates->BrakePedalPercent != PreviousBrake){
+            OSSemPost (&VelocityChange_Sem4, OS_OPT_POST_ALL, &err);
+            OSSemPost (&LightsChange_Sem4, OS_OPT_POST_ALL, &err);
+        }
```
JimothyGreene, line 29: The velocity change only occurs when the Acceleration pedal is changed, and the light change only occurs when the brake pedal is changed. You can separate these two.

JimothyGreene (changes requested):

Cam0Cow (approved):
Seems to work just fine on my end

JimothyGreene (approved):
LGTM

PR #127, opened 2021-02-13
Rtos/activate motor

### LINKED ISSUE #116: [RTOS] Activate Motor Task
This RTOS task is responsible for turning on the motor precharge board before closing the main contactors. The state diagrams and some pseudocode can be found on the Google Drive (Electrical Team > Controls > New Design > Software > RTOS), but don't take either resource as absolute truth. This is the time to reason about this application code, so try to think about what exactly each module should be doing.

Cam0Cow:
Activate motor now spawns the four CAN related tasks when it is run instead of signaling their semaphores. Closes #116


Apps/Src/ActivateMotor.c
```diff
@@ -0,0 +1,118 @@
...
+        (OS_TICK)NULL,
+        (void*)NULL,
+        (OS_OPT)(OS_OPT_TASK_STK_CLR),
+        (OS_ERR*)&err
+    );
+    // TODO: check for task creation error
+
+    while (1) {
+        // Wait until some change needs to be made to the motor state
+        OSSemPend(&MotorConnectionChange_Sem4, 0, OS_OPT_PEND_BLOCKING, &ts, &err);
```
JimothyGreene, line 103: Why is this pending a semaphore? I thought this task was spawned by the Switches task. It should just kill itself after completing its purpose and wait to be spawned again. 

Essentially, this should act as a regular function. The reason I would recommend keeping it as a task, though, is because of the 5 second delay. If this were a regular function call from the `ReadSwitches` task, then that task would be blocked for 5 seconds while it waited for this to finish.

Cam0Cow: We wanted to have the ability to turn the motor on or off, depending on the state of the key. Read Switches checks where the key is, and if it decides that the state of the motor should change, then it posts to the semaphore. If we just killed the task when it was done, there would be no way to turn the motor on again more than once.

JimothyGreene: Well I thought that's what the state variable was for. To communicate to this task whether it should turn it on or off. That doesn't require a semaphore or a blocked task as far as I can see

Cam0Cow: But if the task continuously checked the state variables for changes, it would just chew up cpu time. The semaphore lets it know that something has changed, at which point it can check the state variables and figure out what do. Another option would also be to respawn the task every time we want to turn the motor on, and have a function to turn it off.

JimothyGreene: I'm not saying to continuously check. What I mean is that the Switches task turns this task on when something changes (whether that's off->on or on->off). This task then runs once (no looping). It first checks the state variable to know whether it's turning the motor on or off. Depending on which one, it either spawns the new tasks or kills those tasks and then either turns the motor on or turns it off. No need for a loop or a blocked task.

chaseblock: I think I agree with @Cam0Cow here. A blocked task doesn't take up any resources that we need anyways, and it's going to be quicker to unblock and manipulate the motor state than to spawn a new task every time.

JimothyGreene: Okay well then we shouldn't have it do both. Either the switches task just blocks this one or it spawns it. In its current state it spawns it once and then blocks it.

chaseblock: Maybe interesting to look into using a task semaphore instead (as opposed to a regular semaphore). See pg. 267 here https://www.analog.com/media/en/dsp-documentation/software-manuals/Micrium-uCOS-III-UsersManual.pdf

Cam0Cow: That looks like exactly what we need @chaseblock

Apps/Src/ActivateMotor.c
```diff
@@ -0,0 +1,118 @@
...
+        (OS_MSG_QTY)NULL,
+        (OS_TICK)NULL,
+        (void*)NULL,
+        (OS_OPT)(OS_OPT_TASK_STK_CLR),
+        (OS_ERR*)&err
+    );
+    // TODO: check for task creation error
+
+    // Spawn ReadCarCAN
+    OSTaskCreate(
```
JimothyGreene, line 47: So is this task still being spawned by the ReadSwitches task, or is it just going to be using the Task Semaphore to block/unblock? If it's the latter, these task creations here need to only happen once it's determined that the motor is connected.

Cam0Cow: It's still being spawned by ReadSwitches, but I think it makes sense to spawn it somewhere else now. The real question is whether we should spawn the four CAN tasks here or somewhere else, and have them also wait on a task semaphore.

JimothyGreene: Up to you honestly. Just make sure for the tasks that should run freely once "spawned" that the semaphores aren't blocking on each iteration of the loop.

PR #128, opened 2021-02-20
STM32 CAN BSP

### LINKED ISSUE #121: [BSP] STM32f413 CAN Implementation
Adapt BPS's BSP_CAN code to work with our system.

JimothyGreene:
Adapts the CAN BSP layer from BPS code to Controls Leader. We use pins PA11 and PA12 for CAN RX and TX respectively, so those configurations were modified slightly from BPS to accommodate. 

**Note**: BPS is in the process of modifying its CAN code to avoid busy waits on transmission. It is still not fully functional, so I adapted the known working code for Controls use at this time. If we feel the need later to avoid the busy waits on transmission, we can use the (hopefully working) BPS code at that time.

Closes #121

### DIFF
--- a/BSP/STM32F413/Src/BSP_CAN.c
+++ b/BSP/STM32F413/Src/BSP_CAN.c
@@ -1,6 +1,18 @@
 /* Copyright (c) 2020 UT Longhorn Racing Solar */
 
+/**
+ * Pin Configuration:
+ * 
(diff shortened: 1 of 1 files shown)


JimothyGreene:
Closing until @babusid adapts the new BSP_CAN from BPS

PR #129, opened 2021-02-20
Rtos/send tritium

### LINKED ISSUE #111: [RTOS] Send Tritium Task
This RTOS task is responsible for sending the drive command to the motor controller every 100 ms. The state diagrams and some pseudocode can be found on the Google Drive (Electrical Team > Controls > New Design > Software > RTOS), but don't take either resource as absolute truth. This is the time to reason about this application code, so try to think about what exactly each module should be doing.

Cam0Cow:
Implemented the send tritium task to work with the restructuring of the tasks.
Closes #111

### DIFF
--- a/Apps/Inc/Tasks.h
+++ b/Apps/Inc/Tasks.h
@@ -135,7 +135,5 @@ extern OS_SEM SendCarCAN_Sem4;
  */
 // TODO: Put all global state variables here
 
-extern float RequestedMotorRPM;
-extern float RequestedMotorCurrentPercent;
 
 #endif
--- a/Apps/Inc/Tasks.h
+++ b/Apps/Inc/Tasks.h
@@ -134,4 +134,7 @@ extern OS_SEM SendCarCAN_Sem4;
  */
 // TODO: Put all global state variables here
 
+extern float RequestedMotorRPM;
+extern float RequestedMotorCurrentPercent;
+
 #endif
--- a/Apps/Src/SendTritium.c
+++ b/Apps/Src/SendTritium.c
@@ -1,10 +1,11 @@
 #include "SendTritium.h"
 #include "CarState.h"
+#include <math.h>
 
 
 static float velocity_to_rpm(float velocity) {
-    // TODO: fill this in
-    return velocity;
+    float wheel_rpm = velocity * WHEEL_DIAMETER * M_PI * 2;
+    return wheel_rpm * GEAR_REDUCTION;
 }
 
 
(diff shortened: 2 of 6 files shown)


Apps/Inc/Tasks.h
```diff
@@ -134,4 +134,7 @@ extern OS_SEM SendCarCAN_Sem4;
  */
 // TODO: Put all global state variables here
 
+extern float RequestedMotorRPM;
+extern float RequestedMotorCurrentPercent;
+
```
JimothyGreene, line 139: I would restructure this to go with the new CarState struct

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,21 @@
...
+
+    (void) p_arg;
+
+    OS_ERR err;
+
+    MotorController_Init();
+
+    while (1) {
+        float requested_rpm = RequestedMotorRPM;
+        float requested_current_percent = RequestedMotorCurrentPercent;
```
JimothyGreene, line 13: Is there a need for these intermediary variables?

Apps/Src/Tasks.c
```diff
@@ -62,3 +62,6 @@ OS_SEM SendCarCAN_Sem4;
  * Global Variables
  */
 // TODO: Put all global state variables here
+
+float RequestedMotorRPM; // Requested motor RPM (set by VelocityChange)
+float RequestedMotorCurrentPercent; // Requested motor current as fraction of max (set by VelocityChange)
```
JimothyGreene, line 67: Same as Tasks.h

JimothyGreene (changes requested):

Apps/Inc/Tasks.h
Cam0Cow: The Car State struct currently holds velocity as a uint8_t, but the motor controller wants a float value. Moreover, motor controller drive command requires a value in RPM, not m/s, so I assume those fields in car state have to change

JimothyGreene: Yeah so I would change the data types in the struct to match that. But we're just trying to get rid of these globals

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,28 @@
+#include "SendTritium.h"
+#include "CarState.h"
+#include <math.h>
+
+
+static float velocity_to_rpm(float velocity) {
```
JimothyGreene, line 6: Is the naming convention for static methods snake case? No preference from me really, just make sure it's consistent.

JimothyGreene (approved):
Looks good to me

Apps/Src/SendTritium.c
Cam0Cow: I'm fairly sure that it's the convention that we've mostly been using. At the end of the day it doesn't really matter, the function isn't accessible outside of this file.

ErickCortez98 (approved):

PR #130, opened 2021-02-20
Adds new Car State struct to hold all relevant state information

JimothyGreene:
An instance of this new struct will be a local variable in the `main` function that is passed as part of the `p_arg` parameter to each task that needs it. This way we avoid all the getter functions as well as the problems that come with separate, global, cross-module variables.

### DIFF
--- a/Apps/Inc/CarState.h
+++ b/Apps/Inc/CarState.h
@@ -1,0 +1,64 @@
+/* Copyright (c) 2020 UT Longhorn Racing Solar */
+
+#ifndef __CAR_STATE_H
+#define __CAR_STATE_H
+
+#include "common.h"
+#include "config.h"
+
+/**
+ * Switch States
+ * 
+ * Stores the current state of each of
+ * the switches that control this system
+ */
+typedef struct {
+    State LT;
+    State RT;
+    State FWD;
+    State REV;
+    State CRS_EN;
+    State CRS_SET;
+    State REGEN;
+    State HZD;
+    State HDLT;
+    State IGN_1;
+    State IGN_2;
+} switch_states_t;
+
+/**
+ * Blinker States
+ * 
+ * Stores the desired blinker state
+ * to indicate which lights to toggle
+ */
+typedef struct {
+    State HZD;
+    State LT;
+    State RT;
+} blinker_states_t;
+
+/**
+ * Car State
+ * 
+ * Stores all relevant data about the current
+ * state of the car
+ * This is used by all the tasks to communicate
+ */
+typedef struct {
+    uint8_t DesiredVelocity;
+    uint8_t CruiseControlVelocity;
+    uint8_t CurrentVelocity;
+
+    uint8_t AccelPedalPercent;
+    uint8_t BrakePedalPercent;
+    uint8_t MotorCurrentSetpoint;
+
+    switch_states_t SwitchStates;
+    blinker_states_t BlinkerStates;
+
+    State CruiseControlEnable;
+    State CruiseControlSet;
+} car_state_t;
+
+#endif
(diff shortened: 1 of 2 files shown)


babusid:
Are global states not needed for all of the lights? Why are only the blinkers included?

JimothyGreene:
> Are global states not needed for all of the lights? Why are only the blinkers included?

There's only one thread that will ever be messing with the main lights, plus none of the tasks rely on being able to read the state of the lights. The Blinker lights are meant as an indication to the `BlinkLight` task of _which_ light it should be blinking.

Cam0Cow (approved):
Looks good to me. At the moment I can't think of anything else that has to be added to the struct, but it can always be expanded later.

SoroushF79 (approved):

babusid:
I had just one more comment/question. Would it be possible to make the switchstates match the structure of the switchstates enum from the switches driver? Something like:

JimothyGreene:
> I had just one more comment/question. Would it be possible to make the switchstates match the structure of the switchstates enum from the switches driver? Something like:
> 
> typedef struct switch_states_t {
> uint8_t CRUZ_SW;
> uint8_t CRUZ_EN;
> uint8_t HZD_SQ;
> uint8_t FR_SW;
> uint8_t HEADLIGHT_SW;
> uint8_t LEFT_SW;
> uint8_t RIGHT_SW;
> uint8_t REGEN_SW;
> uint8_t IGN_1;
> uint8_t IGN_2;
> uint8_t REV_SW;
> } switch_states_t;

We should definitely use the `State` type for each of these to keep it consistent. In terms of the naming, we can append the _SW to them. But also with this, I think the enum needs to be updated to have a separate FW and REV switches since they are separate inputs now to accommodate for our neutral gear.

babusid:
> > I had just one more comment/question. Would it be possible to make the switchstates match the structure of the switchstates enum from the switches driver? Something like:
> > typedef struct switch_states_t {
> > uint8_t CRUZ_SW;
> > uint8_t CRUZ_EN;
> > uint8_t HZD_SQ;
> > uint8_t FR_SW;
> > uint8_t HEADLIGHT_SW;
> > uint8_t LEFT_SW;
> > uint8_t RIGHT_SW;
> > uint8_t REGEN_SW;
> > uint8_t IGN_1;
> > uint8_t IGN_2;
> > uint8_t REV_SW;
> > } switch_states_t;
> 
> We should definitely use the `State` type for each of these to keep it consistent. In terms of the naming, we can append the _SW to them. But also with this, I think the enum needs to be updated to have a separate FW and REV switches since they are separate inputs now to accommodate for our neutral gear.

I think the enum in the switches driver already has a discrete FR_SW and REV_SW.

JimothyGreene:
> > > I had just one more comment/question. Would it be possible to make the switchstates match the structure of the switchstates enum from the switches driver? Something like:
> > > typedef struct switch_states_t {
> > > uint8_t CRUZ_SW;
> > > uint8_t CRUZ_EN;
> > > uint8_t HZD_SQ;
> > > uint8_t FR_SW;
> > > uint8_t HEADLIGHT_SW;
> > > uint8_t LEFT_SW;
> > > uint8_t RIGHT_SW;
> > > uint8_t REGEN_SW;
> > > uint8_t IGN_1;
> > > uint8_t IGN_2;
> > > uint8_t REV_SW;
> > > } switch_states_t;
> > 
> > 
> > We should definitely use the `State` type for each of these to keep it consistent. In terms of the naming, we can append the _SW to them. But also with this, I think the enum needs to be updated to have a separate FW and REV switches since they are separate inputs now to accommodate for our neutral gear.
> 
> I think the enum in the switches driver already has a discrete FR_SW and REV_SW.

Ah yes sorry. Should probably change the `FR_SW` to just a `FWD_SW`

PR #131, opened 2021-02-27
Read tritium task

SoroushF79:
Some parts I wasn't sure how to deal with, like error handling.

### DIFF
--- a/Apps/Src/ReadTritium.c
+++ b/Apps/Src/ReadTritium.c
@@ -10,16 +10,19 @@ void Task_ReadTritium(void* p_arg) {
 
 	while (1) {
 
-		uint8_t *message;
+		uint8_t message[8];
 		ErrorStatus status = CANbus_Read(message);
+		printf("got data\n");
 
 		if(status == SUCCESS) {
 
-			OSQPost(&CANBus_MsgQ, (void *)message, strlen((char *) message), OS_OPT_POST_FIFO, &err);
+			// Is using strlen here correct?
+			OSQPost(&CANBus_MsgQ, (void *)message, 8, OS_OPT_POST_FIFO, &err);
 			
 			if (err != OS_ERR_NONE) {
 				// TODO: error handling
 			}
 		}
+		OSTimeDlyHMSM(0, 0, 0, 1000, OS_OPT_TIME_HMSM_NON_STRICT, &err);
 	}
 }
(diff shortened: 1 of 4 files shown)


Apps/Src/ReadTritium.c
```diff
@@ -0,0 +1,36 @@
...
+	CPU_TS ts;
+
+	OSSEMPend(&READTritium_Sem4, 0, OS_OPT_PEND_BLOCKING, &ts, &err);
+
+	CANbus_Init();
+
+	uint8_t *message;
+	ErrorStatus status = CANbus_Read(message);
+
+	if(message){
```
JimothyGreene, line 17: We shouldn't be conditioning this on `message` since that's just a pointer. I think it should be conditioned on the `status`

Apps/Src/ReadTritium.c
```diff
@@ -0,0 +1,36 @@
+/* Copyright (c) 2021 UT Longhorn Racing Solar */
+
+#include "ReadTritium.h"
+
+void Task_ReadTritium(void* p_arg){
+
+	OS_ERR err;
+	CPU_TS ts;
+
+	OSSEMPend(&READTritium_Sem4, 0, OS_OPT_PEND_BLOCKING, &ts, &err);
```
JimothyGreene, line 10: Check the capitalization here

Apps/Src/ReadTritium.c
```diff
@@ -0,0 +1,36 @@
...
+		switch(err){
+			case OS_Q_FULL:
+				break;
+			case OS_ERR_EVENT_TYPE:
+				break;
+			case OS_ERR_PEVENT_NULL:
+				break;
+			case OS_ERR_POST_NULL_PTR:
+				// This shouldn't happen since we check if message was null earlier
+				break;
```
JimothyGreene, line 31: Make sure all of these are valid Error cases and add in a `default` case to get rid of the mess of warnings during compilation.

JimothyGreene (changes requested):

Apps/Src/ReadTritium.c
```diff
@@ -0,0 +1,25 @@
...
+
+void Task_ReadTritium(void* p_arg) {
+
+	OS_ERR err;
+
+	CANbus_Init();
+
+	while (1) {
+
+		uint8_t *message;
```
JimothyGreene, line 13: This needs to be an array so it allocates space to actually store the values. Max size I believe is 8 bytes so it can just be `uint8_t message[8]`

JimothyGreene (changes requested):

JimothyGreene (approved):
LGTM

PR #132, opened 2021-02-27
Uart update

Cam0Cow:
Adapts BPS's UART code to work with the existing Controls codebase

### DIFF
--- a/BSP/STM32F413/Src/BSP_UART.c
+++ b/BSP/STM32F413/Src/BSP_UART.c
@@ -2,6 +2,7 @@
 
 #include "BSP_UART.h"
 #include "stm32f4xx.h"
+#include "os.h"
 
 #define TX_SIZE     128
 #define RX_SIZE     64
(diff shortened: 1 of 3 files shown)


BSP/STM32F413/Src/BSP_UART.c
```diff
@@ -6,60 +6,47 @@
...
-static void BSP_UART2_Init(void) {
     GPIO_InitTypeDef GPIO_InitStruct = {0};
     USART_InitTypeDef UART_InitStruct = {0};
 
-    Fifo_Init(&fifos[UART_2]);
-
+    // Initialize clocks
     RCC_APB1PeriphClockCmd(RCC_APB1Periph_USART2, ENABLE);
-    RCC_AHB1PeriphClockCmd(RCC_AHB1Periph_GPIOA, ENABLE);
+    RCC_APB1PeriphClockCmd(RCC_AHB1Periph_GPIOA, ENABLE);
```
JimothyGreene, line 47: I'm not sure this change is right

BSP/STM32F413/Src/BSP_UART.c
```diff
@@ -172,277 +111,213 @@ static void BSP_UART3_Init() {
...
 }
 
 void USART2_IRQHandler(void) {
-    UartFifo_t *fifo = &fifos[UART_2];
+#ifdef RTOS
+    CPU_SR_ALLOC();
+    CPU_CRITICAL_ENTER();
+    OSIntEnter();
+    CPU_CRITICAL_EXIT();
+#endif
```
JimothyGreene, line 237: BPS has these ifdef guards because they write the software for bare metal as well. Controls isn't doing this, and we don't define RTOS in our build configuration, so make sure to remove these guards wherever they appear.

JimothyGreene (changes requested):
Controls isn't using the exact same configuration as BPS. It's similar but not the same. Make sure to make the appropriate changes to account for the differences.

JimothyGreene (approved):
Looks good. Again, assuming BPS implementation works, which is should.

PR #133, opened 2021-02-27
RTOS_SwitchRead

### LINKED ISSUE #108: [RTOS] Read Switches Task
This RTOS task should read all of the input switches and set the car state accordingly. The state diagrams and some pseudocode can be found on the Google Drive (Electrical Team > Controls > New Design > Software > RTOS), but don't take either resource as absolute truth. This is the time to reason about this application code, so try to think about what exactly each module should be doing.

babusid:
Implemented the Read Switches task with the new task structure. Closes #108


Apps/Inc/CarState.h
```diff
@@ -12,19 +12,21 @@
...
-    State CRS_SET;
-    State REGEN;
-    State HZD;
-    State HDLT;
+
+typedef struct switch_states_t {
+    State CRUZ_SW;
+    State CRUZ_EN;
+    State HZD_SQ;
+    State FR_SW;
```
JimothyGreene, line 20: Make this `FWD_SW` just to avoid confusion

Apps/Inc/CarState.h
```diff
@@ -12,19 +12,21 @@
...
-    State CRS_EN;
-    State CRS_SET;
-    State REGEN;
-    State HZD;
-    State HDLT;
+
+typedef struct switch_states_t {
+    State CRUZ_SW;
+    State CRUZ_EN;
+    State HZD_SQ;
```
JimothyGreene, line 19: This should be `HZD_SW`

Apps/Inc/StateTypes.h
```diff
@@ -0,0 +1,31 @@
...
+    uint8_t CTRL_FAULT;
+    uint8_t LEFT_BLINK;
+    uint8_t RIGHT_BLINK;
+    uint8_t Headlight_ON;
+    uint8_t BPS_FAULT;
+    uint8_t BPS_PWR;
+    uint8_t  BrakeLight;
+    uint8_t RSVD_LED;
+} LightStates;
+
```
JimothyGreene, line 31: We don't need this file anymore

Apps/Inc/Tasks.h
```diff
@@ -135,4 +135,7 @@ extern OS_SEM SendCarCAN_Sem4;
  */
 // TODO: Put all global state variables here
 
+extern car_state_t carState;
+
+
```
JimothyGreene, line 140: This doesn't need to be global. It'll be passed to each task in `main()` as a `p_arg`

Drivers/Inc/Switches.h
```diff
@@ -6,7 +6,19 @@
 #include "config.h"
 
 
-typedef enum {CRUZ_SW=0, CRUZ_EN, HZD_SQ, FR_SW, HEADLIGHT_SW, LEFT_SW, RIGHT_SW, REGEN_SW, IGN_1, IGN_2, REV_SW} switches_t;
+typedef enum {
+    CRUZ_SW=0,
+    CRUZ_EN,
+    HZD_SQ, 
+    FR_SW,
```
JimothyGreene, line 13: Same comments as above

Apps/Src/Tasks.c
```diff
@@ -62,3 +63,6 @@ OS_SEM SendCarCAN_Sem4;
  * Global Variables
  */
 // TODO: Put all global state variables here
+
+car_state_t carState;
+
```
JimothyGreene, line 68: Same as above

Apps/Src/ReadSwitches.c
```diff
@@ -0,0 +1,167 @@
...
+    Switches_Init();
+    OS_ERR err;
+    while(1){
+        //Spawns arrayConnection thread and MotorConnection thread when it is appropriate
+        uint8_t ignStates = 0x00; //holds states read from Switches Driver
+        uint8_t ignStored = 0x00; //holds states stored in globals
+        ignStates = ignStates || (Switches_Read(IGN_2)<<1);
+        ignStates = ignStates || (Switches_Read(IGN_1));
+        ignStored = ignStored || ((carState.SwitchStates.IGN_2&&0x01)<<1);
+        ignStored = ignStored || (carState.SwitchStates.IGN_1&&0x01);
```
JimothyGreene, line 20: You're overwriting these variables immediately after writing to them

Apps/Src/ReadSwitches.c
```diff
@@ -0,0 +1,167 @@
+//#include "Tasks.h"
+#include <os.h>
+#include "Switches.h"
+#include "Tasks.h"
+
+/**
+ * @brief Reads Switches using Switches Driver and signals appropriate
+ * Semaphores when Switches state changes. Also updates a switches global defined in StateTypes.h, declared in Tasks.c and externed in Tasks.h
+*/
+void Task_ReadSwitches (void* p_arg) {
```
JimothyGreene, line 10: The logic here makes sense. A recommendation:

Instead of reading the switches grouped on how they're used, read all switches up front and update the `carState` struct. Then, you can go through each of the groupings and just read directly from the `carState` struct. That way, the `carState` values are always properly up to date.

JimothyGreene (changes requested):
**Note:** I wouldn't merge this in until the ArrayConnection and MotorConnection tasks have been tested and merged. We want our `master` code to compile.

Apps/Src/ReadSwitches.c
babusid: Don't the semaphores have to get signaled only if there is a difference between the prior car state and the new car state? If I overwrote the prior car state with the new values at the top, how would I know whether to signal or whether the switch had been on forever?

JimothyGreene (dismissed):
Couple notes left

babusid:
@Cam0Cow Do you recall why this hasn't been merged in just yet? Is it because we can't test it without the board?

Cam0Cow:
> @Cam0Cow Do you recall why this hasn't been merged in just yet? Is it because we can't test it without the board?

Yes, it didn't make sense to merge it into master untested.

babusid:
This Branch does not have the BSP CAN layer at present and as such is failing build checks. We need to merge in the CAN BSP layer into master, merge master into this branch, and test before we can merge this.

babusid:
@Cam0Cow @chaseblock This is failing compilation when ` make stm32f413 TEST=SwitchRead` is run but not when ` make stm32f413 ` is run. Could either of you run the same thing and see if you catch any errors strictly in terms of compilation?

ClarkPoon:
> @Cam0Cow @ClarkPoon This is failing compilation when ` make stm32f413 TEST=SwitchRead` is run but not when `make stm32f413` is run. Could either of you run the same thing and see if you catch any errors strictly in terms of compilation?

BSP_CAN is not implemented on this branch, so it can't find the BSP_CAN functions declared in BSP_CAN.h when it tries to link.

https://github.com/lhr-solar/Controls/blob/RTOS_SwitchRead/BSP/STM32F413/Src/BSP_CAN.C

babusid:
> > @Cam0Cow @ClarkPoon This is failing compilation when ` make stm32f413 TEST=SwitchRead` is run but not when `make stm32f413` is run. Could either of you run the same thing and see if you catch any errors strictly in terms of compilation?
> 
> 
> 
> BSP_CAN is not implemented on this branch, so it can't find the BSP_CAN functions declared in BSP_CAN.h when it tries to link.
> 
> 
> 
> https://github.com/lhr-solar/Controls/blob/RTOS_SwitchRead/BSP/STM32F413/Src/BSP_CAN.C

Ah, these are the things you miss when you forget to sleep. I'll add dummy functions to the C file since this thread doesn't depend on CAN.

Apps/Src/ReadSwitches.c
```diff
@@ -0,0 +1,225 @@
...
+                    OSSemPost(
+                        &ArrayConnectionChange_Sem4,
+                        (OS_OPT) OS_OPT_POST_ALL,
+                        (OS_ERR*)&err
+                    );
+                    OSSemPost(
+                        &MotorConnectionChange_Sem4,
+                        (OS_OPT) OS_OPT_POST_ALL,
+                        (OS_ERR*)&err
+                    );
```
ClarkPoon, line 67: Is there a reason you are posting semaphores to control the contactors instead of just calling the contactor driver directly? It would get rid of the context switch overhead and reduce the number of tasks you need. I also think it would make the code more readable and easier to debug.

Apps/Src/ReadSwitches.c
```diff
@@ -0,0 +1,225 @@
...
+        
+        if(lightSwitchStored!=lightSwitchStates){
+            //if any of the bits dont match, CarState needs to get updated and LightsChange must get signaled
+            //TODO: use memcpy to move currSwitchStates.lightSwitches into globalSwitchStates->lightSwitches
+            memcpy(&(globalSwitchStates->lightSwitches),&(currSwitchStates.lightSwitches),sizeof(currSwitchStates.lightSwitches));
+            OSSemPost(
+                &LightsChange_Sem4,
+                (OS_OPT) OS_OPT_POST_ALL,
+                (OS_ERR*)&err
+            );
```
ClarkPoon, line 183: Why would you not just make a function to set the states of the different lights and call that? I don't see why you need to populate a global data structure and post a semaphore to make a task run to change the lights.

Apps/Src/ReadSwitches.c
```diff
@@ -0,0 +1,225 @@
...
+            OSSemPost(
+                &DisplayChange_Sem4,
+                (OS_OPT) OS_OPT_POST_ALL,
+                (OS_ERR*)&err
+            );
+            OSSemPost(
+                &VelocityChange_Sem4,
+                (OS_OPT) OS_OPT_POST_ALL,
+                (OS_ERR*)&err
+            );
```
ClarkPoon, line 219: Why not use a function?

Drivers/Src/Switches.c
```diff
@@ -1,6 +1,9 @@
 #include "Switches.h"
 
+#define IODIRA_INIT 0xFF
+#define IODIRB_INIT 0x80
```
ClarkPoon, line 4: These names don't seem very intuitive to me

Drivers/Src/Switches.c
```diff
@@ -20,9 +23,10 @@ void Switches_Init(void){
     BSP_SPI_Write(initTxBuf,3);
     do{
         BSP_SPI_Read(initRxBuf, 2);
-    }while(initRxBuf[1] == SPI_IODIRA);
+    }while(initRxBuf[1] == SPI_IODIRA); //keep reading until we get a result that's not what was written by line 20
+
     //OR Result of IODIRA read to set all to 1, then write it back to IODIRA
-    initTxBuf[2] = initRxBuf[1]|0xFF;
+    initTxBuf[2] = initRxBuf[1]|IODIRA_INIT;
```
ClarkPoon, line 29:
```suggestion
    initTxBuf[2] = initRxBuf[1] | IODIRA_INIT;
```

Drivers/Src/Switches.c
```diff
@@ -35,7 +39,7 @@ void Switches_Init(void){
         BSP_SPI_Read(initRxBuf, 2);
     }while(initRxBuf[1] == SPI_IODIRB);
     //OR IODIRB to set pin 7 to input and write it back
-    initTxBuf[2] = initRxBuf[1]|0x80;
+    initTxBuf[2] = initRxBuf[1]|IODIRB_INIT;
```
ClarkPoon, line 42:
```suggestion
    initTxBuf[2] = initRxBuf[1] | IODIRB_INIT;
```

Tests/Test_SwitchRead.c
```diff
@@ -0,0 +1,96 @@
...
+#include "config.h"
+#include <os.h>
+#include "Tasks.h"
+
+/**
+ * This file is a basic test file that spawns the readSwitches thread, 
+ * and another thread that prints those states to the console. 
+ * 
+*/
+static car_state_t CarState;
```
ClarkPoon, line 11: What is the point of declaring this `static` if you just pass a pointer to it into literally all of your tasks?

Apps/Src/ReadSwitches.c
```diff
@@ -0,0 +1,225 @@
...
+                            (CPU_CHAR*)"Array Connection",
+                            (OS_TASK_PTR)Task_ArrayConnection,
+                            (void*)NULL,
+                            (OS_PRIO)TASK_ARRAY_CONNECTION_PRIO,
+                            (CPU_STK*)ArrayConnection_Stk,
+                            (CPU_STK_SIZE)sizeof(ArrayConnection_Stk)/10,
+                            (CPU_STK_SIZE)sizeof(ArrayConnection_Stk),
+                            (OS_MSG_QTY)NULL,
+                            (OS_TICK)NULL,
+                            (void*)NULL,
```
babusid, line 84: This task needs the CarState struct pointer

Apps/Src/ReadSwitches.c
```diff
@@ -0,0 +1,225 @@
...
```
babusid, line 112: This task also needs the CarState struct pointer

Apps/Src/ReadSwitches.c
```diff
@@ -0,0 +1,225 @@
...
+                            (CPU_CHAR*)"Motor Connection",
+                            (OS_TASK_PTR)Task_MotorConnection,
+                            (void*)NULL,
+                            (OS_PRIO)TASK_MOTOR_CONNECTION_PRIO,
+                            (CPU_STK*)ArrayConnection_Stk,
+                            (CPU_STK_SIZE)sizeof(MotorConnection_Stk)/10,
+                            (CPU_STK_SIZE)sizeof(MotorConnection_Stk),
+                            (OS_MSG_QTY)NULL,
+                            (OS_TICK)NULL,
+                            (void*)NULL,
```
babusid, line 136: Same as above

Apps/Src/ReadSwitches.c
```diff
@@ -0,0 +1,136 @@
...
+#include "ReadSwitches.h"
+
+// Macros
+#define OS_DELAY_MS(ms) \
+    OSTimeDlyHMSM(0, 0, 0, (ms), OS_OPT_TIME_HMSM_NON_STRICT, &err)
+#define MAX_SWITCH_LAG          100  // delay (ms) for checking switches while precharging
+#define PRECHARGE_DELAY_CYCLES  (PRECHARGE_ARRAY_DELAY*1000 / MAX_SWITCH_LAG)
+
+// Globals
+switch_states_t CurrSwitches;
```
IshDeshpa, line 13: A bitmap that serves this same purpose exists in the BSP layer. Instead of using this to store the current switch states, use Switch_Read to access the bitmap.

Apps/Src/ReadSwitches.c
```diff
@@ -0,0 +1,136 @@
...
+    Switches_UpdateStates();
+
+    CurrSwitches.IGN_1 = Switches_Read(IGN_1);
+    CurrSwitches.IGN_2 = Switches_Read(IGN_2);
+    
+    CurrSwitches.LT = Switches_Read(LEFT_SW);
+    CurrSwitches.RT = Switches_Read(RIGHT_SW);
+    CurrSwitches.HDLT = Switches_Read(HEADLIGHT_SW);
+    CurrSwitches.HZD = Switches_Read(HZD_SW);
+
```
IshDeshpa, line 120: Instead of updating a new CurrSwitches in this file, implementing a Switches_ReadBitmap function in the BSP layer might be a good idea. Or just use Switches_Read whenever necessary.

IshDeshpa (changes requested):
- CurrSwitches is redundant because the switches bitmap exists on the BSP layer. 
- A Switches_ReadBitmap function on the BSP layer might be necessary.
- ArrayPrecharge is no longer a necessary state in the FSM.
- Use State struct correctly

Apps/Src/ReadSwitches.c
```diff
@@ -0,0 +1,136 @@
...
+     * indexed by ignition switch read value (0, 1, or 3)
+     *      0: electronics on only
+     *      1: delay for array precharge -> array on
+     *      3: motor on
+     */
+    uint8_t next_state[4];
+
+    void (* statefunction)();   // function executed at state
+    uint32_t delay_ms;          // ms delay after executing function, should be >= 1
+} fsm_state_t;
```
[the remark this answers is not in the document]
tiandahuang: These functions and structs are used only for this task so they should probably stay within the .c

Apps/Src/ReadSwitches.c
```diff
@@ -0,0 +1,136 @@
...
+    CurrSwitches.HDLT = Switches_Read(HEADLIGHT_SW);
+    CurrSwitches.HZD = Switches_Read(HZD_SW);
+
+    UpdateLights();
+
+    return (CurrSwitches.IGN_2 << 1) + CurrSwitches.IGN_1;
+}
+
+static void UpdateLights() {
+    Lights_Set(Headlight_ON, CurrSwitches.HDLT);
```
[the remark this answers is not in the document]
tiandahuang: CurrSwitches.HDLT gets its value from Read_Switches, so it'll be a valid State (ON/OFF).

Apps/Src/ReadSwitches.c
```diff
@@ -0,0 +1,136 @@
...
+    
+    int leftblink = CurrSwitches.LT | 
+                    CurrSwitches.HZD;
+    int rightblink = CurrSwitches.RT | 
+                     CurrSwitches.HZD;
+
+    Toggle_Set(RIGHT_BLINK, rightblink);
+    Toggle_Set(LEFT_BLINK, leftblink);
```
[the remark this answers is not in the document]
tiandahuang: left/right blink gets its value from Read_Switches, so it'll be a valid State (ON/OFF)

Apps/Src/ReadSwitches.c
```diff
@@ -0,0 +1,136 @@
...
+     */
+    uint8_t next_state[4];
+
+    void (* statefunction)();   // function executed at state
+    uint32_t delay_ms;          // ms delay after executing function, should be >= 1
+} fsm_state_t;
+
+fsm_state_t IgnitionFSM[] = {
+    {{ELEC_ON, ARRAY_PRE,   UNUSED_SWITCH_NUM, ARRAY_PRE},  ElectronicsOn,  10}, // electronics on, array/motor off
+    {{ELEC_ON, ARRAY_ON,    UNUSED_SWITCH_NUM, ARRAY_ON},   ArrayPrecharge, 10}, // array precharge, array/motor off
```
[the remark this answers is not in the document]
tiandahuang: I think we still need to precharge the array

Apps/Src/ReadSwitches.c
tiandahuang: This was written before we had that bitmap; I've updated it to use Switches_Read 😄👍

Apps/Src/ReadSwitches.c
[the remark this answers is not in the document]
IshDeshpa: You are right, I was mistaken.

Apps/Src/ReadSwitches.c
```diff
@@ -0,0 +1,128 @@
...
+    while ((i < PRECHARGE_DELAY_CYCLES) && (UpdateSwitches() != 0)) {
+        OS_DELAY_MS(MAX_SWITCH_LAG);
+        i++;
+    }
+}
+
+static void ArrayOn() {
+    Contactors_Set(MOTOR_CONTACTOR, OFF);
+    Contactors_Set(ARRAY_PRECHARGE, OFF);
+    Contactors_Set(ARRAY_CONTACTOR, ON);
```
babusid, line 102: Reverse the order of these two, the point of having the array precharge is to reduce current spikes coming into the system. If we disconnect array Precharge before connecting the main contactor, we have a small period of 0 current, which sort of negates the point. While it's not a major concern, flipping them will be more in line with the designed behavior.
```suggestion
    Contactors_Set(ARRAY_CONTACTOR, ON);
    Contactors_Set(ARRAY_PRECHARGE, OFF);
```

Apps/Src/ReadSwitches.c
```diff
@@ -0,0 +1,128 @@
+/* Copyright (c) 2020 UT Longhorn Racing Solar */
+
+// Includes
+#include "ReadSwitches.h"
+
+// Macros
+#define OS_DELAY_MS(ms) \
+    OSTimeDlyHMSM(0, 0, 0, (ms), OS_OPT_TIME_HMSM_NON_STRICT, &err)
+#define MAX_SWITCH_LAG          100  // delay (ms) for checking switches while precharging
```
babusid, line 9: This is just the interval we want to check the ignition switch for changes during array precharge right?

Apps/Src/ReadSwitches.c
```diff
@@ -0,0 +1,128 @@
...
+    Lights_Init();
+    Contactors_Init();
+
+    // Delay for precharge
+    OSTimeDlyHMSM(0, 0, PRECHARGE_MOTOR_DELAY, 0, OS_OPT_TIME_HMSM_NON_STRICT, &err);
+
+    // Main loop
+    while (1) {
+        curr_fsm_state->statefunction();
+        OS_DELAY_MS(curr_fsm_state->delay_ms);
```
babusid, line 76: not a fan of this macro still

Apps/Src/ReadSwitches.c
```diff
@@ -0,0 +1,128 @@
...
+    Contactors_Set(MOTOR_CONTACTOR, OFF);
+    return;
+}
+
+static void ArrayPrecharge() {
+    Contactors_Set(ARRAY_PRECHARGE, ON);
+
+    uint16_t i = 0;
+    while ((i < PRECHARGE_DELAY_CYCLES) && (UpdateSwitches() != 0)) {
+        OS_DELAY_MS(MAX_SWITCH_LAG);
```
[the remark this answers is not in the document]
babusid: Upon re-reviewing, I see the global on line fifteen, but I have a strong feeling that down the line someone will copy this macro for usage elsewhere, and forget the global error variable that it needs. Let's just use the standard function.

Apps/Src/ReadSwitches.c
```diff
@@ -0,0 +1,128 @@
...
+    Contactors_Set(ARRAY_CONTACTOR, OFF);
+    Contactors_Set(MOTOR_CONTACTOR, OFF);
+    return;
+}
+
+static void ArrayPrecharge() {
+    Contactors_Set(ARRAY_PRECHARGE, ON);
+
+    uint16_t i = 0;
+    while ((i < PRECHARGE_DELAY_CYCLES) && (UpdateSwitches() != 0)) {
```
babusid, line 93: If UpdateSwitches returns a zero, this function exits immediately. I believe the intended behavior of this is to exit the Array Precharge state and return to Electronics On. 
However, on Line 76, it delays a further ten milliseconds (based on the FSMs delay field of the precharge state). If during this small delay the switch gets turned to position three or back to position one (ie. someone turns it to zero and immediately back to 1 or three), line 77 of the FSM driver loop could potentially turn the array on before the precharge has actually finished.
I would suggest simply removing the 10-millisecond delay (or changing it to 1 ms) on Line 50 of the FSM table to solve this. While it is an edge case, I just want to bring it in line with the intended behavior. Let me know if I'm misinterpreting/missing something here.

babusid (commented):
Also, can you make a task test file that spawns the ReadSwitches task so that we can test the task itself?

Apps/Src/ReadSwitches.c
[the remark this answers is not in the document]
babusid: I agree with Ishan here, just for the sake of not casting when you don't have to, but it shouldn't be an issue.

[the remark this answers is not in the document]
tiandahuang: Enums are int by default so there's not any actual casting going on. Also, I'm doing bitwise operations to yield left/right blink and it's probably best not to do bitwise operations on an enum.

Tests/Test_ReadSwitches.c
```diff
@@ -0,0 +1,133 @@
...
+}
+
+void TestReadSwitches(void *p_arg) {
+    OS_ERR err;
+
+    CPU_Init();
+    OS_CPU_SysTickInit(SystemCoreClock / (CPU_INT32U) OSCfg_TickRate_Hz);
+
+    OSTaskCreate(
+        (OS_TCB*)&ReadTritium_TCB,
```
babusid, line 96: Wrong TCB

Tests/Test_Switches.c
```diff
@@ -10,8 +10,8 @@ int main() {
 
         printf("%d\t", Switches_Read(CRUZ_SW));
         printf("%d\t", Switches_Read(CRUZ_EN));
-        printf("%d\t", Switches_Read(HZD_SQ));
-        printf("%d\t", Switches_Read(FR_SW));
+        printf("%d\t", Switches_Read(HZD_SW));
+        printf("%d\t", Switches_Read(FWD_SW));
```
babusid, line 14: FWD_SW isn't defined, I think it got replaced with FOR_SW

Tests/Test_ReadSwitches.c
```diff
@@ -0,0 +1,133 @@
...
+        (CPU_STK_SIZE)DEFAULT_STACK_SIZE,
+        (OS_MSG_QTY)0,
+        (OS_TICK)0,
+        (void*)NULL,
+        (OS_OPT)(OS_OPT_TASK_STK_CLR),
+        (OS_ERR*)&err
+    );
+
+    // Task not created
+    if (err != OS_ERR_NONE) {
```
babusid, line 79: If we hit an OS error anywhere the program should ideally hang in a while loop or something so that it doesn't keep trying to execute and makes tracing the source more annoying.

babusid (commented):
Not many changes needed, couple quick things in the testfiles, and making OS errors hang in the location that they happen. Can't merge this in before Minions is merged in however, and once Minion driver is merged into master, master needs to get merged into this branch.

Apps/Src/ReadSwitches.c
```diff
@@ -0,0 +1,168 @@
+/* Copyright (c) 2022 UT Longhorn Racing Solar */
+
+// Includes
+#include "ReadSwitches.h"
+
+// Macros
+#define MAX_SWITCH_LAG          100  // period (ms) for checking switches while precharging
+#define READ_SWITCH_PERIOD      20   // period (ms) that switches will be read.
```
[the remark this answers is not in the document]
tiandahuang: update switches gets called at every state, so if it loops at one state (like electronics on), then the delay there is the period that switches are updated

Apps/Src/Tasks.c
```diff
@@ -48,6 +48,7 @@ OS_Q CANBus_MsgQ;
 /**
  * Semaphores
  */
+OS_SEM FaultState_Sem4;
```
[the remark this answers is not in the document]
tiandahuang: Just took a look in master and I can't find it anywhere, should it be located in Tasks.c?

Cam0Cow (approved):

PR #135, opened 2021-03-06
Blinker light task

ErickCortez98:
Implemented BlinkLights task which is spawned by the ReadSwitches


Apps/Inc/BlinkerLight.h
```diff
@@ -0,0 +1,21 @@
...
+
+#include "Lights.h"
+
+
+/**
+* @brief   Toggle proper indicator lights
+* @param   None
+* @return  None
+*/ 
+void Task_BlinkLight(void *p_arg);
```
JimothyGreene, line 18: This is already defined in `Tasks.h`

Apps/Src/BlinkerLight.c
```diff
@@ -0,0 +1,37 @@
...
+* @param   None
+* @return  None
+*/ 
+void Task_BlinkLight(void *p_arg){
+    car_state_t *carStates = ((car_state_t*)p_arg);
+    
+    OS_ERR err;
+    CPU_TS ts;
+
+    while(true){
```
JimothyGreene, line 16: I would just make this a `while (1)` for consistency's sake

Apps/Src/BlinkerLight.c
```diff
@@ -0,0 +1,37 @@
...
+        }else if(carStates->BlinkerStates.LT){
+            // Only the left light blinks
+            Lights_Set(LEFT_BLINK, !Lights_Read(LEFT_BLINK));
+            Lights_Set(RIGHT_BLINK, 0);
+
+        }else if(carStates->BlinkerStates.RT){
+            // Only the right light blinks
+            Lights_Set(RIGHT_BLINK, !Lights_Read(RIGHT_BLINK));
+            Lights_Set(LEFT_BLINK, 0);
+        }
```
JimothyGreene, line 31: Logic seems good to me

JimothyGreene (changes requested):
Just a few small changes. Also, to keep our PRs small, make sure that you branch off of `master` so that changes from another unmerged branch don't pop up in this PR. Just a note for future

Cam0Cow (dismissed):

Tests/Test_BlinkerLight.c
```diff
@@ -0,0 +1,55 @@
...
+        (OS_PRIO)6,
+        (CPU_STK*)ReadSwitches_Stk,
+        (CPU_STK_SIZE)WATERMARK_STACK_LIMIT/10,
+        (CPU_STK_SIZE)TASK_READ_SWITCHES_STACK_SIZE,
+        (OS_MSG_QTY)NULL,
+        (OS_TICK)NULL,
+        (void*)NULL,
+        (OS_OPT)(OS_OPT_TASK_STK_CLR),
+        (OS_ERR*)&err
+    );
```
JimothyGreene, line 25: It doesn't seem like you're passing in the global car state struct anywhere. Is it declared at all anywhere?

JimothyGreene (changes requested):
Looks good for the most part. Make sure you're passing in the global car state to all the tasks, otherwise they'll run into some null pointer issues. Also, I wouldn't merge this in until the ReadSwitches task has been merged in as well. Without that working, we can't fully test this.

babusid:
I think the newer BlinkLights PR #168 may be more suitable as it's just a bit more current. That one is in line with some planned updates to the car state struct that I don't think this one has. @Cam0Cow do you have any additional insight? I'm going to close this for now, and we can open it back up if we need to.

PR #136, opened 2021-03-13
implemented UpdateVelocity task

ErickCortez98:


Apps/Src/UpdateVelocity.c
```diff
@@ -0,0 +1,36 @@
...
+    //Set desired velocity to 0
+    car_state->DesiredVelocity = 0;
+
+    while(1){
+        //Read global pedal percentage
+        uint8_t pedalPercentage = car_state->AccelPedalPercent;
+
+        if(car_state->CruiseControlEnable){
+            if(car_state->CruiseControlSet){
+                car_state->DesiredVelocity = car_state->CruiseControlVelocity;
```
Cam0Cow, line 25: You should also set the desired motor current here. For now it's probably best to set it to 100% (1.0f) but that might be a parameter we want to tweak in the future.

Apps/Src/UpdateVelocity.c
```diff
@@ -0,0 +1,36 @@
+#include "SendTritium.h"
+#include "CarState.h"
+#include "UpdateVelocity.h"
+
+// As of 03/13/2021 the function just returns a 1 to 1 conversion
+static convertPedaltoMotorPercent(uint8_t pedalPercentage){
```
Cam0Cow, line 6: This function needs to return a float in the range of 0.0 to 1.0, so you probably want to divide the pedal percentage by 100.0.

Apps/Src/UpdateVelocity.c
```diff
@@ -0,0 +1,36 @@
...
+                car_state->DesiredVelocity = car_state->CruiseControlVelocity;
+            }else{
+                car_state->DesiredVelocity = MAX_VELOCITY;
+                car_state->DesiredMotorCurrent = convertPedaltoMotorPercent(pedalPercentage);
+            }
+        }else{
+            car_state->DesiredVelocity = MAX_VELOCITY;
+            car_state->DesiredMotorCurrent = convertPedaltoMotorPercent(pedalPercentage);
+        }
+    }
```
Cam0Cow, line 34: This loop could probably use some short delay of a few milliseconds at least, so it's not just eating up CPU time.

Apps/Src/UpdateVelocity.c
```diff
@@ -0,0 +1,36 @@
...
+
+    //Set desired velocity to 0
+    car_state->DesiredVelocity = 0;
+
+    while(1){
+        //Read global pedal percentage
+        uint8_t pedalPercentage = car_state->AccelPedalPercent;
+
+        if(car_state->CruiseControlEnable){
+            if(car_state->CruiseControlSet){
```
JimothyGreene, line 24: This can be made into one condition since the same thing happens in each `else` case. So, 
```if (car_state->CruiseControlEnable && car_state->CruiseControlSet)```

File: Apps/Src/Tasks.c [hunk 5]
```diff
@@ -61,4 +61,11 @@ OS_SEM SendCarCAN_Sem4;
 /**
  * Global Variables
  */
-// TODO: Put all global state variables here
\ No newline at end of file
+
+const switch_states_t switchStates = {OFF, OFF, OFF, OFF, OFF, OFF, OFF, OFF, OFF, OFF, OFF};
+const blinker_states_t blinkerStates = {OFF, OFF, OFF};
+
+car_state_t carState = {0.0f, 0.0f, 0.0f, 0.0f, 0, 0, 0, switchStates, blinkerStates, OFF, OFF};
```
JimothyGreene, line 68: I would prefer it if these were defined in the `main` function. So it'll be contained to a specific test or given case if we want to initialize things differently. Also, just to keep things in smaller scope. Keeping these global sort of circumvents the passing it to every task.

Apps/Inc/Tasks.h
```diff
@@ -133,7 +133,12 @@ extern OS_SEM SendCarCAN_Sem4;
 /**
  * Global Variables
  */
+
+
 // TODO: Put all global state variables here
+//const switch_states_t switchStates = {OFF, OFF, OFF, OFF, OFF, OFF, OFF, OFF, OFF, OFF, OFF};
+//const blinker_states_t blinkerStates = {OFF, OFF, OFF};
 
+//car_state_t carState = {0.0f, 0.0f, 0.0f, 0.0f, 0, 0, 0, switchStates, blinkerStates, OFF, OFF};
```
JimothyGreene, line 142: Can remove these comments now

JimothyGreene (changes requested):

JimothyGreene (approved):
Looks good and tests are working properly. Really cool to see this come together

Cam0Cow (approved):

PR #140, opened 2021-03-27
Rtos/read car can

Cam0Cow:


babusid:
CANBus_Read may have to be changed on a driver level to use a mutex lock to make sure that only one read happens at a time.

Apps/Src/ReadCarCAN.c
```diff
@@ -0,0 +1,52 @@
...
+            OSTaskSemPend(0, OS_OPT_PEND_BLOCKING, &ts, &err);
+            // TODO: error handling
+        } else if (err != OS_ERR_PEND_WOULD_BLOCK) {
+            // TODO: error handling
+        }
+
+        // Normal task countinues here
+
+        // Check if BPS told us to shut down the motor at least THRESHOLD times
+        if (CANbus_Read(buffer) == SUCCESS && (++faultCounter == THRESHOLD)) {
```
JimothyGreene, line 30: Also need to check what the actual message is. BPS will send either a 0 (charging disabled) or a 1 (charging enabled). If charging is disabled, we need to disconnect the contactors. If BPS is sending a 1, we don't need to change anything about the connections

JimothyGreene: Also, the threshold isn't really needed here. You should use the threshold to check if the message was received at all (i.e.) if `CANbus_Read` returns ERROR, that means BPS hasn't sent the message. If that happens THRESHOLD times, we should disconnect.

JimothyGreene (changes requested):

Apps/Src/ReadCarCAN.c
```diff
@@ -0,0 +1,50 @@
...
+                car->IsRegenBrakingAllowed = ON;
+            }
+
+            faultCounter = 0;
+        } else {
+            faultCounter++;
+        }
+
+        if (faultCounter >= THRESHOLD) {
+            car->IsRegenBrakingAllowed = ON;
```
JimothyGreene, line 43: I would say this should default to off if we're not receiving a message from BPS saying it's safe to have it on.

JimothyGreene (changes requested):

Tests/Test_ReadCarCAN.c
```diff
@@ -0,0 +1,89 @@
...
+    );
+
+    printf("Spawned ReadCarCAN\n");
+
+    OSTimeDlyHMSM(0, 0, 1, 0, OS_OPT_TIME_HMSM_STRICT, &err);
+    printf("Array is %d\n", car.ShouldArrayBeActivated);
+    CANbus_Send(MOTOR_DISABLE, payload);
+    OSTimeDlyHMSM(0, 0, 1, 0, OS_OPT_TIME_HMSM_NON_STRICT, &err);
+    printf("Array is %d\n", car.ShouldArrayBeActivated);
+    OSTaskDel(NULL, &err);
```
JimothyGreene, line 88: Can you change this to reflect the new operation?

JimothyGreene (changes requested):
Looks mostly good. Just one change on the test file to accommodate for the design change

JimothyGreene (approved):

PR #141, opened 2021-03-27
SendCarCan Task

babusid:
Implementation of Sending Messages over CAN using proper RTOS synchronization, targeted for the STFM microcontroller board


Apps/Src/Task_CANBusSend.c
```diff
@@ -0,0 +1,27 @@
...
+        (msg) = (CANMSG_t*)OSQPend(
+            &CANBus_MsgQ,
+            0,
+            OS_OPT_PEND_BLOCKING,
+            msgSize,
+            &timestamp,
+            &err
+        );
+        if (err == OS_ERR_NONE){ //if a message is recieved
+            CANbus_BlockSend(msg->id,msg->payload); //send message
```
ErickCortez98, line 23: I think this function has a different name in the CANbus driver file

Drivers/Src/CANbus.c
```diff
@@ -16,22 +37,84 @@ void CANbus_Init(void) {
  * @param   CAN bus line
  * @param   id : CAN id of the message
  * @param 	payload : the data that will be sent.
+ * @param blocking: Whether or not this should be a blocking call
  * @return  0 if data wasn't sent, otherwise it was sent.
  */
-int CANbus_Send(CANId_t id, CANPayload_t payload) {
+ErrorStatus CANbus_Send(CANId_t id, CANPayload_t payload, CAN_blocking_t blocking) {
```
ErickCortez98, line 43: I believe CAN_blockin_t is not a defined struct

Drivers/Src/CANbus.c
```diff
@@ -16,22 +37,84 @@ void CANbus_Init(void) {
...
  * @param 	payload : the data that will be sent.
+ * @param blocking: Whether or not this should be a blocking call
  * @return  0 if data wasn't sent, otherwise it was sent.
  */
-int CANbus_Send(CANId_t id, CANPayload_t payload) {
+ErrorStatus CANbus_Send(CANId_t id, CANPayload_t payload, CAN_blocking_t blocking) {
+    CPU_TS timestamp;
+    OS_ERR err;
+    //make sure that Can mailbox is available
+    if (blocking == CANBLOCKING){
```
ErickCortez98, line 47: CANBLOCKING is not defined

Drivers/Src/CANbus.c
babusid: It should be defined in the CANbus.h file

Drivers/Src/CANbus.c
babusid: Same as the previous comment, it should be defined on line 43 of CANbus.h

Drivers/Src/CANbus.c
```diff
@@ -1,61 +1,201 @@
...
+        return ERROR;
+    }
+
+    //GETMSG
+    uint8_t status = BSP_CAN_Read(CAN_1,ID,buffer);
     
-    if(ID == MOTOR_DISABLE){
-        for(int i=0;i<count;i++){
-            buffer[i]=data[i];
-        }
```
JimothyGreene, line 54: This function should still only return a `SUCCESS` if the message that was read was the correct ID. The open PR #140 has changed what specific ID this is, but it's still expecting this function to return based on the ID. I would wait for that PR to merge in then resolve any conflicts and keep the ID check.

JimothyGreene (changes requested):

Drivers/Src/CANbus.c
```diff
@@ -1,61 +1,201 @@
...
+static OS_SEM CANMail_Sem4;
+static OS_SEM CANBus_RecieveSem4; 
+static OS_MUTEX CANbus_TxMutex;
+static OS_MUTEX CANbus_RxMutex;
 
+//function that releases the mailbox semaphores
+//this will get passed to the BSP_CAN_Init function which will bind the release function to the tx ending handler provided by STFM
+void CANbus_Release(){ 
+    
+};
```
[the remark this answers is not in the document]
babusid: Yes it should, I just straight up forgot to write this. I'll finish this on the upcoming workday.

Apps/Src/Task_CANBusSend.c
```diff
@@ -0,0 +1,27 @@
+#include "os.h"
```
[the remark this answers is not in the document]
JimothyGreene: Still this

JimothyGreene (changes requested):
Logic seems good to me. Just change that file name and get a test written for this task

Apps/Src/ReadCarCAN.c
```diff
@@ -31,7 +31,7 @@ void Task_ReadCarCAN(void *p_arg) {
         // Normal task countinues here
 
         // Check if BPS sent us a message
-        if (CANbus_Read(buffer) == SUCCESS) {
+        if (CANbus_Read(buffer,CAN_NON_BLOCKING) == SUCCESS) {
             long msg = *((long *)(&buffer[0]));
             if (msg == 0) {
```
chaseblock, line 36: How are we sure that this message come from the BPS? How are we sure which message from the BPS it is?

chaseblock: I also have problems with this function in general, and think that the logic of it may need to be thought through again.

Apps/Src/SendCarCAN.c
```diff
@@ -0,0 +1,49 @@
...
+
+    // int size;
+
+    // size = 0;
+    // char str[128];
+    // uint8_t output;
+
+    // while (1){
+    //     output = CANbus_Send(ids[0], payload, CAN_BLOCKING);
+    // }
```
chaseblock, line 33: Can we take out the commented code if it isn't being used?

BSP/Inc/BSP_CAN.h
```diff
@@ -18,9 +18,11 @@ typedef enum {CAN_1=0, CAN_3, NUM_CAN} CAN_t;
  *          communicate with the motor controllers
  *          and other car systems
  * @param   bus the CAN line to initialize
+ * @param txHandler the Function that will be called every time TX completes
+ * @param rxHanlder the function that will be called on every RX event
  * @return  None
  */ 
-void BSP_CAN_Init(CAN_t bus);
+void BSP_CAN_Init(CAN_t bus,void (*txHandler)(void), void (*rxHander)(void));
```
chaseblock, line 25:
```suggestion
void BSP_CAN_Init(CAN_t bus, callback_t txHandler, callback_t rxHander);
```

BSP/Inc/BSP_OS.h
```diff
@@ -0,0 +1,10 @@
+/* Copyright (c) 2020 UT Longhorn Racing Solar */
+#ifndef BSP_OS_H
+#define BSP_OS_H
+
+typedef struct {
+    void (*pend)(void);
+    void (*post)(void);
```
chaseblock, line 7:
```suggestion
    callback_t pend;
    callback_t post;
```

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -0,0 +1,377 @@
...
+    uint8_t data[8];
+} msg_t;
+
+// Set up a fifo for receiving
+#define FIFO_TYPE msg_t
+#define FIFO_SIZE 10
+#define FIFO_NAME msg_queue
+#include "fifo.h"
+
+static msg_queue_t gRxQueue;
```
chaseblock, line 19: Why do we just have one global rx queue now? How are we to differentiate between the two busses?

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -0,0 +1,377 @@
...
+
+// Required for receiving CAN messages
+static CanTxMsg gTxMessage;
+static CanRxMsg gRxMessage;
+
+// User parameters for CAN events
+static void (*gRxEvent)(void);
+static void (*gTxEnd)(void);
```
chaseblock, line 27:
```suggestion
static callback_t gRxEvent;
static callback_t gTxEnd;
```

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -0,0 +1,377 @@
...
+void BSP_CAN3_Init();
+
+/**
+ * @brief   Initializes the CAN module that communicates with the rest of the electrical system.
+ * @param   rxEvent : the function to execute when recieving a message. NULL for no action.
+ * @param   txEnd   : the function to execute after transmitting a message. NULL for no action.
+ * @return  None
+ */
+
+void BSP_CAN_Init(CAN_t bus, void (*rxEvent)(void), void (*txEnd)(void)) {
```
chaseblock, line 40:
```suggestion
void BSP_CAN_Init(CAN_t bus, callback_t rxEvent, callback_t txEnd) {
```

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -0,0 +1,377 @@
...
+
+/**
+ * @brief   Gets the data that was received from the CAN bus.
+ * @note    Non-blocking statement
+ * @pre     The data parameter must be at least 8 bytes or hardfault may occur.
+ * @param   id : pointer to store id of the message that was received.
+ * @param   data : pointer to store data that was received. Must be 8bytes or bigger.
+ * @return  ERROR if nothing was received so ignore id and data that was received. SUCCESS indicates data was received and stored.
+ */
+ErrorStatus BSP_CAN_Read(CAN_t bus, uint32_t *id, uint8_t *data) {
```
chaseblock, line 279: We don't appear to be using the `bus` parameter at all here. Is it still necessary? If so, we should separate out the rx queue into two (one for each bus). If not, we should remove the parameter from the argument list.

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -0,0 +1,377 @@
...
+            // If the queue was not already full...
+            // Call the driver-provided function, if it is not null
+            if(gRxEvent != NULL) {
+                gRxEvent();
+            }
+        } else {
+            // If the queue is already full, then we can't really do anything else
+            break;
+        }
+    }
```
chaseblock, line 324: Need to insert the `OSIntExit();` call (see `CAN_1_RX0_IRQHandler`)

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -0,0 +1,377 @@
...
+    // Acknowledge 
+    CAN_ClearFlag(CAN1, CAN_FLAG_RQCP0 | CAN_FLAG_RQCP1 | CAN_FLAG_RQCP2);
+
+    // Call the function provided
+    gTxEnd();
+
+    #ifdef RTOS
+    OSIntExit();      // Signal to uC/OS
+    #endif
+}
```
chaseblock, line 377: Where's our CAN3 transmit handler?

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -0,0 +1,377 @@
...
+ * @param   rxEvent : the function to execute when recieving a message. NULL for no action.
+ * @param   txEnd   : the function to execute after transmitting a message. NULL for no action.
+ * @return  None
+ */
+
+void BSP_CAN_Init(CAN_t bus, void (*rxEvent)(void), void (*txEnd)(void)) {
+
+    // Configure event handles
+    gRxEvent  = rxEvent;
+    gTxEnd    = txEnd;
```
chaseblock, line 44: This initialization function will be called multiple times: one for each bus. Therefore, we should have multiple callbacks (one for each bus) or they will be overwritten.

BSP/STM32F413/Src/how_to_run_gdb.txt
```diff
@@ -0,0 +1,8 @@
+run `st-util`
+
+run `gdb-multiarch Objects/control-leader.elf`
+
+> target extended-remote localhost:4242
+> type 'start'
+> type ctrl + x + a to have a better view of the code
+> asdfghasdfg
```
chaseblock, line 8: What is this supposed to mean?

BSP/STM32F413/Src/how_to_run_gdb.txt
```diff
@@ -0,0 +1,8 @@
+run `st-util`
+
+run `gdb-multiarch Objects/control-leader.elf`
+
+> target extended-remote localhost:4242
+> type 'start'
+> type ctrl + x + a to have a better view of the code
```
chaseblock, line 7: Why do we now have the `>` prompt, if we're just using english words to describe actions instead of showing?

BSP/Simulator/Src/BSP_CAN.c
```diff
@@ -6,9 +6,13 @@
  */
 
 #include "BSP_CAN.h"
+#include "os.h"
 
 #define FILE_NAME DATA_PATH(CAN_CSV)
 
+static void (*txEvent)(void);
+static void (*rxEvent)(void);
```
chaseblock, line 14:
```suggestion
static callback_t txEvent;
static callback_t rxEvent;
```

Drivers/Src/CANbus.c
```diff
@@ -1,61 +1,197 @@
...
-    
-    }
-
-    return ERROR;
-    
-}
+#include "CANbus.h"
+#include "config.h"
+#include "os.h"
+static OS_SEM CANMail_Sem4; //sem4 to count how many sending hardware mailboxes we have left (start at 3)
```
chaseblock, line 4: Space out the includes and declarations.

Drivers/Src/CANbus.c
```diff
@@ -1,61 +1,197 @@
...
+    CANId_t ret;
+    uint8_t status = BSP_CAN_Read(CAN_1,&ret,buffer);
+    
+
+    OSMutexPost( //unlock RX line
+        &CANbus_RxMutex,
+        OS_OPT_POST_1,
+        &err
+    );
+    if(status&&(ret == CHARGE_ENABLE)){
```
chaseblock, line 189: I think that this is the wrong layer to be performing this check. Should probably either be doing this in the Apps layer or through the hardware filters in the BSP layer, or even better: in both (in case the system needs to pay attention to more messages in the future).

Drivers/Src/MotorController.c
```diff
@@ -11,13 +11,21 @@
 #define MASK_CC_ERR 1<<2 //checks velocity on 2 bit
 #define MASK_OVER_SPEED_ERR 1<<24 //check if motor overshot max RPM on 24 bit
 
+/**
+ * BSP_CAN_INIT requires a tx and rx event handler, and this driver doesn't actually require any tx/rx event handling. As such, this 
+ * is a dummy function just to satisfy the BSP layer's constraints.
+*/
+static void CanFunc(){
```
chaseblock, line 18: Do we not need to report the speed of the motor to anyone?

Apps/Src/ReadCarCAN.c
```diff
@@ -30,22 +31,24 @@ void Task_ReadCarCAN(void *p_arg) {
...
-
-            faultCounter = 0;
-        } else {
-            faultCounter++;
         }
 
         if (faultCounter >= THRESHOLD) {
             car->IsRegenBrakingAllowed = OFF;
+            Contactors_Set(ARRAY, OFF);
         }
```
ClarkPoon, line 52: I'm interested to see how often `faultCounter` is incremented when we run the car. One potential issue I see with this is that if we very occasionally miss a CAN message (eg. once every 10 minutes) the car might seem to randomly shut off after a certain period of time. It might make sense to reset `faultCounter` to zero periodically

Apps/Src/ReadCarCAN.c
```diff
@@ -30,22 +31,24 @@ void Task_ReadCarCAN(void *p_arg) {
...
+        if (CANbus_Read(&canId, buffer, CAN_NON_BLOCKING) == SUCCESS) {
+            // If charge_enable, set regen flag
+            if (canId == CHARGE_ENABLE) {
+                car->IsRegenBrakingAllowed = (buffer[0] == 0) ? OFF : ON;
+                Contactors_Set(ARRAY, car->IsRegenBrakingAllowed);
+                faultCounter = 0;
             } else {
-                car->IsRegenBrakingAllowed = ON;
+                // If we didn't get a message, something might have gone wrong
+                faultCounter++;
```
ClarkPoon, line 45: Doesn't this line only get run if we receive an CAN message with an id other that `CHARGE_ENABLE`? This code will not run if you aren't receiving any CAN messages

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -0,0 +1,402 @@
...
+    NVIC_Init(&NVIC_InitStruct);	
+
+    /*if(NULL != txEnd) {
+        // Enable Tx Interrupts
+        NVIC_InitStruct.NVIC_IRQChannelPreemptionPriority = 0x0; // TODO: assess both of these priority settings
+        NVIC_InitStruct.NVIC_IRQChannelSubPriority = 0x0;
+        NVIC_InitStruct.NVIC_IRQChannelCmd = ENABLE;
+        NVIC_InitStruct.NVIC_IRQChannel = CAN1_TX_IRQn;
+        NVIC_Init(&NVIC_InitStruct);
+    }*/
```
ClarkPoon, line 151: Are you planning on using tx interrupts? If not, are you planning on removing this?

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -0,0 +1,402 @@
...
+    CAN_FilterInitStruct.CAN_FilterScale = CAN_FilterScale_32bit;
+    CAN_FilterInitStruct.CAN_FilterIdHigh = 0x0000;
+    CAN_FilterInitStruct.CAN_FilterIdLow = 0x0000;
+    CAN_FilterInitStruct.CAN_FilterMaskIdHigh = 0x0000;
+    CAN_FilterInitStruct.CAN_FilterMaskIdLow = 0x0000;
+    CAN_FilterInitStruct.CAN_FilterFIFOAssignment = 0;
+    CAN_FilterInitStruct.CAN_FilterActivation = ENABLE;
+    CAN_FilterInit(CAN1, &CAN_FilterInitStruct);
+
+    //CAN_SlaveStartBank(CAN1, 0);
```
ClarkPoon, line 119: Are you planning on removing this?

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -0,0 +1,402 @@
...
+    GPIO_InitStruct.GPIO_OType = GPIO_OType_PP;
+    GPIO_InitStruct.GPIO_PuPd = GPIO_PuPd_UP;
+    GPIO_Init(GPIOA, &GPIO_InitStruct);
+
+    /* CAN configuration ********************************************************/  
+    /* Enable CAN clock */
+    RCC_APB1PeriphClockCmd(RCC_APB1Periph_CAN1, ENABLE);
+
+    /* CAN register init */
+    //CAN_DeInit(CAN1);
```
ClarkPoon, line 86: Are you planning on removing this?

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -0,0 +1,402 @@
...
+    GPIO_InitStruct.GPIO_OType = GPIO_OType_PP;
+    GPIO_InitStruct.GPIO_PuPd = GPIO_PuPd_UP;
+    GPIO_Init(GPIOA, &GPIO_InitStruct);
+
+    /* CAN configuration ********************************************************/  
+    /* Enable CAN clock */
+    RCC_APB1PeriphClockCmd(RCC_APB1Periph_CAN3, ENABLE);
+
+    /* CAN register init */
+    //CAN_DeInit(CAN1);
```
ClarkPoon, line 189: Are you planning on removing this?

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -0,0 +1,402 @@
...
+ */
+ErrorStatus BSP_CAN_Write(CAN_t bus, uint32_t id, uint8_t data[8], uint8_t length) {
+    
+    gTxMessage[bus].StdId = id;
+    gTxMessage[bus].DLC = length;
+	for(int i = 0; i < length; i++){
+        gTxMessage[bus].Data[i] = data[i];
+    }
+
+    uint8_t retVal = (CAN_Transmit(bus == CAN_1 ? CAN1 : CAN3, &gTxMessage[bus]) != 0);
```
ClarkPoon, line 264: FYI, `CAN_Transmit()` returns the mailbox it used, so you are returning errors, when you shouldn't (if it uses mailbox 1 or 2). I would refer to [BPS's implementation](https://github.com/lhr-solar/BPS/blob/6eb5140300e02cd123a3e009d797d15ab42d0769/BSP/STM32F413/Src/BSP_CAN.c#L156) to fix this

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -0,0 +1,402 @@
...
+    for(int i = 0; i < 8; i++){
+        data[i] = msg.data[i];
+    }
+    *id = msg.id;
+
+    return SUCCESS;
+}
+
+void CAN3_RX0_IRQHandler() {
+    #ifdef RTOS
```
ClarkPoon, line 299: Are y'all defining `RTOS` anywhere? Not sure if y'all copied that part of the makefile from BPS or not

BSP/STM32F413/Src/how_to_run_gdb.txt
ClarkPoon: @babusid

BSP/STM32F413/Src/how_to_run_gdb.txt
ClarkPoon: @babusid

Drivers/Src/MotorController.c
```diff
@@ -11,13 +11,21 @@
 #define MASK_CC_ERR 1<<2 //checks velocity on 2 bit
 #define MASK_OVER_SPEED_ERR 1<<24 //check if motor overshot max RPM on 24 bit
 
+/**
+ * BSP_CAN_INIT requires a tx and rx event handler, and this driver doesn't actually require any tx/rx event handling. As such, this 
+ * is a dummy function just to satisfy the BSP layer's constraints.
```
ClarkPoon, line 16: This works, but a more elegant solution might be to modify the BSP to not do anything for the rx or tx event handlers if you pass in `NULL`

Apps/Src/ReadCarCAN.c
```diff
@@ -15,21 +16,21 @@ void Task_ReadCarCAN(void *p_arg) {
...
         // Check if BPS sent us a message
-        if (CANbus_Read(buffer) == SUCCESS) {
-            long msg = *((long *)(&buffer[0]));
-            if (msg == 0) {
-                car->IsRegenBrakingAllowed = OFF;
+        if (CANbus_Read(&canId, buffer, CAN_NON_BLOCKING) == SUCCESS) {
+            // If charge_enable, set regen flag
+            if (canId == CHARGE_ENABLE) {
+                car->IsRegenBrakingAllowed = (buffer[0] == 0) ? OFF : ON;
+                Contactors_Set(ARRAY, car->IsRegenBrakingAllowed);
```
babusid, line 23: @Cam0Cow Shouldn't we be sending something to the motor to turn off the motor side Regen Braking here? Or is this handled somewhere else at the moment

Cam0Cow: The logic for handling this is in the UpdateVelocity task. However, looking back at it, I'm not sure if actually disables regen braking when in normal (non cruise-control) mode. I don't see any specific command for disabling regenerative braking in the wavesculptor documentation, and the code doesn't try to do anything clever to make sure regen can't happen. We should definitely figure out how we want to handle it this coming workday.

babusid:
> The good news is that the two tasks and the driver that are associated with this branch are fully validated. The bad news is that I'm going to go through and remove the CarState code from our CAN module **(will not break SendCarCAN and ReadCarCAN)** but will break the mainline version of UpdateVelocity until it gets its rewrite. Since we are rewriting that thread regardless, I see no issues in doing this. @ClarkPoon feel free to correct me if you feel this is a bad approach.

`hotfix_RTOS_SencCarCAN` branch may need to be merged into this one for this comment to be fully correct

babusid:
Need to merge master into this and add OS error assertions everywhere there's an os function call

babusid:
Things that need to get done to finish this package:

- Merge in Hotfix_SendCarCan branch
- Merge in master
- add OS error assertions everywhere
- address clark's comments regarding the fault counter on readCarCAN

babusid:
Remaining TODO list:

- Add OS error assertions into the driver layer
- Re-write ReadCarCAN using One-shot timers and handle array precharge and fault correctly (I'm doing this first)
- Ensure car state code is fully removed from CAN package (I think this is done, but one more sweep is needed.)

BSP/STM32F413/Src/BSP_CAN.c
babusid: @Cam0Cow @ErickCortez98  Is there a reason this is commented out? I'm very sure that I pass in a TXEnd handler on the driver layer, is the transmit ISR initialized elsewhere?

Drivers/Inc/Contactors.h
```diff
@@ -33,9 +34,8 @@ State Contactors_Get(contactor_t contactor);
 
 /**
  * @brief   Sets the state of a specified contactor
+ * @param   contactor the contactor (MOTOR/ARRAY)
  * @param   contactor the contactor
```
ClarkPoon, line 38:
```suggestion
 * @param   contactor the contactor (MOTOR/ARRAY)
```

Drivers/Src/CANbus.c
```diff
@@ -1,61 +1,209 @@
...
+static OS_MUTEX CANbus_RxMutex;   // mutex to lock Rx line
+
+/**
+ * @brief this function will be passed down to the BSP layer to trigger on RX events. Increments the recieve semaphore to signal message in hardware mailbox. Do not access directly.
+ */
+void CANbus_RxHandler()
+{
+    OS_ERR err;
+    OSSemPost(&CANBus_RecieveSem4, OS_OPT_POST_1, &err); // increment our queue counter
+    assertOSError(0,err);
```
ClarkPoon, line 18: You probably want a location for this instead of 0

Drivers/Src/Contactors.c
```diff
@@ -10,6 +10,7 @@
 #define MOTOR_CONTACTOR_PIN GPIO_Pin_12
 
 
+
```
ClarkPoon, line 13:
```suggestion
```

Tests/Test_ReadCarCAN.c
```diff
@@ -1,16 +1,12 @@
-#include "common.h"
-#include "config.h"
+
+
```
ClarkPoon, line 2:
```suggestion
```

Tests/Test_ReadCarCAN.c
```diff
@@ -19,16 +15,28 @@ void main(void) {
...
+    }
+
+    OSTaskCreate( //create readCarCAN task
+        (OS_TCB*)&ReadCarCAN_TCB,
+        (CPU_CHAR*)"Read Car CAN",
+        (OS_TASK_PTR)Task_ReadCarCAN,
         (void*)NULL,
-        (OS_PRIO)3,
-        (CPU_STK*)Task1_STK,
+        (OS_PRIO)12,
```
ClarkPoon, line 35: `#define` and `const OS_PRIO` are your friends

Apps/Src/ReadCarCAN.c
```diff
@@ -1,41 +1,180 @@
...
+static inline void chargingEnable(void) {
+    OS_ERR err;
+    CPU_TS ts;
+
+    // mark regen as disabled
+    RegenAllowed = ON;
+
+    // enable contactors
+    Contactors_Enable(ARRAY_CONTACTOR);
+    Contactors_Enable(ARRAY_PRECHARGE);
```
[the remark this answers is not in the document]
ClarkPoon: Yeah, I'll fix this

Apps/Src/ReadCarCAN.c
```diff
@@ -1,41 +1,180 @@
...
-        if(err != OS_ERR_NONE){
-            car->ErrorCode.ReadCANErr = ON;
-        }
+/**
+ *  * @brief This function is the array precharge thread that gets instantiated when array restart needs to happen.
+*/
+static void ArrayRestart(void *p_arg){
+    OS_ERR err;
+
+    if (Contactors_Set(ARRAY_PRECHARGE, ON) == SUCCESS) {    // only run through precharge sequence if we successfully start precharge
```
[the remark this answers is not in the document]
ClarkPoon: If this condition fails, the task will do nothing and delete itself (which is probably good enough for handling the error)

Apps/Src/ReadCarCAN.c
```diff
@@ -1,41 +1,180 @@
...
+
+        OSTimeDlyHMSM(0,0,0,PRECHARGE_ARRAY_DELAY*100,OS_OPT_TIME_HMSM_STRICT,&err); //delay
+        assertOSError(OS_READ_CAN_LOC,err);
+
+        Contactors_Set(ARRAY_CONTACTOR, ON);
+        Contactors_Set(ARRAY_PRECHARGE, OFF);
```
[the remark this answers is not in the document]
ClarkPoon: Ok. I can add this

ClarkPoon (dismissed):
I wrote a lot of this at the end and tested it. Y'all can read through the code and review if you want

Apps/Inc/Tasks.h
```diff
@@ -7,6 +7,7 @@
 #include "os.h"
 #include "common.h"
 #include "CarState.h"
```
babusid, line 9: Should probably get rid of this header being included at some point

ClarkPoon: Fixed in #201

PR #142, opened 2021-04-17
Bsp/gpio

ErickCortez98:
BSP_GPIO for STM board implemented (needs review for sure)


BSP/STM32F413/Src/BSP_GPIO.c
```diff
@@ -0,0 +1,121 @@
...
+	
+	if(err != OS_ERR_NONE){
+        printf("OS error code %d\n", err);
+    }
+}
+
+// Use this inline function to wait until GPIO communication is complete
+static inline void GPIO_Wait(){
+	gpio_pend();
+}
```
chaseblock, line 73: Any reason in particular to have these two functions instead of just one?

Also, what resources are we trying to protect with these semaphores?

BSP/STM32F413/Src/BSP_GPIO.c
```diff
@@ -0,0 +1,121 @@
...
+		break;
+		case PORTC:
+		gpio_port = GPIOC;
+		break;
+		case PORTD:
+		gpio_port = GPIOD;
+		break;
+		default:
+		return;
+	}
```
chaseblock, line 93: These one-to-one mappings can usually be turned into something better. If pointer math seems too opaque, then something like the below might be better.
```c
GPIO_TypeDef * const gpio_mapping[4] = {GPIOA, GPIOB, GPIOC, GPIOD};
...
gpio_port = gpio_mapping[port];
```

In this case though, why not just have the user pass PORTA, PORTB, PORTC, etc. directly?

BSP/STM32F413/Src/BSP_GPIO.c
```diff
@@ -0,0 +1,121 @@
...
+ * @brief   Writes data to a specified port 
+ * @param   port to write to
+ * @param   data to write 
+ * @return  None
+ */ 
+void BSP_GPIO_Write(port_t port, uint16_t data){
+	GPIO_TypeDef *gpio_port = GPIO_GetPort(port);
+	
+	GPIO_Wait(gpio_port);
+	gpio_port->ODR = data & 0xFFFF;
```
chaseblock, line 120: If we use pin-specific addressing, we may not need to worry about semaphores, unless the task that is manipulating the pin shares access to the pin with some other process.

Also, whenever you take a semaphore, be sure to release it once you're done, otherwise it will stall everything.

chaseblock: Same comment on the BSP_GPIO_Read function.

BSP/STM32F413/Src/BSP_GPIO.c
```diff
@@ -0,0 +1,121 @@
...
+#include "Tasks.h"
+
+OS_SEM GPIO_Update_Sem4;
+
+/**
+ * @brief   Initializes a GPIO port
+ * @param   port to initialize
+ * @return  None
+ */ 
+void BSP_GPIO_Init(port_t port){
```
chaseblock, line 14: See our recent discussion on this BSP for some idea of what could be potentially improved with the init function.

chaseblock (commented):
Just some thoughts. Take these and our recent discussion into account while revising.

BSP/STM32F413/Src/BSP_GPIO.c
ErickCortez98: I think it waiting was implemented based on some BPS code without given much thought by the previous person working on this, the same applies to me. But yeah, I think we don't even have a GPIO semaphore being triggered at another place in the software so I'll get rid of this.

BSP/STM32F413/Src/BSP_GPIO.c
ErickCortez98: We do the mapping in order to maintain compatibility with the simulator. Since the port_t enum was used originally when the file was created for the simulator all calls to the GPIO functions use it. I really like the mapping you suggested, I think we can implement it that way.

BSP/STM32F413/Src/BSP_GPIO.c
ErickCortez98: Yeah, in the most recent code I still didn't change the init function, I'm just waiting to get some feedback from Rishi in regards to it because of possible conflicts with the simulator, but I copied the improvements you suggested as comments for now

BSP/Inc/BSP_GPIO.h
```diff
@@ -16,10 +16,10 @@ typedef enum {PORTA = 0, PORTB, PORTC, PORTD, NUM_PORTS} port_t;
 
 /**
  * @brief   Initializes a GPIO port
- * @param   port to initialize
+ * @param   port -port to initialize, mask - pins, write - input or output
```
JimothyGreene, line 19: Just a documentation thing. You can use multiple `@param` lines for multiple parameters. Just like how the other methods in here have done. Just in case we ever do some automatic documentation generator, it'll want it to be in that format.

BSP/Inc/BSP_GPIO.h
```diff
@@ -36,4 +36,29 @@ uint16_t BSP_GPIO_Read(port_t port);
  */ 
 void BSP_GPIO_Write(port_t port, uint16_t data);
 
+/**
+ * @brief   Reads data to a specified pin
+ * @param   port The port to read from
+ * @param   pin The pin to read from 
+ * @return  State of the pin
+ */ 
+uint8_t BSP_GPIO_Read_Pin(port_t port, uint8_t pin);
```
JimothyGreene, line 45: Maybe put a note here that this cannot be used for output pins

BSP/STM32F413/Src/BSP_GPIO.c
```diff
@@ -0,0 +1,102 @@
+/* Copyright (c) 2021 UT Longhorn Racing Solar */
+
+#include "BSP_GPIO.h"
+#include "stm32f4xx.h"
+#include "Tasks.h"
+
+OS_SEM GPIO_Update_Sem4;
```
JimothyGreene, line 7: Where is this used?

BSP/STM32F413/Src/BSP_GPIO.c
```diff
@@ -0,0 +1,102 @@
...
+	RCC_AHB1PeriphClockCmd(1 << port, ENABLE);
+	// Configure the pins to be generic GPIO
+    GPIO_InitStruct.GPIO_Pin   = mask;
+    GPIO_InitStruct.GPIO_Mode  = write ? GPIO_Mode_OUT : GPIO_Mode_IN;
+    GPIO_InitStruct.GPIO_OType = GPIO_OType_PP;
+    GPIO_InitStruct.GPIO_Speed = GPIO_Low_Speed;
+    GPIO_InitStruct.GPIO_PuPd  = GPIO_PuPd_NOPULL;  // TODO: verify
+
+    // Compute the offset for the port handle from the port passed in
+    GPIO_TypeDef *portHandle = GET_HANDLE_FROM_ENUM(port);
```
JimothyGreene, line 27: Where is the definition for this `GET_HANDLE_FROM_ENUM` macro? It seems like you could use the `GPIO_GetPort` function here instead.

Drivers/Src/Contactors.c
```diff
@@ -13,7 +13,7 @@
  * @return  None
  */ 
 void Contactors_Init(contactor_t contactor) {
-    BSP_GPIO_Init(CONTACTORS_PORT);
+    BSP_GPIO_Init(CONTACTORS_PORT, 0xFFFF, 1); //All pins in Port C are output
```
JimothyGreene, line 16: Only initialize the pins that are being used for the contactors here. Not all of pins 0-15

Drivers/Src/Lights.c
```diff
@@ -7,7 +7,7 @@
 * @return  void
 */ 
 void Lights_Init(void) {
-    BSP_GPIO_Init(LIGHTS_PORT);
+    BSP_GPIO_Init(LIGHTS_PORT, 0xFFFF, 1); // All pins from Port C are out
```
JimothyGreene, line 10: Same as Contactors comment

JimothyGreene, line 10: Same as Contactors comment

JimothyGreene, line 10: Same as Contactors comment

JimothyGreene (changes requested):
Few questions I had regarding where some of these macros/functions are coming from. And then just some small changes to the driver stuff

BSP/STM32F413/Src/BSP_GPIO.c
ErickCortez98: Not used, it was used in a previous version of the file, I'll delete it

BSP/STM32F413/Src/BSP_GPIO.c
ErickCortez98: Yep, I'll change it to use GPIO_GetPort, GET_HANDLE_FROM_ENUM was a macro we could use instead of the function but we sticked to using the function

Cam0Cow (approved):

PR #148, opened 2021-09-25
added error code struct and error checking to most of the app files

### LINKED ISSUE #139: Error Handling in Application Software
Since we can't just shut down the Controls subsystem anytime something goes wrong, we need a way to classify errors as recoverable and unrecoverable, as well as set up the necessary mechanisms for recovering and for completely shutting down.

IshDeshpa:
closes #139


Cam0Cow (dismissed):
Looks good to me, but won't merge it for now. Might be worth to amend it later to include the ability to show the error codes on the display (once that driver is finalized).

Cam0Cow (approved):
Still good

PR #153, opened 2021-10-02
added a struct and error handling for certain bit values from the motor controller

tasniabari:


Drivers/Src/MotorController.c
```diff
@@ -46,7 +45,8 @@ void MotorController_Drive(float newVelocity, float motorCurrent){
  * @param   message the buffer in which the info for the CAN message will be stored
  * @return  SUCCESS if a message is read
  */ 
-ErrorStatus MotorController_Read(CANbuff *message){
+ErrorStatus MotorController_Read(CANbuff *message, void *p_arg){
```
Cam0Cow, line 48: Is there any particular reason this is taking a void pointer instead of a pointer to a car state? I can't see a reason for this to be generic.

Drivers/Src/MotorController.c
```diff
@@ -70,6 +70,36 @@ ErrorStatus MotorController_Read(CANbuff *message){
         }
         message->firstNum = firstSum;
         message->secondNum = secondSum;
+
+        //read first number (bits 0-31) to get motor error issues
+        uint32_t maskMotorTemp=64; //check if motor temperature is an issue
```
Cam0Cow, line 75: Probably should define this as a constant somewhere (perhaps at the top of the file). It would also be a bit nicer to define it as 1<<6 instead of 64, to make it clear we're checking the sixth bit.

Drivers/Src/MotorController.c
```diff
@@ -70,6 +70,36 @@ ErrorStatus MotorController_Read(CANbuff *message){
...
+        uint32_t motorTempError=maskMotorTemp & firstSum;
+        if(motorTempError==64)
+        {
+             car_state->MotorErrorCode.motorTempErr = ON;
+        }
+
+        uint32_t maskSSErr=524288; //check for slip or hall sequence position error on 19 bit
```
Cam0Cow, line 82: Likewise, make this 1<<19.

Drivers/Src/MotorController.c
```diff
@@ -70,6 +70,36 @@ ErrorStatus MotorController_Read(CANbuff *message){
...
+        uint32_t maskCCErr=4; //checks velocity on 2 bit
+        uint32_t CCErr=maskCCErr & firstSum;
+        if(CCErr==4)
+        {
+             car_state->MotorErrorCode.CCVelocityErr = ON;
+        }
+
+        uint32_t maskOverSpeed=16777216; //check if motor overshot max RPM on 24 bit
+        uint32_t motorOverSpeedErr=maskOverSpeed & firstSum;
+        if(motorOverSpeedErr==16777216)
```
Cam0Cow, line 98: Likewise, 1<<24.

Drivers/Src/MotorController.c
```diff
@@ -2,6 +2,10 @@
 #include "CarState.h"
 #define MOTOR_DRIVE 0x221
 #define MAX_CAN_LEN 8
+#define maskMotorTempErr=1<<6; //check if motor temperature is an issue on bit 6
```
Cam0Cow, line 5: Since these are constants, they should be defined in all caps e.g. MASK_MOTOR_TEMP_ERR.

Drivers/Src/MotorController.c
```diff
@@ -45,13 +49,14 @@ void MotorController_Drive(float newVelocity, float motorCurrent){
  * @param   message the buffer in which the info for the CAN message will be stored
  * @return  SUCCESS if a message is read
  */ 
-ErrorStatus MotorController_Read(CANbuff *message, void *p_arg){
-    car_state_t *car_state = (car_state_t *) p_arg;
+ErrorStatus MotorController_Read(CANbuff *message, car_state_t *car){
+    car_state_t *car_state = (car_state_t *) car;
```
Cam0Cow, line 53: Since you're passing in a car_state_t pointer now, there's no reason to cast it, you can just use it directly.

Tests/Test_MotorConErrMessages.c
```diff
@@ -28,11 +28,33 @@ int main(){
...
+    //printf("The ID on the bus was: %x\n\rMessage: %x, %x\n\rSuccess: %d",tester.id, tester.firstNum, tester.secondNum, check);
 
-    printf("Error flags for motor temp, CC velocity, slipspeed, overspeed: %d\n\r, %d\n\r, %d\n\r, %d\n\r",car->MotorErrorCode.motorTempErr,
-    car->MotorErrorCode.slipSpeedErr, car->MotorErrorCode.CCVelocityErr, car->MotorErrorCode.overSpeedErr); 
-    //I assume that the error flags are decimal numbers, but I'm not sure??
+    printf("Error flags for motor temp, slipspeed, CC velocity, overspeed are on: %d\n\r, %d\n\r, %d\n\r, %d\n\r",
+    car.MotorErrorCode.motorTempErr,car.MotorErrorCode.slipSpeedErr, car.MotorErrorCode.CCVelocityErr, 
+    car.MotorErrorCode.overSpeedErr); 
+
+    test_data=0x00080044
```
Cam0Cow, line 37: Don't forget semicolons.

Tests/Test_MotorConErrMessages.c
```diff
@@ -28,11 +28,33 @@ int main(){
...
+    printf("Error flags for motor temp, slipspeed, CC velocity are on. Overspeed off: %d\n\r, %d\n\r, %d\n\r, %d\n\r",
+    car.MotorErrorCode.motorTempErr,car.MotorErrorCode.slipSpeedErr, car.MotorErrorCode.CCVelocityErr, 
+    car.MotorErrorCode.overSpeedErr); 
+
+    test_data=0x00080040
```
Cam0Cow, line 42: All of these test_data assignments need a semicolon.

Tests/Test_MotorConErrMessages.c
```diff
@@ -34,22 +34,22 @@ int main(){
     car.MotorErrorCode.motorTempErr,car.MotorErrorCode.slipSpeedErr, car.MotorErrorCode.CCVelocityErr, 
     car.MotorErrorCode.overSpeedErr); 
 
-    test_data=0x00080044
+    test_data=0x00080044;
     printf("Error flags for motor temp, slipspeed, CC velocity are on. Overspeed off: %d\n\r, %d\n\r, %d\n\r, %d\n\r",
```
Cam0Cow, line 38: At what point are these flags set?

Drivers/Src/MotorController.c
```diff
@@ -78,22 +78,22 @@ ErrorStatus MotorController_Read(CANbuff *message, car_state_t *car){
 
         //read first number (bits 0-31) from CAN message to check motor error issue flags
         
-        if(maskMotorTempErr & firstSum)
+        if(MASK_MOTOR_TEMP_ERR & firstSum)
         {
              car.MotorErrorCode.motorTempErr = ON;
```
Cam0Cow, line 83: car is a pointer, so this should be car->MotorErrorCode

Cam0Cow (changes requested):

Issue #151, opened 2021-09-25
Implement regenerative braking


Cam0Cow:
Regenerative braking is not currently implemented. Make sure to respect the IsRegenAllowed flag in the global car state, as regen braking when the flag is not set could be very dangerous and might lead to system shutdown.

PR #154, opened 2021-10-09
add regenerative braking

IshDeshpa:
#151

### DIFF
--- a/Apps/Inc/CarState.h
+++ b/Apps/Inc/CarState.h
@@ -39,11 +39,10 @@ typedef struct {
 } blinker_states_t;
 
 /**
- * Car State
+ * Error States
  * 
- * Stores all relevant data about the current
- * state of the car
- * This is used by all the tasks to communicate
+ * Stores error data to indicate which part of the code
+ * an error is coming from.
  */
 
 typedef struct{
@@ -58,6 +57,29 @@ typedef struct{
     State MotorConnectionErr;
 } error_code_t;
 
+/**
+ * Regen Brake Mode Enum
+ * 
+ * Different modes of the regenerative braking
+ * button
+ */
+typedef enum{OFF, RATE1, RATE2, RATE3} RegenMode;
+
+/**
+ * Regen/Cruise Enable Enum
+ */
+
+typedef enum{OFF, REGEN, CRUISE} CruiseRegenEnable;
+
+
+/**
+ * Car State
+ * 
+ * Stores all relevant data about the current
+ * state of the car
+ * This is used by all the tasks to communicate
+ */
+
 typedef struct {
     float DesiredVelocity;
     float CruiseControlVelocity;
@@ -71,15 +93,19 @@ typedef struct {
     switch_states_t SwitchStates;
     blinker_states_t BlinkerStates;
 
-    State CruiseControlEnable;
+    CruiseRegenEnable CREnable;    
     State CruiseControlSet;
+    
+    double RegenBrakeRate;
+    State IsRegenBrakingAllowed;
+    RegenMode RegenButtonMode;
 
     State ShouldArrayBeActivated;
     State ShouldMotorBeActivated;
 
-    State IsRegenBrakingAllowed;
-
     error_code_t ErrorCode;
+
+    
 
 } car_state_t;
 
--- a/Apps/Inc/CarState.h
+++ b/Apps/Inc/CarState.h
@@ -66,10 +66,10 @@ typedef struct{
 typedef enum{REGEN_OFF, RATE3, RATE2, RATE1} RegenMode;
 
 /**
- * Regen/Cruise Enable Enum
+ * Regen/Cruise Set Enum
  */
 
-typedef enum{CR_OFF, REGEN, CRUISE} CruiseRegenEnable;
+typedef enum{ACCEL, REGEN, CRUISE} CruiseRegenSet;
 
 
 /**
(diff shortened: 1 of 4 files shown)


Apps/Inc/CarState.h
```diff
@@ -58,6 +57,29 @@ typedef struct{
     State MotorConnectionErr;
 } error_code_t;
 
+/**
+ * Regen Brake Mode Enum
+ * 
+ * Different modes of the regenerative braking
+ * button
+ */
+typedef enum{OFF, RATE1, RATE2, RATE3} RegenMode;
```
Cam0Cow, line 66: Since we already have an enum with an OFF defined, you'll have to rename this option to something else (maybe REGEN_OFF?).

Apps/Inc/CarState.h
```diff
@@ -58,6 +57,29 @@ typedef struct{
...
+
+/**
+ * Regen/Cruise Enable Enum
+ */
+
+typedef enum{OFF, REGEN, CRUISE} CruiseRegenEnable;
```
Cam0Cow, line 72: Likewise here, maybe call this CRUISE_REGEN_OFF or something similar.

Apps/Inc/CarState.h
```diff
@@ -71,16 +93,20 @@ typedef struct {
     switch_states_t SwitchStates;
     blinker_states_t BlinkerStates;
 
-    State CruiseControlEnable;
+    CruiseRegenEnable CREnable;    
     State CruiseControlSet;
+    
+    double RegenBrakeRate;
```
Cam0Cow, line 99: We're trying to avoid floating point math so this should be done with fixed point. If the setpoints we have are going to be in 25% increments, maybe keep regenBrakeRate as an integer between 0 and 4, and then calculate the desired velocity as (currentVelocity * regenBrakeRate) >> 2. This math would be slightly different for different setpoint arrangements, but that's the general idea.

Apps/Src/UpdateVelocity.c
```diff
@@ -22,13 +22,44 @@ void Task_UpdateVelocity(void* p_arg){
...
+
+            car_state->RegenButtonMode = OFF;
+        }
+        else if(car_state->CREnable == REGEN){
+            switch(car_state->RegenButtonMode){
+                case OFF:
+                    car_state->CREnable = OFF;
+                    break;
+                case RATE1:
+                    car_state->CREnable = REGEN;
```
Cam0Cow, line 37: We only get to this point if CREnable is set to REGEN, so there's not reason to set it again.

Apps/Src/UpdateVelocity.c
```diff
@@ -22,13 +22,44 @@ void Task_UpdateVelocity(void* p_arg){
...
+                    car_state->RegenBrakeRate = 0.50;
+                    break;
+                case RATE3:
+                    car_state->CREnable = REGEN;
+                    car_state->RegenBrakeRate = 0.25;
+                    break;
+            }
+
+            if(car_state->IsRegenBrakingAllowed){
+                car_state->DesiredVelocity = car_state->CurrentVelocity*car_state->RegenBrakeRate;
```
Cam0Cow, line 51: See comment in CarState.h

Apps/Src/UpdateVelocity.c
```diff
@@ -22,13 +22,52 @@ void Task_UpdateVelocity(void* p_arg){
...
         uint8_t pedalPercentage = car_state->AccelPedalPercent;
 
-        if(car_state->CruiseControlEnable && car_state -> CruiseControlSet){
-            car_state->DesiredVelocity = car_state->CruiseControlVelocity;
-            car_state->DesiredMotorCurrent = 1.0f;
-        }else{
+        if(car_state->CREnable == CRUISE){
+            if(car_state->IsRegenBrakingAllowed){
+                if(car_state -> CruiseControlSet){
+                    car_state->DesiredVelocity = car_state->CurrentVelocity;
```
Cam0Cow, line 28: CruiseControlSet should be on whenever the car is in cruise control with a velocity already set. This line will keep setting the desired velocity to the current velocity when it should be set to CruiseControlVelocity. The cruise control velocity needs to be set seperately whenever the cruise set button is pressed.

Cam0Cow (changes requested):

Apps/Src/UpdateVelocity.c
```diff
@@ -22,13 +22,46 @@ void Task_UpdateVelocity(void* p_arg){
...
-        if(car_state->CruiseControlEnable && car_state -> CruiseControlSet){
-            car_state->DesiredVelocity = car_state->CruiseControlVelocity;
-            car_state->DesiredMotorCurrent = 1.0f;
-        }else{
-            car_state->DesiredVelocity = MAX_VELOCITY;
-            car_state->DesiredMotorCurrent = convertPedaltoMotorPercent(pedalPercentage);
-        }        
+        if(car_state -> IsRegenBrakingAllowed && car_state -> BrakePedalPercent < 5 && car_state -> AccelPedalPercent < 5){
+            if(car_state -> CruiseControlEnable){
+                if(car_state -> CRSet == CRUISE){
```
Cam0Cow, line 27: These two nested ifs should probably be combined into one.

Apps/Src/UpdateVelocity.c
```diff
@@ -22,13 +22,46 @@ void Task_UpdateVelocity(void* p_arg){
...
+                    car_state->DesiredVelocity = car_state->CruiseControlVelocity;
+                    car_state->DesiredMotorCurrent = 1.0f;
+                }
+            }
+
+            if(car_state -> CRSet == REGEN){
```
Cam0Cow, line 33: Could be an else if for better clarity

Cam0Cow (changes requested):

Cam0Cow (approved):

PR #156, opened 2021-10-16
Pin changes

Cam0Cow:
Changes some pin assignments as per Controls Leaderboard revisions.


babusid:
In PR #158 @NathanielDelgado implemented the same ADC Fixes, but also changed the ADC Input channels from 2 and 3 to channels 9 and 14. Those changes may need to be rolled into this PR as well, so that we can just merge this one and close that one.

Cam0Cow:
These changes were already merged into SendCarCAN, and will soon be merged into master. As such, this branch is no longer needed.

PR #157, opened 2021-10-23
Can Implementation for STM32 Board

NathanielDelgado:
1. We added an implementation for CAN for both CAN 1 and CAN 3 (CAN 3 was determined to be used in the latest iteration of the board instead of CAN2 for the motor CAN).
2. Added a test file called Test_CarCANBoard to test CAN 1
3. Modified motorcontroller.c and carcan.c to account for the changes in the BSP_CAN
NOTE: This version of BSP_CAN.h doesn't work for the simulator, we might need to work on this if necessary  (failing build in github because of this)


Cam0Cow (approved):
Looks good.

babusid:
@NathanielDelgado Is this still current?

NathanielDelgado:
@babusid Yes, it is still current.

babusid (changes requested):
@NathanielDelgado Can you leave comments in the two test files that indicate clearly what they're testing? I understand that the CarCANBoard test is testing the CANBus Send driver and is used to verify the board, and likewise for the MotorCANBoard, but I'd like that inside the files In a documentation comment.

Cam0Cow:
> @Cam0Cow is there a reason this has been left unmerged for this long? 

It couldn't be fully tested with the old revision of the board. Only once it's confirmed to work on the new board would I feel comfortable merging it.

babusid:
@NathanielDelgado @ErickCortez98 Take a look at this version of the [CAN Driver](https://github.com/lhr-solar/Controls/blob/RTOS_SendCarCan/Drivers/Src/CANbus.c). I believe that we do in fact need TX/RX event handling because as per my understanding there is a hardware queue for both sending and receiving that needs to be handled. This may not have to be added directly into this branch and can be merged on top of it, but let me know your thoughts.

chaseblock:
I tested this today and some additional changes were required. I would appreciate it if someone could take a look at the changes, which now appear to be working.

babusid:
The BSP Layer developed on this branch has been merged into the RTOS_SendCarCAN branch and has been validated (mostly) with the RTOS driver and the sendcarcan task. That Branch is being merged in PR #141. I'm going to close this for now, but let's not delete this branch just yet.

PR #158, opened 2021-10-23
FIxes to ADC

NathanielDelgado:
Modified the pins that ADC driver uses to be PC0 and PC1 since they are the pins used in the newest iteration of the leaderboard


Cam0Cow (commented):
Code looks good, can't merge until tested on new board revision.

PR #168, opened 2021-12-04
Made initial version of BlinkLights task.

babusid:
Modified CarState and Switch Headers to bring in line with the new Struct paradigms from ReadSwitches.

### DIFF
--- a/Apps/Src/BlinkLights.c
+++ b/Apps/Src/BlinkLights.c
@@ -85,7 +85,7 @@ void Task_BlinkLight(void* p_arg){
             0,
             0,
             0,
-            11, //11 milliseconds corresponds to a 90HZ 
+            650, //650 milliseconds corresponds to 92ish times per minute 
             OS_OPT_TIME_HMSM_STRICT,
              &err
         );
--- a/Apps/Src/BlinkLights.c
+++ b/Apps/Src/BlinkLights.c
@@ -1,0 +1,93 @@
+#include "Tasks.h"
+#include "Lights.h"
+
+void Task_BlinkLight(void* p_arg){
+    //get necessary state pointers
+    car_state_t *car_state = (car_state_t*) p_arg;
+    blinker_states_t *blinkerStates = &(car_state -> BlinkerStates);
+    light_switches_t *storedLightSwitches = &(car_state -> SwitchStates.lightSwitches);
+    CPU_TS ts;
+    OS_ERR err;
+
+    while(1){
+        //wait for blinkLight to get signaled
+        OSSemPend( 
+            &BlinkLight_Sem4,
+            0,
+            OS_OPT_PEND_BLOCKING,
+            &ts,
+            &err
+        );    
+        //TODO: Main light toggle logic
+
+        //Read the storedLightSwitches and toggle the blinkerSwitches that are on. If else, set to 0. At least one is on.
+
+        if(storedLightSwitches->HZD_SW = ON){ //if the hazards are on, both lights need to get toggled
+            blinkerStates->LT = OFF;
+            blinkerStates->RT = OFF;
+            switch(blinkerStates->HZD){
+                case OFF:
+                    blinkerStates->HZD = ON;
+                    break;
+                case ON:
+                    blinkerStates->HZD = OFF;
+                    break;
+            }
+        } else if (storedLightSwitches->LEFT_SW){
+            //turn off right switch and toggle the left
+            blinkerStates->RT = OFF;
+            blinkerStates->HZD = OFF;
+            switch(blinkerStates->LT){
+                case OFF:
+                    blinkerStates->LT = ON;
+                    break;
+                case ON:
+                    blinkerStates->LT = OFF;
+                    break;
+            }
+        } else if (storedLightSwitches->RIGHT_SW = ON){
+            //turn off the left switch and toggle the right
+            blinkerStates->LT = OFF;
+            blinkerStates->HZD = OFF;
+            switch(blinkerStates->RT){
+                case OFF:
+                    blinkerStates->RT = ON;
+                    break;
+                case ON:
+                    blinkerStates->RT = OFF;
+                    break;
+            }
+        } else {
+            //turn all states off
+            blinkerStates->HZD = OFF;
+            blinkerStates->LT = OFF;
+            blinkerStates->RT = OFF;
+        }
+        
+        //set the lights as defined by the toggled states here.
+        if(blinkerStates->HZD = ON){ //Hazard on cycle
+            Lights_Set(LEFT_BLINK,ON);
+            Lights_Set(RIGHT_BLINK,ON);
+        } else if (blinkerStates->LT = ON){ //Left Blink on cycle
+            Lights_Set(LEFT_BLINK,ON);
+            Lights_Set(RIGHT_BLINK,OFF);
+        } else if (blinkerStates->RT = ON){ //Right Blink on cycle
+            Lights_Set(LEFT_BLINK,OFF);
+            Lights_Set(RIGHT_BLINK,ON);
+        } else { //Off cycle
+            Lights_Set(LEFT_BLINK,OFF);
+            Lights_Set(RIGHT_BLINK,OFF);
+        } 
+
+
+        //lock thread to run at 90Hz if blinkLights needed.
+        OSTimeDlyHMSM(
+            0,
+            0,
+            0,
+            11, //11 milliseconds corresponds to a 90HZ 
+            OS_OPT_TIME_HMSM_STRICT,
+             &err
+        );
+    }
+}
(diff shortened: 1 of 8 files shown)


Apps/Src/BlinkLights.c
```diff
@@ -0,0 +1,93 @@
...
+            Lights_Set(LEFT_BLINK,OFF);
+            Lights_Set(RIGHT_BLINK,ON);
+        } else { //Off cycle
+            Lights_Set(LEFT_BLINK,OFF);
+            Lights_Set(RIGHT_BLINK,OFF);
+        } 
+
+
+        //lock thread to run at 90Hz if blinkLights needed.
+        OSTimeDlyHMSM(
```
Cam0Cow, line 84: Should be 60 to 120 times per minute (so 500 to 1000 ms delay).

Apps/Src/BlinkLights.c
```diff
@@ -0,0 +1,93 @@
...
+                case ON:
+                    blinkerStates->RT = OFF;
+                    break;
+            }
+        } else {
+            //turn all states off
+            blinkerStates->HZD = OFF;
+            blinkerStates->LT = OFF;
+            blinkerStates->RT = OFF;
+        }
```
Cam0Cow, line 65: Logic seems sound

Cam0Cow (changes requested):

Apps/Src/BlinkLights.c
```diff
@@ -0,0 +1,93 @@
...
+    //get necessary state pointers
+    car_state_t *car_state = (car_state_t*) p_arg;
+    blinker_states_t *blinkerStates = &(car_state -> BlinkerStates);
+    light_switches_t *storedLightSwitches = &(car_state -> SwitchStates.lightSwitches);
+    CPU_TS ts;
+    OS_ERR err;
+
+    while(1){
+        //wait for blinkLight to get signaled
+        OSSemPend(
```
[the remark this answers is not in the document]
babusid: If the blinkLights Sem4 isn't signaled, we should never go into this routine to blink the lights at all. That's why I've put it within the loop runtime. Do you think there's a better way to implement that?

Apps/Src/BlinkLights.c
```diff
@@ -0,0 +1,93 @@
...
+        } else if (blinkerStates->RT = ON){ //Right Blink on cycle
+            Lights_Set(LEFT_BLINK,OFF);
+            Lights_Set(RIGHT_BLINK,ON);
+        } else { //Off cycle
+            Lights_Set(LEFT_BLINK,OFF);
+            Lights_Set(RIGHT_BLINK,OFF);
+        } 
+
+
+        //lock thread to run at 90Hz if blinkLights needed.
```
ClarkPoon, line 83: This comment should be 90 times per minute (1.5 Hz), not 90 Hz, right?

Apps/Src/BlinkLights.c
```diff
@@ -0,0 +1,93 @@
...
+    while(1){
+        //wait for blinkLight to get signaled
+        OSSemPend( 
+            &BlinkLight_Sem4,
+            0,
+            OS_OPT_PEND_BLOCKING,
+            &ts,
+            &err
+        );    
+        //TODO: Main light toggle logic
```
ClarkPoon, line 21: Is this TODO still accurate, or is the main light logic the code on lines 23-80?

Apps/Src/BlinkLights.c
```diff
@@ -0,0 +1,93 @@
+#include "Tasks.h"
+#include "Lights.h"
+
+void Task_BlinkLight(void* p_arg){
+    //get necessary state pointers
+    car_state_t *car_state = (car_state_t*) p_arg;
+    blinker_states_t *blinkerStates = &(car_state -> BlinkerStates);
+    light_switches_t *storedLightSwitches = &(car_state -> SwitchStates.lightSwitches);
```
ClarkPoon, line 8: Are you just using this to implement a global variable of type `car_state_t`, or am I missing something here?

Apps/Src/BlinkLights.c
```diff
@@ -0,0 +1,93 @@
...
+    CPU_TS ts;
+    OS_ERR err;
```
ClarkPoon, line 10: Are there any safety issues if we get an OS error, or does the controls system recover fine?

Apps/Src/BlinkLights.c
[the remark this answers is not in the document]
ClarkPoon: What is this semaphore used for?

Apps/Src/BlinkLights.c
```diff
@@ -0,0 +1,93 @@
+#include "Tasks.h"
+#include "Lights.h"
+
+void Task_BlinkLight(void* p_arg){
```
ClarkPoon, line 4: What is the expected behavior of this task? It would be nice if there was a comment

ClarkPoon (commented):
I figured I'd do some code reviews on controls to get a better idea of how the system works

Apps/Src/BlinkLights.c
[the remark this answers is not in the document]
babusid: @ClarkPoon This semaphore gets signaled from the UpdateLights thread for which the source can be found [here](https://github.com/lhr-solar/Controls/blob/RTOS_UpdateLights/Apps/Src/UpdateLights.c). Essentially, if the Hazard switch or either of the indicator switches is toggled, that thread then signals this semaphore. This thread then controls the blinking on/off until it is no longer necessary.

Apps/Src/BlinkLights.c
babusid: @ClarkPoon I do not believe we currently have OS Error fault handling, does BPS address this in any particular manner? I believe the plan moving forward was to build fault handling once we've got the core application threads completed.

Apps/Src/BlinkLights.c
babusid: @ClarkPoon All of our threads upon creation will be given a pointer to our global car state struct which will eventually be instantiated in the main.c file. The definition for this struct can be found [here](https://github.com/lhr-solar/Controls/blob/master/Apps/Inc/CarState.h). This section is just extracting the relevant fields from the car state that pertain to this thread so we can work with them locally.

Apps/Src/BlinkLights.c
ClarkPoon: The BPS handles OS faults by shutting down the car using [this function.](https://github.com/lhr-solar/BPS/blob/c0aabfb1390afedae6c89bb5627b340984ad060a/Apps/Src/Tasks.c#L71) I'm not sure what the expected behavior should be for controls (eg. shut down the motor or try to recover). 

Fault handling can sometimes be helpful for debugging the application/thread layer, since the OS errors I have seen were mainly caused by bugs in the BPS's application layer (eg. from forgetting to initialize something or forgetting to release a mutex)

Apps/Src/BlinkLights.c
ClarkPoon: Ok. I noticed that you aren't using a mutex for this struct. I don't see any obvious critical sections that would require a mutex in your code, so I am guessing this is intentional. 

If you are modifying the same fields of this struct in multiple places, or relying on accessing multiple fields to be atomic, I would consider adding a mutex

Apps/Src/BlinkLights.c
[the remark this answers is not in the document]
ClarkPoon: Responding to your TODO comment:

Personally, I would not use the semaphore here since the rate you are flashing the lights probably doesn't need to synchronize with the UpdateLights thread.

I am guessing you are using the semaphore because you are worried about `LightsSet()` wasting time by setting the lights when they don't need to be set. I think the semaphore works for this as long as you keep setting it at the same rate in your other task. One potentially cleaner solution to this would be:
```C
// have some array to keep track of the states of all of the lights
// this array would need to be initialized on startup
// NUM_LIGHTS can just be an extra at the end of your light_t enum
static State lights[NUM_LIGHTS];   

void Lights_Set(light_t light, State state) {
    if (state == lights[light]) {
        return;
    else {
        lights[light] = state;
    }

    // the rest of your lights code
    . . .
}
```

Apps/Src/BlinkLights.c
babusid: I was considering the same thing (mutex locking the global state), however, as far as I know currently, each field of the car state is only modified by one thread. I am planning on inspecting the code later on to make sure that this is the case though.

Apps/Src/BlinkLights.c
babusid: I like this idea. I think ideally, each individual application tries to recover from any faults, and if not, we resort to a fault handler that shuts down the system. I'll make an issue ticket/note for this.

babusid:
BlinkLights is implemented in PR #199. Closing this now, as it is part of the outdated architecture.

Issue #109, opened 2021-01-16
[RTOS] Update Lights Task


Cam0Cow:
This RTOS task should update the various lights of the car depending on the internal car state. The state diagrams and some pseudocode can be found on the Google Drive (Electrical Team > Controls > New Design > Software > RTOS), but don't take either resource as absolute truth. This is the time to reason about this application code, so try to think about what exactly each module should be doing.

babusid:
This task needs to be deprecated and rolled into function logic that is directly called by the ReadSwitches task

PR #170, opened 2021-12-31
UpdateLights Task

babusid:
Addresses Issue #109.  Implements the Update lights thread with the semaphore signaling the BlinkLights thread architecture. This may have to be revisited if we decide to remove the signaling semaphore for BlinkLights as discussed [here][https://github.com/lhr-solar/Controls/pull/168#discussion_r776517042](https://github.com/lhr-solar/Controls/pull/168/files#r762465782)

### DIFF
--- a/Apps/Src/UpdateLights.c
+++ b/Apps/Src/UpdateLights.c
@@ -1,6 +1,6 @@
 #include "Lights.h"
 #include "Tasks.h"
-#define BRAKETHRESHOLD 2
+#define BRAKETHRESHOLD 3
 
 /**
  * @brief This task waits on the Update Lights Semaphore, then reads the switches that pertain to the lights from the global carstate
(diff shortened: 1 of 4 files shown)


Apps/Src/UpdateLights.c
```diff
@@ -0,0 +1,62 @@
...
+            Lights_Set(A_CNCTR,ON);
+        } else {
+            Lights_Set(A_CNCTR,OFF);
+        }
+
+        if(GlobalCarState->ShouldMotorBeActivated==ON){
+            Lights_Set(M_CNCTR,ON);
+        } else {
+            Lights_Set(M_CNCTR,OFF);
+        }
```
ClarkPoon, line 59:
```suggestion
        Lights_Set(M_CNCTR, GlobalCarState->ShouldMotorBeActivated);
```

Apps/Src/UpdateLights.c
```diff
@@ -0,0 +1,62 @@
...
+            Lights_Set(Headlight_ON,ON);
+        } else {
+            Lights_Set(Headlight_ON,OFF);
+        }
+
+        if(GlobalLightSwitches->HZD_SW == ON || GlobalLightSwitches->LEFT_SW == ON || GlobalLightSwitches->RIGHT_SW ==ON){
+            OSSemPost(
+                &BlinkLight_Sem4,
+                OS_OPT_POST_ALL,
+                &err
```
ClarkPoon, line 36: Where is this semaphore initialized? 

Also, I don't think this is the best way to do this. Why not just have the blink lights task run constantly at 1.5 Hz and read `GlobalLightSwitches` directly?

ClarkPoon (commented):
Why did you create `Apps/Src/BlinkLights.c` as an empty file?

babusid:
This task needs to be turned into functional logic that is integrated into the ReadSwitches task. This task will then be deprecated and deleted after ReadSwitches verification.

PR #173, opened 2022-01-22
New Task Suspension Code

### LINKED ISSUE #172: Task Suspension
The current mechanism for suspending and resuming tasks is needlessly confusing, and doesn't allow other tasks to check if a task is suspended or not. This scheme should be replaced by a variable for each task, letting it know if it should be running or not. If not, the task should stop and pend on a semaphore. Otherwise, it'll run as normal. To resume a task, another tasks should set it's variable to the running state and then signal the task's semaphore.

Cam0Cow:
Closes #172

### DIFF
--- a/Apps/Src/ArrayConnection.c
+++ b/Apps/Src/ArrayConnection.c
@@ -62,14 +62,10 @@ void Task_ArrayConnection(void *p_arg) {
 
         if (desiredState == ON && currentState != ON) {
             arrayStartup(&err); // Reactivate the array
-            car_state->ActiveTasks.readCarCAN = ON; // set readCarCAN to active again
-            OSTaskSemPost(&ReadCarCAN_TCB, OS_OPT_POST_NONE, &err); // Tell the task it can start up again
-            if(err != OS_ERR_NONE){
-                car_state->ErrorCode.ArrayErr = ON;
-            }
+            OSTaskResume(&ReadCarCAN_TCB, &err); // Resume task
         } else if (desiredState != ON && currentState == ON) {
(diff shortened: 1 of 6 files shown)


Cam0Cow:
Doesn't modify SendCarCAN, as it's not yet present in the master branch. This needs to be done before both are merged.

Cam0Cow:
Doesn't modify SendCarCAN, as it's not yet present in the master branch. This needs to be done before both are merged.

Cam0Cow:
Doesn't modify SendCarCAN, as it's not yet present in the master branch. This needs to be done before both are merged.

ClarkPoon (commented):
Looks good to me. 

Not approving yet because @Cam0Cow 's comment indicates that some other changes are needed for this to work, and I am guessing this has not been tested

babusid:
@ClarkPoon It shouldn't break SendCarCAN regardless, roie and I talked about it, just gonna do a double verify on the weekend

Apps/Src/ArrayConnection.c
```diff
@@ -62,16 +62,10 @@ void Task_ArrayConnection(void *p_arg) {
 
         if (desiredState == ON && currentState != ON) {
```
babusid, line 63: The only task I can think of that activates the array is read-switches and that hasn't been merged in yet. Will this task work without it? I see that we're accessing car_state->ShouldArrayBeActivated, and that's not initialized anywhere.

babusid (approved):
The only thing that concerns me about this is that the ArrayConnection task depe

PR #174, opened 2022-01-25
Fix issue with uCOS and a few other things.

chaseblock:


JimothyGreene (approved):
@chaseblock always able to solve the issue. Not sure how you knew this was the issue, but I'm glad you're a genius lol.

babusid (approved):
@chaseblock how did you even figure this out

ClarkPoon:
@JimothyGreene @babusid There's a register that gets set when there's a hard fault that tells you what caused the hard fault.

babusid:
> @JimothyGreene @babusid There's a register that gets set when there's a hard fault that tells you what caused the hard fault.

What register is it? Just for future reference if we hit any other hard faults

ClarkPoon:
> > @JimothyGreene @babusid There's a register that gets set when there's a hard fault that tells you what caused the hard fault.
> 
> What register is it? Just for future reference if we hit any other hard faults

I think it is the CFSR register, which I found described in [this article](https://interrupt.memfault.com/blog/cortex-m-fault-debug)

Issue #169, opened 2021-12-31
[RTOS] Global Fault Handling 


babusid:
# Description and Background
If we have an OS error that is not handled by the thread that it occurs in, we currently do not handle this properly the way BPS does. BPS's way of handling this is shown [here](https://github.com/lhr-solar/BPS/blob/c0aabfb1390afedae6c89bb5627b340984ad060a/Apps/Src/Tasks.c#L71) and [here](https://github.com/lhr-solar/BPS/blob/c0aabfb1390afedae6c89bb5627b340984ad060a/Apps/Src/Task_FaultState.c#L22). (Context provided below)

# BPS Solution
BPS has an application thread that pends on a global fault semaphore, and whenever a thread is unable to recover from an OS fault it attempts to post to that semaphore through the `assertOSError` function. This thread then runs the `EnterFaultState` function which accesses BSP (board layer code) functions to shut down the system safely. If the OS fault is such that posting to that fault semaphore is impossible, the `assertOSError` manually runs the `EnterFaultState` function. This `EnterFaultState` function is also bound to all of the board level fault interrupts (HardFault_Handler, BusFault_Handler, etc.)

# What needs to be done: 
- We're going to do the same thing BPS does
1. Create an `EnterFaultState` function that safely shuts down the motor and signals a controls fault. This function should ONLY rely on BSP level code, as we need to go directly to hardware level here.
2. Create a global fault semaphore 
3. Create a global fault Bitmap (a variable that has certain bits set depending on what kind of fault has occurred). The enterFaultState can rely upon this to do certain things in case of different fault types.
3. Create a `FaultHandler` task that pends on that semaphore
4. Create an `assertOSError` function that posts to that semaphore, and if that semaphore post fails, it then calls the EnterFaultState function directly. This function should accept an argument of the `OS_ERR` type, and if the argument provided is equivalent to `OS_ERR_NONE`, it shouldn't do anything. (This allows us to call this function in every thread without having to check that an error has actually occurred ahead of time)
5. Bind all Interrupt Fault Handlers to the EneterFaultState Function. In this case, the Fault_Bitmap should be set such that it is clear the fault has occurred due to an Interrupt Fault.

babusid:
I found an additional error case, [here](https://github.com/lhr-solar/Controls/blob/RTOS_SwitchRead/Apps/Src/ReadSwitches.c). On line 149. This wouldn't be an OS Error case but could be a different error type that we still need to assert.

[...]

PR #179, opened 2022-02-05
Fault Handling

IshDeshpa:
Fault Handler adapted to be more similar to bsp #169


babusid:
@Cam0Cow @IshDeshpa what is this merge waiting on? All OS tasks have to be completed and merged before this is finished if I'm not mistaken, but is there something else that is delaying it?

IshDeshpa:
> @Cam0Cow @IshDeshpa what is this merge waiting on? All OS tasks have to be completed and merged before this is finished if I'm not mistaken, but is there something else that is delaying it?

Don't SendCarCAN and any other branches involving OS errors have to be finished so that this branch can be updated with all error checking?

babusid:
@IshDeshpa This thread is going to need an update to conform w Rearchitecting. The contactors driver should have a function called FaultSet that sets a static field in the driver to true. This should get triggered when we pull the contactors from the fault thread. The contactors set function should not work if that field is true. 
Note: you shouldn't be using the driver to pull the contactors, but you should utilize the faultSet function once you write it.

ClarkPoon:
@IshDeshpa @babusid should we close this?

babusid:
Since we're rearchitecting, I'm going to close this PR, but I'll leave the branch open if we need it for reference.

PR #182, opened 2022-02-19
Lights Driver Rearchitecting

### LINKED ISSUE #180: [FIX] Lights Driver
Lights driver needs to actually initialize the GPIO Expander's pins over SPI. This should go in Lights_Init. Pins GPIOB 1-6 need to be configured as outputs. How this is done can be seen in the Switches Driver, but needs to be modified to fit the Lights Pins.

IshDeshpa:
closes #180

### DIFF
--- a/Drivers/Src/Lights.c
+++ b/Drivers/Src/Lights.c
@@ -9,8 +9,12 @@
 */
 
 
-
 #include "Lights.h"
+#include "common.h"
+#include <unistd.h>
+#include "BSP_SPI.h"
+#include "stm32f4xx.h"
+
 
 /**
 * @brief   Initialize Lights Module
@@ -19,6 +23,24 @@
 void Lights_Init(void) {
     BSP_GPIO_Init(LIGHTS_PORT, 0x3C0, 1); // Pins 6,7,8,9 from Port C are out (0b1111000000)
     BSP_SPI_Init();
+
+    uint8_t txReadBuf[2] = {SPI_OPCODE_R, 0x01}; //0x01 is IODIRB in Bank 0 Mode
+    uint8_t rxBuf[1] = {0};
+    
+    GPIO_WriteBit(GPIOA, GPIO_Pin_4, Bit_RESET);
+    BSP_SPI_Write(txReadBuf, 2);
+    BSP_SPI_Read(rxBuf,1);
+    GPIO_WriteBit(GPIOA, GPIO_Pin_4, Bit_SET);
+
+    uint8_t txWriteBuf[3] = {
+        SPI_OPCODE_W, //write to IODIRB
+        0x01, 
+        rxBuf[0]&0x40 //clear IODIRB to set as outputs except for hazard lights switch pin (bit 6, 0x40)
+    };
+    
+    GPIO_WriteBit(GPIOA, GPIO_Pin_4, Bit_RESET);
+    BSP_SPI_Write(txWriteBuf, 3);
+    GPIO_WriteBit(GPIOA, GPIO_Pin_4, Bit_SET);
 }
 
 /**
@@ -28,11 +50,13 @@ void Lights_Init(void) {
 */ 
 State Lights_Read(light_t light) {
     // Get internal lights from SPI
-    uint8_t txBuf[3] = {SPI_OPCODE_R, SPI_GPIOB, 0x00};
+    uint8_t txBuf[2] = {SPI_OPCODE_R, SPI_GPIOB};
     uint8_t rxBuf[2] = {0};
-    BSP_SPI_Write(txBuf, 3);
     
+    GPIO_WriteBit(GPIOA, GPIO_Pin_4, Bit_RESET);
+    BSP_SPI_Write(txBuf, 2);
     BSP_SPI_Read(rxBuf, 2);
+    GPIO_WriteBit(GPIOA, GPIO_Pin_4, Bit_SET);
     
     uint8_t currData = rxBuf[1];
     // Different depending on the light
@@ -52,11 +76,14 @@ State Lights_Read(light_t light) {
  */
 void Lights_Set(light_t light, State state) {
     // Get internal lights from SPI
-    uint8_t txBuf[3] = {SPI_OPCODE_R, SPI_GPIOB, 0x00};
+    uint8_t txReadBuf[2] = {SPI_OPCODE_R, SPI_GPIOB};
+    uint8_t txWriteBuf[3] = {SPI_OPCODE_W, SPI_GPIOB, 0x00};
     uint8_t rxBuf[2] = {0};
-    BSP_SPI_Write(txBuf, 3);
-
+    
+    GPIO_WriteBit(GPIOA, GPIO_Pin_4, Bit_RESET);
+    BSP_SPI_Write(txReadBuf, 2);
     BSP_SPI_Read(rxBuf, 2);
+    GPIO_WriteBit(GPIOA, GPIO_Pin_4, Bit_SET);
     
     uint8_t portc = BSP_GPIO_Read(LIGHTS_PORT);
     if (light == BrakeLight) {
@@ -66,8 +93,8 @@ void Lights_Set(light_t light, State state) {
         uint8_t currData = rxBuf[1];
         uint8_t newData = currData & ~(0x01 << light);
         newData |= state << light;
-        txBuf[0] = SPI_OPCODE_W;
-        txBuf[2] = newData;
+        
+        txWriteBuf[2] = newData;
 
         // Lights that are external and internal
         switch (light) {
@@ -87,6 +114,10 @@ void Lights_Set(light_t light, State state) {
                 break;
         }
     }
-    BSP_SPI_Write(txBuf, 3);
+
+    GPIO_WriteBit(GPIOA, GPIO_Pin_4, Bit_RESET);
+    BSP_SPI_Write(txWriteBuf, 3);
+    GPIO_WriteBit(GPIOA, GPIO_Pin_4, Bit_SET);
+    
     BSP_GPIO_Write(LIGHTS_PORT, portc);
 }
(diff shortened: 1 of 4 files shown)


Drivers/Src/Lights.c
```diff
@@ -52,11 +76,14 @@ State Lights_Read(light_t light) {
...
-    uint8_t txBuf[3] = {SPI_OPCODE_R, SPI_GPIOB, 0x00};
+    uint8_t txReadBuf[2] = {SPI_OPCODE_R, SPI_GPIOB};
+    uint8_t txWriteBuf[3] = {SPI_OPCODE_W, SPI_GPIOB, 0x00};
     uint8_t rxBuf[2] = {0};
-    BSP_SPI_Write(txBuf, 3);
-
+    
+    GPIO_WriteBit(GPIOA, GPIO_Pin_4, Bit_RESET);
+    BSP_SPI_Write(txReadBuf, 2);
     BSP_SPI_Read(rxBuf, 2);
```
babusid, line 85: Why are you giving this a length of 2 bytes? GPIOB is only one byte long

Drivers/Src/Lights.c
```diff
@@ -66,8 +93,8 @@ void Lights_Set(light_t light, State state) {
         uint8_t currData = rxBuf[1];
```
babusid, line 93: Shouldn't this be rxBuf[0]?

babusid: It *might* put it in buf[1] since you passed in a length of two bytes, but I'm not sure. I would highly recommend making the rxBuf one byte long because that's all it ever needs.

babusid (changes requested):
Make sure RxBuf Lengths are only one byte

Drivers/Src/Lights.c
```diff
@@ -52,22 +76,25 @@ State Lights_Read(light_t light) {
...
+    uint8_t txReadBuf[2] = {SPI_OPCODE_R, SPI_GPIOB};
+    uint8_t txWriteBuf[3] = {SPI_OPCODE_W, SPI_GPIOB, 0x00};
+    uint8_t rxBuf[1] = {0};
+    
+    GPIO_WriteBit(GPIOA, GPIO_Pin_4, Bit_RESET);
+    BSP_SPI_Write(txReadBuf, 2);
+    BSP_SPI_Read(rxBuf, 1);
+    GPIO_WriteBit(GPIOA, GPIO_Pin_4, Bit_SET);
     
     uint8_t portc = BSP_GPIO_Read(LIGHTS_PORT);
```
NathanielDelgado, line 88: @IshDeshpa Should be uint16_t or else the left and right blink lights won't be affected because they are bits 8 and 9 in a 16 bit port

NathanielDelgado (dismissed):
Make changes

Drivers/Src/Lights.c
IshDeshpa: Fixed

babusid:
Since we're going to have to make lights rearchitecting conformant anyways, I'd like this pull request renamed to Lights Driver Rearchitecting.

Furthermore, Lights_Set is going to have to be made threadsafe. Lights_read could be made massively simpler as well (use a static array on this layer to store all light states as we're setting them, just do an array lookup instead of going and reading the port.) I also want a lights_toggle function that looks up that particular light in the array and uses the newly threadsafe lights_set to set it to the opposite value.

Drivers/Src/Lights.c
```diff
@@ -9,16 +9,59 @@
...
+
+    OSMutexPend(&lightMutex,0,OS_OPT_PEND_BLOCKING,&timestamp,&err); //lock light hardware for initialization (TODO: may not be needed, just here for additional safety)
+    if(err != OS_ERR_NONE){
+        assertOSError(OS_BLINK_LIGHTS_LOC, err);
+    }
+
     BSP_GPIO_Init(LIGHTS_PORT, 0x3C0, 1); // Pins 6,7,8,9 from Port C are out (0b1111000000)
     BSP_SPI_Init();
+
+    uint8_t txReadBuf[2] = {SPI_OPCODE_R, 0x01}; //0x01 is IODIRB in Bank 0 Mode
```
ClarkPoon, line 43: It might be nice if there was an `enum` or a `#define` for this instead of just `0x01`

Drivers/Src/Lights.c
```diff
@@ -9,16 +9,59 @@
...
+    uint8_t rxBuf[1] = {0};
+    
+    GPIO_WriteBit(GPIOA, GPIO_Pin_4, Bit_RESET);
+    BSP_SPI_Write(txReadBuf, 2);
+    BSP_SPI_Read(rxBuf,1);
+    GPIO_WriteBit(GPIOA, GPIO_Pin_4, Bit_SET);
```
ClarkPoon, line 49: I think it would be more readable if you had a high level comment for each block of code to say what it does

babusid:
We need a `Toggle_Enable(light_t light)` function and a `Toggle_Disable(light_t light)` function. This will require an additional bitmap on this layer that maintains the enable/disable states of all the toggleable lights. We will also need a function called `Toggle_Read()` that returns the enable/disable state bitmap. This last function will be used by the BlinkLights thread in order to toggle lights on and off by reading the enable/disable bitmap, and the current states using Lights_Read and then changing the states using Lights_set. ReadSwitches will utilize Toggle_Enable and Toggle_Disable to indicate that the light needs to begin its toggle sequence.

Drivers/Src/Lights.c
```diff
@@ -51,42 +81,66 @@ State Lights_Read(light_t light) {
...
-                portc &= ~(0x01 << RIGHT_BLINK_PIN);    // Clear bit corresponding to pin
-                portc |= (state << RIGHT_BLINK_PIN);    // Set value to inputted state
-                break;
-            case Headlight_ON:
-                portc &= ~(0x01 << HEADLIGHT_PIN);  // Clear bit corresponding to pin
-                portc |= (state << HEADLIGHT_PIN);  // Set value to inputted state
-                break;
-            default:
-                break;
+        GPIO_WriteBit(GPIOA, GPIO_Pin_4, Bit_RESET);
```
ClarkPoon, line 126: Is this your chip select pin?

This is probably beyond the scope of this PR, but I think it would be nice if this was more explicit (eg. have a function called `BSP_SPI_SetCS()`)

Drivers/Src/Lights.c
```diff
@@ -9,16 +9,45 @@
...
 
 
-
 #include "Lights.h"
+#include "common.h"
+#include <unistd.h>
+#include "BSP_SPI.h"
+#include "stm32f4xx.h"
+
+static State lightStates[10] = {OFF};
```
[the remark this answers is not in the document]
IshDeshpa: All of them; at the car's initialization all of the lights start as off. However, this has been changed to a bitmap

Drivers/Src/Lights.c
```diff
@@ -9,16 +9,59 @@
...
+    OSMutexPend(&lightMutex,0,OS_OPT_PEND_BLOCKING,&timestamp,&err); //lock light hardware for initialization (TODO: may not be needed, just here for additional safety)
+    if(err != OS_ERR_NONE){
+        assertOSError(OS_BLINK_LIGHTS_LOC, err);
+    }
+
     BSP_GPIO_Init(LIGHTS_PORT, 0x3C0, 1); // Pins 6,7,8,9 from Port C are out (0b1111000000)
     BSP_SPI_Init();
+
+    uint8_t txReadBuf[2] = {SPI_OPCODE_R, 0x01}; //0x01 is IODIRB in Bank 0 Mode
+    uint8_t rxBuf[1] = {0};
```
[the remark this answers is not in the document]
IshDeshpa: Fixed.

Drivers/Src/Lights.c
IshDeshpa: Done.

Drivers/Src/Lights.c
```diff
@@ -9,16 +9,60 @@
...
     BSP_SPI_Init();
+
+    // Initialize txBuf and rxBuf
+    uint8_t txReadBuf[2] = {SPI_OPCODE_R, SPI_IODIRB}; //0x01 is IODIRB in Bank 0 Mode
+    uint8_t rxBuf = 0;
+    
+    // Reads direction register and stores in rxBuf
+    GPIO_WriteBit(GPIOA, GPIO_Pin_4, Bit_RESET);
+    BSP_SPI_Write(txReadBuf, 2);
+    BSP_SPI_Read(rxBuf,1);
```
ClarkPoon, line 45:
```suggestion
    BSP_SPI_Read(&rxBuf,1);
```
If you change rxBuf to a `uint8_t`, you will have to change this as well

Drivers/Src/Lights.c
```diff
@@ -27,21 +71,34 @@ void Lights_Init(void) {
...
+* @brief   Read the state of the lights
+* @param   light Which Light to read
+* @return  returns state enum which indicates ON/OFF
+*/
+State Toggle_Read(light_t light) {
+    return (lightToggleBitmap>>light)&0x01;
+}
+
+/**
+ * @brief   Read toggle bitmap
```
babusid, line 97: Read the toggleable lights bitmap

Drivers/Src/Lights.c
```diff
@@ -9,16 +9,60 @@
...
 
 
-
 #include "Lights.h"
+#include "BSP_SPI.h"
+#include "stm32f4xx.h"
+#include "Tasks.h"
+
+//static State lightStates[10] = {OFF};
+static uint8_t lightStatesBitmap;
```
babusid: Leave a comment indicating this is the current state of all lights

Drivers/Src/Lights.c
```diff
@@ -9,16 +9,60 @@
...
+static uint8_t lightToggleBitmap;
```
babusid: Leave a comment indicating that this is the enable/disable state of toggleable lights. Only certain bits will ever be high and it corresponds to the toggleable lights.

babusid (approved):

PR #183, opened 2022-02-19
Switches Driver Fixes

NathanielDelgado:
- Changed the switches ordering from the enum and the ordering in the switch cases in SwitchRead
- Wrapped CS changes around SPI writes and read
- Changed the number of buffers used
- Fixed the Init and Read logic


babusid:
This needs to get tested first thing after ReadCarCAN on Saturday. I'll try to potentially test it on Wednesday if possible

Drivers/Src/Switches.c
```diff
@@ -1,7 +1,11 @@
 #include "Switches.h"
+#include "stm32f4xx.h"
+#include "os.h"
 
-
-
+static OS_MUTEX SwitchMutex; //Mutex to lock SPI lines
```
babusid, line 5: This mutex needs to actually get initialized before it starts being used

Drivers/Src/Switches.c
```diff
@@ -13,32 +17,50 @@
  * @return  None
  */ 
 void Switches_Init(void){
+    //TODO: Sync isn't expressly needed here because Init will only ever be called once from the main task, however I added it just in case. Removal should be fine.
+    OSMutexPend(
```
babusid, line 21: Usage of mutex before initialization

Drivers/Src/Switches.c
```diff
@@ -48,46 +70,117 @@ void Switches_Init(void){
...
+            &SwitchMutex,
+            0,
+            OS_OPT_PEND_BLOCKING,
+            &timestamp,
+            &err
+        );
+        if(err != OS_ERR_NONE){
+            return ERROR; //Os error, could not properly lock the Minion SPI Line
+        }
+        GPIO_WriteBit(GPIOA, GPIO_Pin_4, Bit_RESET);
```
ClarkPoon, line 95: This function sets the GPIO pin atomically, so you only need the mutexas around this if you want setting all of the GPIO pins together to be atomic

Drivers/Inc/Switches.h
```diff
@@ -7,7 +7,7 @@
 #include "GPIOExpander.h"
 
 
-typedef enum {CRUZ_SW=0, CRUZ_EN, HZD_SQ, FR_SW, HEADLIGHT_SW, LEFT_SW, RIGHT_SW, REGEN_SW, IGN_1, IGN_2, REV_SW} switches_t;
+typedef enum {CRUZ_ST=0, CRUZ_EN, REV_SW, FOR_SW, HEADLIGHT_SW, LEFT_SW, RIGHT_SW, REGEN_SW, HZD_SW, IGN_1, IGN_2} switches_t;
```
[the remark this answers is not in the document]
babusid: the Cruise turn on switch that toggles cruise control logic whenever cruise enable is on

Drivers/Src/Switches.c
```diff
@@ -13,32 +17,43 @@
...
     initTxBuf[0]=SPI_OPCODE_W;
+    GPIO_WriteBit(GPIOA, GPIO_Pin_4, Bit_RESET);
     BSP_SPI_Write(initTxBuf,3);
-
+    GPIO_WriteBit(GPIOA, GPIO_Pin_4, Bit_SET);
+    OSMutexPost(
+        &SwitchMutex,
+        OS_OPT_POST_NONE,
+        &err
+    );
```
babusid, line 56: This mutex is never pended on, I don't think we need this post right here. Let's remove it

Drivers/Src/Switches.c
babusid: The mutex lock is primarily for the SPI send, and I wanted to group the entire chip select - query - chip deselect all in one runthrough

babusid (changes requested):

Drivers/Src/Switches.c
```diff
@@ -13,32 +18,47 @@
...
     BSP_SPI_Init();
     //Sets up pins 0-7 on GPIOA as input 
-    uint8_t initTxBuf[3]={SPI_OPCODE_R, SPI_IODIRA, 0x00};
-    uint8_t initRxBuf[2] = {0}; 
-    BSP_SPI_Write(initTxBuf,3);
-    do{
-        BSP_SPI_Read(initRxBuf, 2);
-    }while(initRxBuf[1] == SPI_IODIRA);
+    uint8_t initTxBuf[3]={SPI_OPCODE_R, SPI_IODIRA, 0};
+    uint8_t initRxBuf[1] = {0};
```
manthanand, line 28: If this is a one element array, why not just make it a uint8_t?

Drivers/Src/Switches.c
```diff
@@ -13,32 +18,47 @@
...
+    GPIO_WriteBit(GPIOA, GPIO_Pin_4, Bit_RESET);
     BSP_SPI_Write(initTxBuf,3);
+    GPIO_WriteBit(GPIOA, GPIO_Pin_4, Bit_SET);
 
     //Sets up pin 7 on GPIOB as input (for ReverseSwitch)
+    initTxBuf[0]=SPI_OPCODE_R;
     initTxBuf[1] = SPI_IODIRB;
+    initTxBuf[2] = 0;
     initRxBuf[0]=0;
     initRxBuf[1]=0;
```
manthanand, line 45: How are we writing to the second element of the array when it only has one element.

Drivers/Src/Switches.c
```diff
@@ -48,46 +68,109 @@ void Switches_Init(void){
  * @return  ON/OFF State
  */ 
 State Switches_Read(switches_t sw){
-    uint8_t query[3]={SPI_OPCODE_R,SPI_GPIOA,0x00}; //query GPIOA
-    uint8_t SwitchReadData[2] = {0};
+    uint8_t query[3]={SPI_OPCODE_R,SPI_GPIOA, 0}; //query GPIOA
+    uint8_t SwitchReadData[1] = {0};
```
manthanand, line 72: Same, why not just make this a variable.

Drivers/Src/Switches.c
```diff
@@ -48,46 +68,109 @@ void Switches_Init(void){
...
     case CRUZ_EN:
-    case HZD_SQ: 
-    case FR_SW: 
+    case REV_SW: 
+    case FOR_SW: 
     case HEADLIGHT_SW: 
     case LEFT_SW: 
     case RIGHT_SW: 
     case REGEN_SW:
+        OSMutexPend(
```
manthanand, line 83: Do we need to pend and post before and after GPIO bit sets?

Drivers/Src/Switches.c
```diff
@@ -1,7 +1,12 @@
 #include "Switches.h"
+#include "stm32f4xx.h"
+#include "os.h"
+#include "Tasks.h"
```
ClarkPoon, line 4: I'm not a huge fan of having the driver layer depend on the apps layer.

I think it's beyond the scope of this pull request to fix this though, since this is relatively common in the codebase.

Drivers/Src/Switches.c
```diff
@@ -1,7 +1,12 @@
 #include "Switches.h"
+#include "stm32f4xx.h"
```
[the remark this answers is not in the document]
IshDeshpa: Noted. Keeping it to the stm32 specific stuff for right now because we know it works. Once Lights and Switches are merged we can go through and replace with BSP stuff. Also, the BSP_GPIO_Write_Pin function in BSP_GPIO.c is not correctly implemented, so I'm going to fix that.

ClarkPoon (approved):

Drivers/Src/Switches.c
```diff
@@ -13,88 +16,107 @@
...
     //Sets up pins 0-7 on GPIOA as input 
-    uint8_t initTxBuf[3]={SPI_OPCODE_R, SPI_IODIRA, 0x00};
-    uint8_t initRxBuf[2] = {0}; 
-    BSP_SPI_Write(initTxBuf,3);
-    do{
-        BSP_SPI_Read(initRxBuf, 2);
-    }while(initRxBuf[1] == SPI_IODIRA);
+    uint8_t initTxBuf[3]={SPI_OPCODE_R, SPI_IODIRA, 0};
+    uint8_t initRxBuf = 0;
+    BSP_GPIO_Write_Pin(PORTA, GPIO_Pin_4, OFF);
```
tiandahuang, line 26: PORTA, GPIO_Pin_4 should probably be replaced with a
`#define SPI_PORT PORTA`
`#define SPI_CS_PIN GPIO_Pin_4` (or something similar)
for readability

Drivers/Src/Switches.c
```diff
@@ -19,125 +16,108 @@ static CPU_TS timestamp;
...
+ *          ignition switch values
+ */ 
+void Switches_Update(void){
+    OS_ERR err;
+    CPU_TS timestamp;
+    uint8_t query[2]={SPI_OPCODE_R,SPI_GPIOA}; //query GPIOA
+    uint8_t SwitchDataReg1 = 0, SwitchDataReg2 = 0;
+
+    //Read all switches except for ignition and hazard
+    BSP_GPIO_Write_Pin(PORTA, GPIO_Pin_4, OFF);
```
ClarkPoon: I assume this is the chip select pin. If so, this is part of the critical section and should be inside of the mutex

Drivers/Src/Switches.c
```diff
@@ -1,7 +1,10 @@
 #include "Switches.h"
+#include "os.h"
+#include "Tasks.h"
+#include "BSP_GPIO.h"
 
-
-
+static OS_MUTEX CommMutex; //Mutex to lock SPI lines
+static uint16_t SwitchStates_Bitmap;
```
babusid, line 7: We should initialize this to all zeroes in case we do a read before we do a write

Drivers/Src/Switches.c
```diff
@@ -13,88 +16,112 @@
...
-        if (SwitchReadData[1] & (1 << 7)) {
-            return ON;
-        } else {
-            return OFF;
-        }
+/**
+ * @brief   Sends SPI messages to read switches values. Also reads from GPIO's for 
+ *          ignition switch values
+ */ 
+void Switches_Update(void){
```
babusid, line 83: I would suggest naming this:
```suggestion
void Switches_UpdateStates(void){
```
it doesn't really matter though

babusid (dismissed):
🤠

babusid:
@manthanand can u resolve merge conflicts

Drivers/Src/Switches.c
```diff
@@ -13,88 +16,112 @@
...
-        BSP_SPI_Write(query,3);
-        do{
-            BSP_SPI_Read(SwitchReadData,2);
-        }while(SwitchReadData[0] == SPI_GPIOA);
-        if (SwitchReadData[1] & (1 << sw)) {
-            return ON;
-        } else {
-            return OFF;
-        }
+    return (State) ((SwitchStates_Bitmap >> sw) && 0x0001);
```
Cam0Cow, line 76: This should probably be & and not &&.

Cam0Cow (changes requested):

Drivers/Src/Switches.c
babusid:
```suggestion
    return (State) ((SwitchStates_Bitmap >> sw) & 0x0001);
```

babusid: @manthanand

Cam0Cow (approved):

PR #190, opened 2022-02-26
Fault task

Cam0Cow:
New fault handling task.

### DIFF
--- a/Apps/Inc/FaultState.h
+++ b/Apps/Inc/FaultState.h
@@ -4,4 +4,6 @@
 #include "os.h"
 #include "Tasks.h"
 
+void EnterFaultState(void);
+
 #endif
--- a/Apps/Inc/FaultState.h
+++ b/Apps/Inc/FaultState.h
@@ -1,0 +1,7 @@
+#ifndef __FAULT_STATE_H
+#define __FAULT_STATE_H
+
+#include "os.h"
+#include "Tasks.h"
+
+#endif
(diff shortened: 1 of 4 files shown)


Apps/Inc/FaultState.h
```diff
@@ -0,0 +1,9 @@
+#ifndef __FAULT_STATE_H
+#define __FAULT_STATE_H
+
+#include "os.h"
+#include "Tasks.h"
+
```
ClarkPoon, line 6: Is there a reason you don't just include these in FaultState.c?

Apps/Src/FaultState.c
```diff
@@ -0,0 +1,48 @@
...
+}
+
+static void MotorKill(void) {
+    Contactors_Set(MOTOR, OFF);
+}
+
+void EnterFaultState(void) {
+    if(FaultBitmap.Fault_OS){
+        ArrayKill();
+        MotorKill();
```
ClarkPoon, line 15: If you always call these together, why not just write a function that kills both?

Apps/Src/FaultState.c
```diff
@@ -0,0 +1,48 @@
...
+    else if(FaultBitmap.Fault_READBPS){
+        ArrayKill();
+        MotorKill();
+    }
+    else if(FaultBitmap.Fault_UNREACH){
+        ArrayKill();
+        MotorKill();
+    }
+    else if(FaultBitmap.Fault_DISPLAY){
+        // TODO: What should be done during a display fault?
```
ClarkPoon, line 30: What can cause a display fault?

ClarkPoon (approved):

PR #191, opened 2022-02-26
Thread Safe Motor Controller Driver

manthanand:
Made motor controller thread safe. Removed redundant structs and added error functionality to driver. Created bitmap for motor error


Drivers/Src/MotorController.c
```diff
@@ -11,13 +13,78 @@
...
+static float CurrentVelocity;
+
+uint16_t Motor_FaultBitmap = 0;
+
+/**
+ * @brief   Assert Error if Tritium sends error. When Fault Bitmap is set,
+ *          and semaphore is posted, Fault state will run.
+ * @param   motor_err Bitmap which has motor error codes
+ */
+void assertTritiumError(uint16_t motor_err){
```
ClarkPoon, line 28:
```suggestion
static void assertTritiumError(uint16_t motor_err){
```

Drivers/Src/MotorController.c
```diff
@@ -27,6 +94,8 @@ void MotorController_Init(){
  * @return  None
  */ 
 void MotorController_Drive(float newVelocity, float motorCurrent){
+    CPU_TS ts;
+	OS_ERR err;
 
     uint32_t nv = *((uint32_t *)((void *) &newVelocity));
     uint32_t mc = *((uint32_t *)((void *) &motorCurrent));
```
ClarkPoon, line 101: I feel like the code for copying the floats into an array could be better. Maybe something like
```
uint8_t data[8];
memcpy(data, &motorCurrent, sizeof(motorCurrent));
memcpy(data + sizeof(motorCurrent), &newVeloctiy, sizeof(newVelocity));
```

Drivers/Src/MotorController.c
```diff
@@ -89,36 +178,66 @@ ErrorStatus MotorController_Read(CANbuff *message, car_state_t *car){
...
-                if(MASK_CC_ERR & firstSum)
-                {
-                    car->MotorErrorCode.CCVelocityErr = ON;
-                }
-
-                if(MASK_OVER_SPEED_ERR & firstSum)
-                {
-                    car->MotorErrorCode.overSpeedErr = ON;
-                }
+                if(MASK_MOTOR_TEMP_ERR & firstSum) Motor_FaultBitmap |= T_TEMP_ERR;
```
ClarkPoon, line 181: Personally, I prefer the newline and curly braces for if statements, even if there's just one line because I think it's easier to read

Drivers/Src/MotorController.c
```diff
@@ -89,36 +178,66 @@ ErrorStatus MotorController_Read(CANbuff *message, car_state_t *car){
...
-                car->CurrentVelocity = convert.f;
+
+                OSMutexPend(&MotorController_VelocityMutex,
+				0,
+				OS_OPT_PEND_BLOCKING,
+				&ts,
+				&err);
+	            assertOSError(0, err);
+
+                CurrentVelocity = convert.f;
```
ClarkPoon, line 203: This is already atomic. You don't need a mutex for this

Drivers/Src/MotorController.c
```diff
@@ -89,36 +178,66 @@ ErrorStatus MotorController_Read(CANbuff *message, car_state_t *car){
...
+	OS_ERR err;
+    
+    OSMutexPend(&MotorController_VelocityMutex,
+	            0,
+				OS_OPT_PEND_BLOCKING,
+				&ts,
+				&err);
+	assertOSError(0, err);
+
+    float val = CurrentVelocity;
```
ClarkPoon, line 236: This is already atomic

BSP/STM32F413/Src/BSP_CAN.C
```diff
@@ -0,0 +1,421 @@
+/* Copyright (c) 2020 UT Longhorn Racing Solar */
```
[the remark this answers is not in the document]
manthanand: different branch of controls

Drivers/Src/MotorController.c
manthanand: Should we leave it just in case we have an update thread and a read thread? Or should I just remove it anyways. I don't think it will hurt to leave it in there.

Drivers/Src/MotorController.c
```diff
@@ -27,39 +88,52 @@ void MotorController_Init(){
...
 void MotorController_Drive(float newVelocity, float motorCurrent){
+    CPU_TS ts;
+	OS_ERR err;
 
     uint32_t nv = *((uint32_t *)((void *) &newVelocity));
     uint32_t mc = *((uint32_t *)((void *) &motorCurrent));
 
-    uint8_t data[8] = {0};
+    uint8_t data[8];
     int index = 0;
```
ClarkPoon, line 98: You can delete `index`, `nv`, and `mc`, since they aren't used anymore

babusid:
> why no build

It seems to not build properly on here, but does it build locally?

babusid:
BSP_CAN.C was an incorrect file, it should be called BSP_CAN.c with a lowercase C. There are BSP changes here that are made redundant by the changes that are in SendCarCAN PR, and some assertion stuff was merged into master that's also in here. I'm going to hold off on merging this until I've merged RTOS_SendCarCAN and master is either merged into this or rebased off of.

Drivers/Src/MotorController.c
```diff
@@ -11,13 +13,71 @@
 #define MASK_CC_ERR 1<<2 //checks velocity on 2 bit
 #define MASK_OVER_SPEED_ERR 1<<24 //check if motor overshot max RPM on 24 bit
 
+static OS_SEM	MotorController_MailSem4;
+static OS_SEM	MotorController_ReceiveSem4;
+static float CurrentVelocity;
```
babusid, line 18: We should initialize the CurrentVelocity to 0, because if we try to access it before a MotorController_Read gets a MOTOR_DRIVE message and updates it, we'll have undefined behavior

babusid (changes requested):

babusid (approved):
Looks good to me. I'll merge it in once we get another pair of eyes on it.

Drivers/Src/MotorController.c
```diff
@@ -11,13 +12,71 @@
 #define MASK_CC_ERR 1<<2 //checks velocity on 2 bit
 #define MASK_OVER_SPEED_ERR 1<<24 //check if motor overshot max RPM on 24 bit
 
+static OS_SEM	MotorController_MailSem4;
+static OS_SEM	MotorController_ReceiveSem4;
+static float CurrentVelocity = 0;
+
+uint16_t Motor_FaultBitmap = 0;
```
ClarkPoon, line 19: Would be more readable if this was initialized to `T_NONE` instead of `0` to be consistent with `assertTritiumError()`

Drivers/Src/MotorController.c
```diff
@@ -89,36 +157,44 @@ ErrorStatus MotorController_Read(CANbuff *message, car_state_t *car){
...
                 }
 
-                if(MASK_OVER_SPEED_ERR & firstSum)
-                {
-                    car->MotorErrorCode.overSpeedErr = ON;
+                if(MASK_OVER_SPEED_ERR & firstSum){
+                    Motor_FaultBitmap |= T_OVER_SPEED_ERR;
                 }
+
+                assertTritiumError(Motor_FaultBitmap);
```
ClarkPoon, line 176: This part technically isn't thread-safe, since `Motor_FaultBitmap` is a global and you are modifying it non-atomically. I don't think it matters though unless you call this function from multiple threads and care about distinguishing between different motor errors (neither of which I think y'all are doing)

Drivers/Src/MotorController.c
```diff
@@ -27,39 +86,48 @@ void MotorController_Init(){
...
+	OSSemPend(&MotorController_ReceiveSem4,
+			  0,    
+			  OS_OPT_PEND_BLOCKING,
+			  &ts,
+			  &err);
+	assertOSError(0, err);
+    uint32_t length = BSP_CAN_Read(CAN_3, &id, data);
 
     if(length>0){
         message->id = id;
```
ClarkPoon, line 133: Lines 133-154 could use some cleaning up, but it might not be the best use of our time.

For example, 
```C
        //get first number (bits 0-31)
        for(int j = 0; j < MAX_CAN_LEN/2; j++){
            firstSum += data[j];
            if(j != MAX_CAN_LEN/2 - 1){
                firstSum = firstSum << 8;
            }
        }
```
could be replaced by
```C
        //get first number (bits 0-31)
        for(int j = 0; j < MAX_CAN_LEN/2; j++){
            firstSum <<= 8;
            firstSum += data[j];
        }
```
and `union convert` doesn't look like it is used anywhere

Drivers/Src/MotorController.c
```diff
@@ -11,13 +12,71 @@
 #define MASK_CC_ERR 1<<2 //checks velocity on 2 bit
 #define MASK_OVER_SPEED_ERR 1<<24 //check if motor overshot max RPM on 24 bit
```
ClarkPoon, line 13: I think we should put these inside of parentheses or declare them as `const uint32_t`, so we don't get any nasty surprises when the preprocessor pastes these in and the `<<` operator has lower precedence than some other operator

Drivers/Src/MotorController.c
manthanand: convert.f is used for velocity

ClarkPoon: We could replace
```C
convert.n = secondSum;
CurrentVelocity = convert.f;
```
with
```C
memcpy(&CurrentVelocity, &secondSum, sizeof(CurrentVelocity));
```
or
```C
CurrentVelocity = *(float *) &secondSum;
```
and delete the union to make it more readable, but I'm fine with it if you leave it how it is.

ClarkPoon (dismissed):

Cam0Cow (dismissed):

Cam0Cow (dismissed):

Drivers/Src/MotorController.c
```diff
@@ -27,98 +87,105 @@ void MotorController_Init(){
...
+    CPU_TS ts;
+	OS_ERR err;
+	
+	// Check to see if a mailbox is available: BLOCKING
+	OSSemPend(&MotorController_ReceiveSem4,
+			  0,    
+			  OS_OPT_PEND_BLOCKING,
+			  &ts,
+			  &err);
+	assertOSError(0, err);
```
babusid, line 130: This indentation is kinda weird

Drivers/Src/MotorController.c
babusid: I think it does matter because we have different behaviors in our fault thread for different motor faults, so I think it does matter, because we don't want a non-critical error to beat a critical error to the fault thread.

babusid (changes requested):

Drivers/Src/MotorController.c
```diff
@@ -1,124 +1,214 @@
...
+
+                assertTritiumError(Motor_FaultBitmap);
+                
                 break;
             }
             case MOTOR_DRIVE: {
-                convert.n = secondSum;
-                car->CurrentVelocity = convert.f;
+                CurrentVelocity = *(float *) &secondSum;
                 break;
```
babusid, line 197: I don't think we'll ever recieve a MOTOR_DRIVE command, we only send it. I don't see handling for the actual messages we receive as outlined in the tritium datasheet.

babusid: We will be receiving a MOTOR_VELOCITY message, I think the case statement got messed up here.

Cam0Cow (dismissed):

Drivers/Src/MotorController.c
```diff
@@ -1,124 +1,217 @@
...
+	
+	// Check to see if a mailbox is available: BLOCKING
+	OSSemPend(
+        &MotorController_ReceiveSem4,
+	    0,    
+		OS_OPT_PEND_BLOCKING,
+		&ts,
+		&err);
+	assertOSError(0, err);
+    uint32_t length = BSP_CAN_Read(CAN_3, &id, data);
```
babusid, line 157: This doesn't return a length, BSP_CAN_Read returns an ERRORSTATUS enum value

Drivers/Inc/MotorController.h
```diff
@@ -13,26 +13,46 @@ typedef struct
...
+    T_CC_VEL_ERR = 0x02,
+    T_SLIP_SPEED_ERR = 0x04,
+    T_OVER_SPEED_ERR = 0x08,
+    T_INIT_FAIL = 0x16
+} tritium_error_code_t;
+
 /**
  * @brief   Initializes the motor controller
- * @param   None
+ * @param   busCurrentPercentSetPoint fraction of the bus current to allow the motor to draw
```
ClarkPoon, line 32:
```suggestion
 * @param   busCurrentFractionalSetPoint fraction of the bus current to allow the motor to draw
```

Drivers/Src/MotorController.c
```diff
@@ -1,124 +1,217 @@
...
+        assertTritiumError(Motor_FaultBitmap);
+	}
 }
 
 /**
- * @brief   Sends MOTOR DRIVE command on CAN2
+ * @brief   Sends MOTOR DRIVE command on CAN3
  * @param   newVelocity desired motor velocity setpoint in m/s
- * @param   motorCurrent desired motor current setpoint as a percentage of max current setting
+ * @param   motorCurrent desired motor current setpoint as a percentage of max current setting (0.0-1.0)
```
ClarkPoon, line 111: Do you want any parameter checking in this function? What happens if you get passed an invalid parameter (and is this possible)?

Drivers/Src/MotorController.c
```diff
@@ -1,124 +1,217 @@
...
+	
+	// Check to see if a mailbox is available: BLOCKING
+	OSSemPend(
+        &MotorController_ReceiveSem4,
+	    0,    
+		OS_OPT_PEND_BLOCKING,
+		&ts,
+		&err);
+	assertOSError(0, err);
+    ErrorStatus status = BSP_CAN_Read(CAN_3, &id, data);
```
ClarkPoon, line 157: Is it possible for this function to error out and not release the mailbox semaphore?

Drivers/Src/MotorController.c
```diff
@@ -1,124 +1,217 @@
...
     return ERROR;
 }
+
+/**
+ * @brief   Mutex protected read from Velocity parameter
+ * @return  velocity value obtained from MotorController_Read
+ */ 
+float MotorController_ReadVelocity(void){
+    return CurrentVelocity;
+}
```
babusid, line 217: Not sure this is actually necessary, I think we'll end up doing a motor controller read every time and just using the values returned from that, but we can leave this in. Whoever writes updateVelocity may use this for convenience instead of manually unpacking the data.

babusid (dismissed):

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -0,0 +1,399 @@
...
+    gRxMessage[0].ExtId = 0x00;
+    gRxMessage[0].IDE = CAN_ID_STD;
+    gRxMessage[0].DLC = 0;
+    gRxMessage[0].FMI = 0;
+
+    /* Enable FIFO 0 message pending Interrupt */
+    CAN_ITConfig(CAN1, CAN_IT_FMP0, ENABLE);
+
+    // Enable Rx interrupts
+    NVIC_InitStruct.NVIC_IRQChannel = CAN1_RX0_IRQn; // TODO: CHECK IRQ CHANNELS
```
ClarkPoon, line 135: Do you still need to do this TODO?

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -0,0 +1,399 @@
...
+    RCC_APB1PeriphClockCmd(RCC_APB1Periph_CAN3, ENABLE);
+
+    /* CAN cell init */
+    CAN_InitStruct.CAN_TTCM = DISABLE;
+    CAN_InitStruct.CAN_ABOM = DISABLE;
+    CAN_InitStruct.CAN_AWUM = DISABLE;
+    CAN_InitStruct.CAN_NART = DISABLE;
+    CAN_InitStruct.CAN_RFLM = DISABLE;
+    CAN_InitStruct.CAN_TXFP = DISABLE;
+    CAN_InitStruct.CAN_Mode = CAN_Mode_Normal; // TODO: change back to Normal after testing
```
ClarkPoon, line 181: Can we get rid of this TODO comment?

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -0,0 +1,399 @@
...
+    gRxMessage[1].StdId = 0x00;
+    gRxMessage[1].ExtId = 0x00;
+    gRxMessage[1].IDE = CAN_ID_STD;
+    gRxMessage[1].DLC = 0;
+    gRxMessage[1].FMI = 0;
+
+    /* Enable FIFO 0 message pending Interrupt */
+    CAN_ITConfig(CAN3, CAN_IT_FMP0, ENABLE);
+
+    //TODO: Double check preemption priority and subpriority
```
ClarkPoon, line 223: Do we still need to do this?

Tests/Test_MotorController.c
```diff
@@ -1,28 +1,160 @@
...
+            OSTimeDlyHMSM(0,0,0,200,OS_OPT_TIME_HMSM_STRICT, &err);
+            if(err != OS_ERR_NONE){
+                volatile int x = 0;
+                while(1){
+                    x++;
+                }
+            }
+        }
+
+        desiredCurrent = 5.0f; //1 amps worth
```
ClarkPoon, line 51: Is this comment accurate? If so, having `5.0f` correspond to 1 amp is a bad interface

Drivers/Src/MotorController.c
```diff
@@ -1,124 +1,217 @@
...
+                "Motor Controller Receive Semaphore",
+                0,	// Number of mailboxes
+                &err);
+	assertOSError(0, err);
+
+    BSP_CAN_Init(CAN_3, MotorController_CountIncoming, MotorController_Release);
+
+    uint8_t data[8] = {0};
+    memcpy(
+        data+4,
```
ClarkPoon, line 90: Might be nice to have a comment or make this use `sizeof` instead of a magic 4

ClarkPoon (commented):
I have some small suggestions for this. If y'all want to go ahead and merge with no more fixes, let me know

babusid (approved):

ClarkPoon (approved):

PR #192, opened 2022-02-26
Cleaned up header files

manthanand:


Apps/Inc/PedalToPercent.h
```diff
@@ -1,6 +1,6 @@
 /* Copyright (c) 2021 UT Longhorn Racing Solar */
-
-
+#ifndef PEDALS_TO_PERCENT
+#define PEDALS_TO_PERCENT
```
ClarkPoon, line 3:
```suggestion
#ifndef PEDALS_TO_PERCENT_H
#define PEDALS_TO_PERCENT_H
```

Apps/Inc/Tasks.h
```diff
@@ -7,6 +7,11 @@
 #include "os.h"
 #include "common.h"
 #include "CarState.h"
+#include "Pedals.h"
+#include "MotorController.h"
+#include "CANbus.h"
+#include "Contactors.h"
+#include "Precharge.h"
```
ClarkPoon, line 14: What is this file using all of these for?

Apps/Src/ArrayConnection.c
```diff
@@ -1,7 +1,6 @@
 /* Copyright (c) 2020 UT Longhorn Racing Solar */
 
-#include "ArrayConnection.h"
-#include "Contactors.h"
+#include "Tasks.h"
```
ClarkPoon, line 3: I'm not a big fan of putting all of the `#include` statements in Tasks.h and making everything just include Tasks.h because it hides the dependencies for each file

PR #207, opened 2022-04-09
Adc overhaul

Cam0Cow:
The pedals are wired in such a way that they don't actually use the full range of 0-3.3 volts. In order to account for this in the pedal driver, we now map a different voltage range (per pedal) to 0-100%. The values selected right now seem to work OK, but should definitely be tuned further by looking at the ADC out for the pedals and the corresponding percentage.

### DIFF
--- a/Drivers/Src/Pedals.c
+++ b/Drivers/Src/Pedals.c
@@ -6,6 +6,19 @@
 
 #include "Pedals.h"
 
+// Constants used to tune the pedals
+// Indexed using pedal_t
+// Refine in testing
+static const int16_t LowerBound[NUMBER_OF_PEDALS] = {
+    1000, // Accelerator lower bound
+    2000, // Brake lower bound
+};
+
+static const int16_t UpperBound[NUMBER_OF_PEDALS] = {
+    ADC_RANGE_MILLIVOLTS, // Accelerator upper bound
+    ADC_RANGE_MILLIVOLTS, // Brake upper bound
+};
+
 /**
  * @brief   Initializes the brake and accelerator by using the 
  *          BSP_ADC_Init function with parameters ACCELERATOR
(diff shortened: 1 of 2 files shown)


Drivers/Src/Pedals.c
```diff
@@ -20,14 +33,22 @@ void Pedals_Init(){
...
  */
 int8_t Pedals_Read(pedal_t pedal){
+    if (pedal >= NUMBER_OF_PEDALS) return 0;
     int16_t millivoltsPedal = BSP_ADC_Get_Millivoltage(pedal);
-    int8_t percentage = millivoltsPedal * 100 / ADC_RANGE_MILLIVOLTS;
+
+    int8_t percentage = 0;
+    
+    if (millivoltsPedal >= LowerBound[pedal]) {
+        percentage = (millivoltsPedal - LowerBound[pedal]) * 100 /
```
ClarkPoon, line 48: If `millivoltsPedal` is a lot higher than `LowerBound[pedal]` (eg. `ADC_RANGE_MILLIVOLTS`), I think you will get overflow when you multiply by 100.

ClarkPoon (changes requested):

Drivers/Src/Pedals.c
Cam0Cow: millivoltsPedal is going to be a value between 0 and 3300, so multiplying it by 100 will easily still be within the range for an integer. The final result should always be in the range of 0-100, fitting within an int8_t

ClarkPoon: You are doing this with 16 bit signed integers though (max around 32,000), not 32 bit integers

Cam0Cow: I was under the impression that integers in C get promoted by default during these kinds of expressions (the literal 100 is an int, and so everything else becomes an int), but I can add some casts to make it more explicit.

Drivers/Src/Pedals.c
```diff
@@ -20,14 +33,22 @@ void Pedals_Init(){
...
-    int8_t percentage = millivoltsPedal * 100 / ADC_RANGE_MILLIVOLTS;
+    if (pedal >= NUMBER_OF_PEDALS) return 0;
+    int16_t millivoltsPedal = (int16_t) BSP_ADC_Get_Millivoltage(pedal);
+
+    int8_t percentage = 0;
+    
+    if (millivoltsPedal >= LowerBound[pedal]) {
+        percentage = (int8_t) (millivoltsPedal - LowerBound[pedal]) * 100 /
+         (UpperBound[pedal] - LowerBound[pedal]);
+    }
```
ClarkPoon, line 50:
```suggestion
    if (millivoltsPedal >= LowerBound[pedal]) {
        percentage = (int8_t) ( (int32_t) (millivoltsPedal - LowerBound[pedal]) * 100 /
         (UpperBound[pedal] - LowerBound[pedal]));
    }
```

ClarkPoon (approved):

babusid (approved):

PR #211, opened 2022-04-12
Contactors

ClarkPoon:
I modified the contactor driver so the contactors will keep track of their desired state and if they are enabled. Hopefully this will make it easier for the application layer to deal with race conditions between the charge enabled message from BPS and other tasks trying to set the state of the contactors.


babusid (approved):

babusid:
I'm merging this without additional reviews. It's small enough + if it breaks its my fault.

PR #214, opened 2022-04-16
make contactor state OFF if disabled and fixed typos

ClarkPoon:


babusid (approved):

Drivers/Inc/Contactors.h
```diff
@@ -49,7 +49,6 @@ void Contators_Enable(contactor_t contactor);
 
 /**
  * @brief   Disable the specified contactor
- *          Contactor will turn on if enabled and state is on
  * @return  None
```
tiandahuang, line 52: Maybe add a comment here stating that the contactor will also be turned off

tiandahuang (approved):

PR #221, opened 2022-04-24
Can fixes

ClarkPoon:


Tests/Test_ReadandSendCarCAN.c
```diff
@@ -0,0 +1,114 @@
...
+    OSTaskDel(&Task_EnableContactorsTCB, &err);
+}
+
+int main(void) {
+    OS_ERR err;
+    OSInit(&err);
+
+    // Initialized with error
+    if (err != OS_ERR_NONE) {
+        printf("OS error code %d\n", err);
```
manthanand, line 39: Will this printf work if BSP_UART_Init isn't called?

ClarkPoon: No it won't

Tests/Test_ReadandSendCarCAN.c
```diff
@@ -0,0 +1,120 @@
...
+    // delete this task
+    OSTaskDel(&Task_EnableContactorsTCB, &err);
+}
+
+int main(void) {
+    OS_ERR err;
+
+    // I don't know which UART y'all are using for the printf's I copy-pasted, but here you go
+    BSP_UART_Init(UART_2);
+    BSP_UART_Init(UART_3);
```
manthanand, line 39: UART_2 is for the USB (which should probably be changed later on).

manthanand (approved):

Drivers/Src/Contactors.c
```diff
@@ -76,10 +76,10 @@ State Contactors_Get(contactor_t contactor) {
...
         case ARRAY_CONTACTOR :
-            state = (BSP_GPIO_Get_State(CONTACTORS_PORT, ARRAY_CONTACTOR_PIN) == 1 ? OFF : ON);
+            state = BSP_GPIO_Get_State(CONTACTORS_PORT, ARRAY_CONTACTOR_PIN);
             break;
         case ARRAY_PRECHARGE :
-            state = (BSP_GPIO_Get_State(CONTACTORS_PORT, ARRAY_PRECHARGE_PIN) == 1? OFF : ON);
+            state = BSP_GPIO_Get_State(CONTACTORS_PORT, ARRAY_PRECHARGE_PIN);
             break;
         case MOTOR_CONTACTOR :
             state = BSP_GPIO_Get_State(CONTACTORS_PORT, MOTOR_CONTACTOR_PIN);
```
manthanand, line 85: Why don't we just return state in this function?

PR #222, opened 2022-04-30
Pedals that go over 100% don't return garbage data now

Cam0Cow:
Need to test this on the motor setup. This should eliminate the case where an unconnected brake pedal might allow the motor keep spinning.


babusid:
Added this logic to debugging_stuff branch

PR #225, opened 2022-04-30
Regenerative Braking Implemented

### LINKED ISSUE #224: Regenerative Braking
Implement regenerative braking in UpdateVelocity

IshDeshpa:
closes #224 
Use UpdateVelocity test file to test

### DIFF
--- a/Apps/Src/UpdateVelocity.c
+++ b/Apps/Src/UpdateVelocity.c
@@ -5,9 +5,13 @@
 #include "MotorController.h"
 #include "Contactors.h"
 #include <math.h>
-#define UNTOUCH_PEDALS_PERCENT 5
+
+// REGEN_CURRENT AS A PERCENTAGE (DECIMAL NUMBER) OF MAX CURRENT
 #define REGEN_CURRENT 0.5f
-//#define UNATTAINABLE_VELOCITY 1000.0f
+// THRESHOLD VELOCITY IN M/S
+#define THRESHOLD_VEL 1.0f
+// UNTOUCH_PEDALS_PERCENT AS A PERCENTAGE OF MAX PEDAL POSITION
+#define UNTOUCH_PEDALS_PERCENT 5
 
 extern const float pedalToPercent[];
 
@@ -34,6 +38,7 @@ void Task_UpdateVelocity(void *p_arg)
     
     float desiredVelocity = 0;
     float desiredMotorCurrent = 0;
+    
     while (1)
     {
         uint8_t accelPedalPercent = Pedals_Read(ACCELERATOR);
@@ -50,9 +55,13 @@ void Task_UpdateVelocity(void *p_arg)
         // Deadband comparison
         if(brakePedalPercent >= UNTOUCH_PEDALS_PERCENT){
             desiredVelocity = 0;
-            desiredMotorCurrent = 0;
-        } else {
-            if (Switches_Read(FOR_SW)) {
+            desiredMotorCurrent = Switches_Read(REGEN_SW) ? REGEN_CURRENT : 0;
+        } else if(Switches_Read(REGEN_SW) && MotorController_ReadVelocity() <= THRESHOLD_VEL){ 
+            desiredVelocity = 0;
+            desiredMotorCurrent = REGEN_CURRENT;
+        } 
+        else {
+            if(!Switches_Read(FOR_SW)){
                 desiredVelocity = MAX_VELOCITY;
             } else if (Switches_Read(REV_SW)) {
                 desiredVelocity = -MAX_VELOCITY;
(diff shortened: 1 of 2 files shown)


Apps/Src/UpdateVelocity.c
```diff
@@ -47,7 +52,12 @@ void Task_UpdateVelocity(void *p_arg)
...
             desiredMotorCurrent = 0;
-        } else {
+        }
+        else if(Switches_Read(REGEN_SW) && (MotorController_ReadVelocity() > THRESHOLD_VEL && RegenEnable)){ 
+            desiredVelocity = 0;
+            desiredMotorCurrent = REGEN_CURRENT;
+        } 
+        else {
             if(!Switches_Read(REV_SW)){
                 desiredVelocity = MAX_VELOCITY;
```
babusid, line 62: I would move this `MAX_VELOCITY` macro within this file rather than leaving it in whatever header it is in currently. It shouldn't be used anywhere else.

Apps/Src/UpdateVelocity.c
```diff
@@ -59,8 +69,8 @@ void Task_UpdateVelocity(void *p_arg)
 
         MotorController_Drive(velocity_to_rpm(desiredVelocity), desiredMotorCurrent);
 
-        // Delay of few milliseconds (10)
-        OSTimeDlyHMSM(0, 0, 0, 10, OS_OPT_TIME_HMSM_STRICT, &err);
+        // Delay of few milliseconds (100)
+        OSTimeDlyHMSM(0, 0, 0, 100, OS_OPT_TIME_HMSM_STRICT, &err);
```
babusid, line 73: This is close to 1/2 of the required frequency, which should be okay, but we may want to speed it up in the event that the motor controller drops the occasional CAN message from us. The low speed will help read telemetry, however.

Apps/Src/UpdateVelocity.c
```diff
@@ -4,9 +4,13 @@
...
 #include <math.h>
-#define UNTOUCH_PEDALS_PERCENT 5
+
+// REGEN_CURRENT AS A PERCENTAGE (DECIMAL NUMBER) OF MAX CURRENT
 #define REGEN_CURRENT 0.5f
-//#define UNATTAINABLE_VELOCITY 1000.0f
+// THRESHOLD VELOCITY IN M/S
+#define THRESHOLD_VEL 1.0f
+// UNTOUCH_PEDALS_PERCENT AS A PERCENTAGE OF MAX PEDAL POSITION
+#define UNTOUCH_PEDALS_PERCENT 5
```
babusid, line 13:
```suggestion
#define THRESHOLD_VEL 1.0f
// UNTOUCH_PEDALS_PERCENT AS A PERCENTAGE OF MAX PEDAL POSITION
#define UNTOUCH_PEDALS_PERCENT 5
#define TEST 4
```

babusid:
We're going to have to update the ReadCarCAN app in order to get the regen fully working, as that application is what signals regen enable or not. I don't think it currently fully supports it.

manthanand (approved):

Apps/Src/UpdateVelocity.c
```diff
@@ -50,14 +59,29 @@ void Task_UpdateVelocity(void *p_arg)
...
+            desiredMotorCurrent = Switches_Read(REGEN_SW) ? REGEN_CURRENT : 0;
+        } 
+        else if(Switches_Read(REGEN_SW) 
+                && (MotorController_ReadVelocity() > THRESHOLD_VEL) 
+                && RegenEnable
+                && accelPedalPercent < NEUTRAL_PEDALS_PERCENT){ 
+            
+            desiredVelocity = 0;
+            // set regen current based on how much the accel pedal is pressed down
+            desiredMotorCurrent = REGEN_CURRENT * REGEN_RANGE / (NEUTRAL_PEDALS_PERCENT - accelPedalPercent);
```
babusid, line 71: Won't this regen more when the accelerator is pushed closer to Neutral point? Ie. if accelPedalPercent is 24, desired motorCurrent will be more than if accelPedalPercent is 10. Wouldn't we want higher regen current if the accelerator is pushed less?

(just a quick thought on how to implement regen getting stronger as the pedal is pushed less)
```suggestion
            #define REGEN_CURVE_OFFSET = 25 // offset to translate the braking strength curve to the right, shouldn't exceed 25, or we might get a negative denominator. Should be used to adjust the steepness of the regen curve here.
            desiredMotorCurrent = REGEN_CURRENT * REGEN_RANGE / (NEUTRAL_PEDALS_PERCENT +  accelPedalPercent-REGEN_STRENGTH_OFFSET : 0);
```

babusid (changes requested):

Apps/Src/UpdateVelocity.c
ClarkPoon: Yes, that's correct. I'll fix this

babusid (approved):

PR #230, opened 2022-05-03
Tested and Debugged version of sysinit

babusid:


Apps/Src/ReadSwitches.c
```diff
@@ -91,4 +91,5 @@ static void UpdateLights() {
     if(rightblink==0){
         Lights_Set(RIGHT_BLINK,OFF);
     }
+    
 }
```
manthanand, line 95: No that important but should probably have newlines at the end of all library files.

manthanand (approved):

ClarkPoon (approved):

PR #233, opened 2022-06-04
Display supp voltage

### LINKED ISSUE #232: Move the BPS supplemental voltage to the controls display
Right now we have three displays planned for the cockpit. This is kind of a pain for ergonomics. The only thing that the BPS display provides us with that's critical is the supplemental battery voltage, so if we can move that info to the controls display, we can remove the BPS display and save space in the cockpit.

ClarkPoon:
I wrote some untested code for reading the supplemental voltage from CAN and I cleaned up a couple of helper functions/macros in the display driver. I don't feel like messing with the display driver right now, so maybe @Cam0Cow or @babusid can take a look at it.

Would fix #232


Cam0Cow (commented):
I can't test this right now but all of these changes look fine. I can also use this pull request to make the necessary changes for the display to work without the UART TX line.

chaseblock:
There are still some more quality-of-life changes to be done here for the display, but the functionality required is there.

PR #236, opened 2022-06-07
New Fault Handling

babusid:
New test file to test the fault handler, as well as updated fault behavior to better address faults.


IshDeshpa:
Add controls fault light

PR #239, opened 2022-06-11
Created Cruise Control

babusid:
Created cruise control functionality


Apps/Src/UpdateVelocity.c
```diff
@@ -48,23 +53,56 @@ void Task_UpdateVelocity(void *p_arg)
...
+        if((prevEnable == OFF) && (currEnable == ON)){ //we went from off to on, so we should toggle it
+            prevEnable = ON;
+            cruiseEnable ^= 1; 
+        } else if ((currEnable == OFF)){ //went from on to OFF, so reset the prev state
+            prevEnable = OFF;
+        }
+
+        //Edge detector to toggle the cruise set on only the rising edge
+        State currSet = Switches_Read(CRUZ_ST);
+        if((prevSet == OFF) && (currSet == OFF)){
```
tiandahuang, line 67: shouldn't the edge occur when ```prevSet == OFF``` and ```currSet == ON```

Apps/Src/UpdateVelocity.c
```diff
@@ -48,23 +53,56 @@ void Task_UpdateVelocity(void *p_arg)
...
+            prevSet = ON;
+            cruiseSet ^= 1;
```
manthanand, line 69: Should Cruise set be allowed to turn on if Cruise enable isn't on?

Apps/Src/UpdateVelocity.c
```diff
@@ -48,23 +53,56 @@ void Task_UpdateVelocity(void *p_arg)
...
+        } else if ((currSet == OFF)){
+            prevSet = OFF;
+        }
+
```
manthanand, line 73: I might be confused but if Cruise enable isn't on, can Cruise set be on? (since you set the speed after the enable and then it will follow that speed once set is enabled).

Apps/Src/UpdateVelocity.c
```diff
@@ -48,23 +53,56 @@ void Task_UpdateVelocity(void *p_arg)
...
             Lights_Set(BrakeLight, ON);
         }
         else{
             Lights_Set(BrakeLight, OFF);
         }
-    
+        
+
         // Deadband comparison
         if(brakePedalPercent >= UNTOUCH_PEDALS_PERCENT){
```
manthanand, line 92: Honestly, this is just from a mechnical perspective. It might be nice to make untouch pedals percent a bit bigger since it will be really hard to not press that pedal when sitting in the car. Just to put it out there.

Apps/Src/UpdateVelocity.c
```diff
@@ -48,23 +53,56 @@ void Task_UpdateVelocity(void *p_arg)
...
-            desiredVelocity = 0;
-            desiredMotorCurrent = (Switches_Read(REGEN_SW) && RegenEnable) ? REGEN_CURRENT : 0;
+            desiredVelocity = 0; //velocity setpoint becomes 0
```
manthanand, line 93: I'm not sure if this is how we are supposed to do it, but is it just a 1 or a zero if brake is pressed or not?

manthanand: If that is the case and it is still smooth, then we should consider making untouch pedals percent a bit larger

Apps/Src/UpdateVelocity.c
babusid: Yes. Fixed this.

Apps/Src/UpdateVelocity.c
babusid: No, you're right. Fixed.

Apps/Src/UpdateVelocity.c
babusid: Yeah, from a software perspective, brake is treated just like a button, so it's just pressed/unpressed. I am concerned that if we make untouch a bigger value, we will also reduce the one-pedal drive regen range. @ClarkPoon What do you think?

babusid: One idea is that we could boost both the untouch and the neutral pedal point for one pedal drive

JimothyGreene:
Logic seems good to me from an initial read. Main thing I would say is to just ensure that when the brake pedal is pressed, we don't continue trying to cruise. Is that case handled?

Apps/Src/UpdateVelocity.c
```diff
@@ -76,16 +119,23 @@ void Task_UpdateVelocity(void *p_arg)
         } 
         else {
             if(Switches_Read(FOR_SW)){
-                desiredVelocity = MAX_VELOCITY;
+                desiredVelocity = ((cruiseState==ON) ? cruiseSpeed : MAX_VELOCITY); //we're in cruise mode so use the recorded cruise velocity
```
ClarkPoon, line 122: This comment looks wrong

Apps/Src/UpdateVelocity.c
```diff
@@ -48,23 +53,61 @@ void Task_UpdateVelocity(void *p_arg)
...
+            State currSet = Switches_Read(CRUZ_ST);
+            if((prevSet == OFF) && (currSet == ON)){
+                prevSet = ON;
+                cruiseSet ^= 1;
+            } else if ((currSet == OFF)){
+                prevSet = OFF;
+            }
+        } else {
+            cruiseSet = OFF;
+        }
```
ClarkPoon, line 77: I think this could be more readable with something like
```C
// declare currEnabled before the start of the while loop
// declare these variables as bools

// edge detector
prevEnabled = currEnabled;
currEnabled = Switches_Read(CRUZ_EN);
if (!prevEnabled && currEnabled) {
    cruiseSet = !cruiseSet;

    if  (cruiseSet) {
        // turn on cruise control
    } else {
        // turn off cruise control
    }
}
```

Apps/Src/UpdateVelocity.c
```diff
@@ -17,9 +17,14 @@
 #define NEUTRAL_PEDALS_PERCENT 25
 #define REGEN_RANGE (NEUTRAL_PEDALS_PERCENT - UNTOUCH_PEDALS_PERCENT)
 
+static float cruiseSpeed = 0; //cruising speed
+static State cruiseSet = OFF; //whether cruise is set
+static State cruiseEnable = OFF; //whether cruise is enabled
+static State prevEnable = OFF; //extra state variable for edge detector
+static State prevSet = OFF; //extra state variable for edge detector
```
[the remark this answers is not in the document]
babusid: I've used it since our Switches Driver is built with the State return type, and I'd rather just use it here instead of casting it to `bool` for readability purposes.

Apps/Src/UpdateVelocity.c
```diff
@@ -28,7 +35,7 @@ static float convertPedaltoMotorPercent(uint8_t pedalPercentage)
     return pedalToPercent[pedalPercentage];
 }
 
-// Convert a float velocity to a desired RPM
+// Convert a float velocity in meters per second to a desired RPM
 static float velocity_to_rpm(float velocity) {
     float velocity_mpm = velocity * 60.0f; // velocity in meters per minute
```
IshDeshpa, line 40: Comment says meters per second, conversion is done on meters per minute?

Apps/Src/UpdateVelocity.c
```diff
@@ -48,44 +55,68 @@ void Task_UpdateVelocity(void *p_arg)
...
+                && accelPedalPercent < NEUTRAL_PEDALS_PERCENT //we're in the one-pedal drive range
+                && (cruiseEnable == OFF) //we're not in cruise mode
+            ){ 
             
             desiredVelocity = 0;
             // set regen current based on how much the accel pedal is pressed down
-            if (accelPedalPercent > UNTOUCH_PEDALS_PERCENT) {
-                desiredMotorCurrent = REGEN_CURRENT * (REGEN_RANGE - (accelPedalPercent - UNTOUCH_PEDALS_PERCENT)) / REGEN_RANGE;
+            if (accelPedalPercent > UNTOUCH_PEDALS_PERCENT_ACCEL) {
+                desiredMotorCurrent = REGEN_CURRENT * (REGEN_RANGE - (accelPedalPercent - UNTOUCH_PEDALS_PERCENT_ACCEL)) / REGEN_RANGE;
```
IshDeshpa, line 96: Is desired behavior for one pedal regen that the more you press the acceleration pedal, the less it applies brake force (the less it slows down)? Is it possible that this goes negative?

Apps/Src/UpdateVelocity.c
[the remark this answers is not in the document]
IshDeshpa: > I'm not a huge fan of the `enum State` to be honest. I feel like `bool` does the same thing.

That's what I said. It's stupid.

Apps/Src/UpdateVelocity.c
```diff
@@ -48,44 +55,68 @@ void Task_UpdateVelocity(void *p_arg)
...
             } else {
                 desiredMotorCurrent = REGEN_CURRENT;
             }
         } 
         else {
             if(Switches_Read(FOR_SW)){
-                desiredVelocity = MAX_VELOCITY;
+                desiredVelocity = (((cruiseEnable==ON)&&(accelPedalPercent < NEUTRAL_PEDALS_PERCENT)) ? cruiseSpeed : MAX_VELOCITY); //we're in cruise mode so use the recorded cruise velocity
```
IshDeshpa, line 103: Why are we comparing with neutral pedals percent? Why not untouched pedals percent?

Apps/Src/UpdateVelocity.c
IshDeshpa: Wait never mind I figured it out, checking if accelPedalPercent > untouch pedal percent will make sure it doesn't go negative. I still would like to know desired behavior though.

babusid: Desired behavior is that under the neutral point the less you press the pedal, the harder it brakes. The more you press, the less it breaks. So brake force increases as you lift your foot off the accelerator.

Apps/Src/UpdateVelocity.c
babusid: good catch I think. Let me look at it.

Apps/Src/UpdateVelocity.c
babusid: It takes in a meters per second value, multiplies by sixty to generate an MPM value, and then determines rpm by doing that value divided by the circumference.

JimothyGreene (approved):
If this has been tested and works according to those tests, then this LGTM. I can't see anything obviously flawed in the logic

PR #249, opened 2022-08-07
Multi-Bus CAN Driver

babusid:
- Made the CAN driver capable of handling more than one CAN Bus. 
- Also made some edits in the interest of scalability (ie. removing explicit switch case for message behavior).
- New Standard CANDATA_t data type should make using driver easier
- Compatibility layer for legacy MC driver. This will get deleted along with the rest of that driver once we rework the Motor Controller-related applications.
- Notice of deprecation for legacy MC driver written in header.


babusid:
I'd like reviews on this, as it does compile, but it shouldn't get merged in before we test it on the hardware.

manthanand:
Is this tested or gonna be a draft?

Drivers/Inc/CANbus.h
```diff
@@ -18,59 +20,57 @@ typedef enum {
...
 	CHARGE_ENABLE = 0x10C,
 	SUPPLEMENTAL_VOLTAGE = 0x10B,
-	STATE_OF_CHARGE = 0x106
+	STATE_OF_CHARGE = 0x106,
+	MOTOR_DRIVE = 0x221,
+	MOTOR_POWER = 0x222,
+	MOTOR_RESET = 0x223,
+	MOTOR_STATUS = 0x241,
+	MOTOR_VELOCITY = 0x243,
+	DO_NOT_USE_FORCE_ENUM_SIZE = 0xFFFFFFFF //force the enum values to be 32 bits wide. This value should not be used. This is here because the BSP layer expects a 32 bit ptr for ID's
```
tiandahuang, line 29: I think a better solution here would be to add ```-fno-short-enums``` to the Makefile and to use ```__attribute__((packed))``` whenever a short/char sized enum is desired

Drivers/Inc/CANbus.h
```diff
@@ -18,59 +20,57 @@ typedef enum {
...
 
 /**
- * @brief   Checks if the CAN ID matches with Motor disable ID
- * @param   canline (not implemented, default is CAN1) can line to read from
- * @param 	id CAN msg ID
- * @param 	buffer pointer to buffer in which to store the can msg
- * @param   blocking whether or not this Read should be a blocking read or a nonblocking read
- * @return  1 if ID matches and 0 if it doesn't
+ * @brief   Reads a CAN message from the CAN hardware and returns it to the provided pointers. 
+ * 	DOES NOT POPULATE IDXen or IDX. You have to manually inspect the first byte and the ID
```
tiandahuang, line 68: If you need to manually populate ```IDXen``` and ```IDX```, I'm not sure how useful it is to have those fields in the CAN datatype. It might be easier to have a lookup function/table for each ID which CAN reads and writes can reference for enabling the IDX byte.

Drivers/Inc/CANbus.h
```diff
@@ -18,59 +20,57 @@ typedef enum {
...
+ * Standard CAN packet
 */
 typedef struct {
-	CANPayload_t payload;
-	CANId_t id;
-}CANMSG_t;
+	CANId_t ID; 		//ID of message
+	uint8_t idx; 		//FOR TRANSMIT ONLY: if message is part of a sequence of messages (for messages longer than 64 bits), this indicates the index of the message. Recieve will not touch this
+	bool idxEn; 		//FOR TRANSMIT ONLY: whether to use idx or not. Recieve will not touch this.
+	uint8_t size; 		//size of this particular message IN BYTES. On writes, this should not Exceed 8 bytes for non-idx messages, and should not exceed 7 for idx messages.
```
tiandahuang, line 42: If ```size``` is always constant for each CAN id, having a ```size``` field would add some unnecessary complexity.

Drivers/Src/CANbus.c
```diff
@@ -98,71 +135,58 @@ ErrorStatus CANbus_Send(CANId_t id, CANPayload_t payload, CAN_blocking_t blockin
...
-    case ARRAY_CONTACTOR_STATE_CHANGE:
-        datalen = 1;
-        txdata[0] = (payload.data.b);
-        break;
-    default:
-        //This should never occur
-        return ERROR;
+    if(CanData.idxEn){ //first byte of txData should be the idx value
+        memcpy(txdata,&CanData.idx,sizeof(CanData.idx));
+        memcpy(&(txdata[1]),&CanData.data,CanData.size);
```
tiandahuang, line 140:
```suggestion
        memcpy(&(txdata[sizeof(CanData.idx)]),&CanData.data,CanData.size);
```

Drivers/Inc/CANbus.h
```diff
@@ -18,59 +20,57 @@ typedef enum {
...
+} CANDATA_t;
+
 
 typedef enum {CAN_BLOCKING=0, CAN_NON_BLOCKING} CAN_blocking_t;
 
 /**
- * @brief   Initializes the CAN system
- * @param   None
+ * @brief   Initializes the CAN system for a given bus
+ * @param   bus The bus to initialize. You can either use CAN_1, CAN_3, or the convenience macros CARCAN and MOTORCAN
```
tiandahuang, line 51: Is support for ```CAN_2``` going to be added? If not, there should probably be a note somewhere that ```CAN_2``` cannot be used.

Drivers/Src/CANbus.c
```diff
@@ -79,16 +115,17 @@ ErrorStatus CANbus_Send(CANId_t id, CANPayload_t payload, CAN_blocking_t blockin
...
+            &(CANMail_Sem4[bus]),
             0,
             OS_OPT_PEND_NON_BLOCKING,
             &timestamp,
             &err);
 
         // don't crash if we are just using this in non-blocking mode and don't block
-        if (err != OS_ERR_PEND_WOULD_BLOCK) {
-            assertOSError(OS_CANDRIVER_LOC,err);
+        if(err == OS_ERR_PEND_WOULD_BLOCK){
```
tiandahuang, line 125: Should the function return if in non-blocking mode and unable to pend on the semaphore?

Drivers/Inc/CANbus.h
```diff
@@ -18,59 +20,57 @@ typedef enum {
...
 /**
- * @brief   Checks if the CAN ID matches with Motor disable ID
- * @param   canline (not implemented, default is CAN1) can line to read from
- * @param 	id CAN msg ID
- * @param 	buffer pointer to buffer in which to store the can msg
- * @param   blocking whether or not this Read should be a blocking read or a nonblocking read
- * @return  1 if ID matches and 0 if it doesn't
+ * @brief   Reads a CAN message from the CAN hardware and returns it to the provided pointers. 
+ * 	DOES NOT POPULATE IDXen or IDX. You have to manually inspect the first byte and the ID
+ * @param   ID pointer to where to store the CAN id of the recieved msg
```
tiandahuang, line 69: These ```@param``` comments don't seem to match the actual function parameters

Drivers/Src/CANbus.c
```diff
@@ -3,73 +3,109 @@
...
  */
-void CANbus_Init(void)
+void CANbus_Init(CAN_t bus)
 {
     // initialize CAN mailbox semaphore to 3 for the 3 CAN mailboxes that we have
     // initialize tx
     OS_ERR err;
-
-    OSMutexCreate(&CANbus_TxMutex, "CAN TX Lock", &err);
+
```
tiandahuang, line 59: To match the error checking in other CAN functions, it might be nice to have an ```assert bus == CAN_1 or bus == CAN_2``` here

Drivers/Inc/CANbus.h
babusid: This makes sense, although you'd still need idx so that you could specify the position of the message in the array you're sending. I'll add a LUT that matches messages to idxEN values though.

Drivers/Inc/CANbus.h
babusid: They're not all the same, but this could be added to the LUT, since size will be a static property of each msg type.

Drivers/Inc/CANbus.h
```diff
@@ -4,73 +4,125 @@
...
+	{false, DOUBLE, 0x247},
+	{false, DOUBLE, 0x248},
+	{false, DOUBLE, 0x24E},
+	{false, BYTE  , 0x24F},
+	{false, DOUBLE, 0x221},
+	{false, DOUBLE, 0x222},
+	{false, DOUBLE, 0x223},
+	{false, DOUBLE, 0x241},
+	{false, DOUBLE, 0x243},
+};
```
ClarkPoon, line 80: This is what is causing your linker errors... Why not just declare this in CANbus.c?

babusid:
Note to self: 
- Change the enum back to the CAN_IDs, and use the sparse array structure for the LUT to index using the enum still. This will make receive faster because it will be an O(1) lookup into the table using the parsed hex code from the message.

Drivers/Src/CANbus.c
```diff
@@ -3,206 +3,241 @@
...
-        datalen = 1;
-        txdata[0] = (payload.data.b);
-        break;
-    default:
-        //This should never occur
-        return ERROR;
+
+    uint8_t txdata[8];
+    if(msginfo.idxEn){ //first byte of txData should be the idx value
+        memcpy(txdata,&CanData.idx,sizeof(CanData.idx));
```
tiandahuang, line 136: instead of ```sizeof(CanData.idx)``` just use 1. idx should be 1 byte anyways and this statement implies that ```sizeof(CanData.idx)``` can differ in length

Drivers/Src/CANbus.c
```diff
@@ -3,206 +3,241 @@
...
+    );
 
     OSMutexPost( // unlock the TX line
-        &CANbus_TxMutex,
+        &(CANbus_TxMutex[bus]),
         OS_OPT_POST_NONE,
         &err);
     assertOSError(OS_CANDRIVER_LOC,err);
+    if(retval == ERROR){
+        CANbus_TxHandler(bus); //release the mailbox by posting back to the counter semaphore
```
tiandahuang, line 164: ```CANbus_TxHandler()``` has a comment saying not to access directly

Drivers/Src/CANbus.c
```diff
@@ -3,206 +3,241 @@
...
     CPU_TS timestamp;
     OS_ERR err;
 
     if (blocking == CAN_BLOCKING)
     {
         OSSemPend( // check if the queue actually has anything
-            &CANBus_RecieveSem4,
+            &(CANBus_RecieveSem4[bus]),
             0,
             OS_OPT_PEND_BLOCKING,
```
tiandahuang, line 188: could be cool
```suggestion
            (blocking == CAN_BLOCKING) ? OS_OPT_PEND_BLOCKING : OS_OPT_PEND_NON_BLOCKING,
```

Drivers/Src/CANbus.c
```diff
@@ -172,37 +199,50 @@ ErrorStatus CANbus_Read(uint32_t *id, uint8_t *buffer, CAN_blocking_t blocking)
...
+        &(CANbus_RxMutex[bus]),
         OS_OPT_POST_1,
         &err);
     assertOSError(OS_CANDRIVER_LOC,err);
+    
+    //search LUT for idmatch to populate idx and trim data
+    CANId_t lookupid = (CANId_t) MsgContainer->data[0];
+    CANLUT_T entry = CANLUT[lookupid];
+    if(entry.idxEn==true){
+
```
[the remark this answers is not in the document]
tiandahuang: idx is actual idx byte value, idxEn is bool saying that message contains an idx

Drivers/Src/CANLUT.c
```diff
@@ -0,0 +1,35 @@
...
+#define DOUBLE 8
+#define NOIDX false
+#define IDX true
+
+
+/**
+ * @brief Lookup table to simplify user-defined packet structs. Contains fields that are always the same for every message of a given ID.
+ *        Indexed by CANId_t values. Any changes or additions must be made in parallel with changes made to the CANID_t enum in CANbus.h
+ */
+CANLUT_T CANLUT[NUM_CAN_IDS] = {
```
tiandahuang, line 16: const this

Drivers/Src/CANbus.c
```diff
@@ -3,206 +3,241 @@
...
         // don't crash if we are just using this in non-blocking mode and don't block
-        if (err != OS_ERR_PEND_WOULD_BLOCK) {
-            assertOSError(OS_CANDRIVER_LOC,err);
+        if(err == OS_ERR_PEND_WOULD_BLOCK){
+            return ERROR;
         }
     }
-
-
     if (err != OS_ERR_NONE)
```
tiandahuang, line 126: assert will terminate everything anyways

Drivers/Src/CANbus.c
```diff
@@ -3,206 +3,241 @@
...
+ * @param   blocking: Whether or not this should be a blocking call
+ * @return  ERROR if data wasn't sent, SUCCESS if it was sent.
  */
-ErrorStatus CANbus_Send(CANId_t id, CANPayload_t payload, CAN_blocking_t blocking)
+ErrorStatus CANbus_Send(CANDATA_t CanData,CAN_blocking_t blocking, CAN_t bus)
 {
     CPU_TS timestamp;
     OS_ERR err;
+
+    CANLUT_T msginfo = CANLUT[CanData.ID]; //lookup msg information in table
```
tiandahuang, line 91: could crash for invalid IDs
need to validate ID < num IDs before accessing

also, there needs to be some way to see if IDs are valid (check for can len zero value?)

tiandahuang (commented):
clean yo code

Drivers/Inc/CANbus.h
```diff
@@ -4,73 +4,99 @@
...
+ * @param idxEn Whether or not this message is part of a sequence of messages.
+ * @param size Size of message's data
+ * @param ID The actual CAN ID of this message
+ */
 typedef struct {
-	uint8_t idx : 8;
-	uint8_t bytes : 8;
-	CANData_t data;
-} CANPayload_t;
+	bool idxEn;
```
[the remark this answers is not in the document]
babusid: 👍

Drivers/Inc/CANbus.h
```diff
@@ -4,73 +4,99 @@
...
-	uint16_t h;
-	uint32_t w;
-	uint64_t d;
-} CANData_t;
-
+/**
+ * @brief Struct to use in CAN MSG LUT
+ * @param idxEn Whether or not this message is part of a sequence of messages.
+ * @param size Size of message's data
+ * @param ID The actual CAN ID of this message
```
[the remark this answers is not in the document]
babusid: 👍

Drivers/Inc/CANbus.h
```diff
@@ -4,73 +4,99 @@
...
+	MOTOR_RESET 					= 0x223,
+	MOTOR_STATUS 					= 0x241,
+	MOTOR_VELOCITY 					= 0x243,
+	MC_BUS 							= 0x242,
+	VELOCITY 						= 0x243,
+	MC_PHASE_CURRENT 				= 0x244,
+	VOLTAGE_VEC 					= 0x245,
+	CURRENT_VEC 					= 0x246,
+	BACKEMF 						= 0x247,
+	TEMPERATURE 					= 0x248,
```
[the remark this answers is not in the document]
babusid: 👍

Drivers/Src/CANbus.c
babusid: 👍

Drivers/Src/CANbus.c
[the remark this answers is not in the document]
babusid: I copied over the idx value from the first byte of the data buffer, and then shifted the data over. The MsgContainer idx should include the actual idx int value, not the boolean indicating whether or not its enabled.

Drivers/Src/CANLUT.c
babusid: 👍

Drivers/Src/MotorController.c
```diff
@@ -24,8 +20,6 @@
 
 #define BYTES_TO_UINT32(bytes) ((bytes[]))
```
IshDeshpa, line 21: This macro isn't being used anywhere

Cam0Cow (approved):
Works on hardware.

tiandahuang (approved):
lgtm! 🚀

Issue #242, opened 2022-06-27
New Display Driver


babusid:
New Display Driver is necessary. The old one sucks.

IshDeshpa:
Also redoing the display HMI on this one

PR #255, opened 2022-09-04
Display rewrite

IshDeshpa:
#242 
Make sure you take a look at the HMI!


Apps/Src/BlinkLights.c
```diff
@@ -13,9 +13,10 @@ void Task_BlinkLight(void* p_arg){
     while(1){
         Lights_Set(LEFT_BLINK, (State) (Lights_Toggle_Read(LEFT_BLINK) ^ Lights_Read(LEFT_BLINK)));
         Lights_Set(RIGHT_BLINK, (State) (Lights_Toggle_Read(RIGHT_BLINK) ^ Lights_Read(RIGHT_BLINK)));
-        Display_SetLight(LEFT_BLINK, (State) (Lights_Toggle_Read(LEFT_BLINK) ^ Lights_Read(LEFT_BLINK)));
-        Display_SetLight(RIGHT_BLINK, (State) (Lights_Toggle_Read(RIGHT_BLINK) ^ Lights_Read(RIGHT_BLINK)));
         
+        // Maybe add a check for if this has changed so we're not sending this command all the time?
```
Cam0Cow, line 17: A check here is probably good practice, since we don't actually have to change the light state if the blinkers aren't on.

Drivers/Inc/Display.h
```diff
@@ -10,39 +10,54 @@
...
+
+typedef struct{
+	char* compOrCmd;
+	char* attr;
+	char* op;
+	uint8_t numArgs;
+	bool argTypes[MAX_ARGS];	// TRUE for integers, FALSE for strings
+	union{
+		char* str;
+		uint8_t num;
```
Cam0Cow, line 31: Should probably be a 32 bit integer, since many of our values are bigger than 8 bits. The union is already 32 bits in size anyways, due to the char *.

Drivers/Src/Display.c
```diff
@@ -1,325 +1,280 @@
...
-    BSP_UART_Write(DISP_OUT, msg, strlen(msg));
-    BSP_UART_Write(DISP_OUT, (char *)TERMINATOR, strlen(TERMINATOR));
-
-    // Get a response from the display
-    char buf[8];
-    BSP_UART_Read(DISP_OUT, buf);
-    return (IsNextionFailure(buf[0])) ? ERROR : SUCCESS;
-}
+	OSMutexPend(&DisplayQ_Mutex, 0, OS_OPT_POST_NONE, &ticks, &err);
+  assertOSError(OS_DISPLAY_LOC, err);
```
Cam0Cow, line 66: Nice indentation.

Drivers/Src/Display.c
```diff
@@ -1,325 +1,280 @@
...
+	
+	bool result = disp_fifo_get(&msg_queue, &cmd);
+	OSMutexPost(&DisplayQ_Mutex, OS_OPT_POST_NONE, &err);
+	assertOSError(OS_SEND_CAN_LOC, err);
+	
+	if(result == ERROR)
+		return ERROR;
+	
+	char msg[32];
+	if((cmd.numArgs == 1 && cmd.args != NULL && cmd.argTypes != NULL) && cmd.op != NULL && cmd.attr != NULL){	// Assignment commands have only 1 arg, an operator, and an attribute
```
Cam0Cow, line 101: The extra set of parentheses isn't technically necessary.

Drivers/Src/Display.c
```diff
@@ -1,325 +1,280 @@
...
+		if(cmd.argTypes[0]){
+			sprintf(msg, "%s.%s%s%d", cmd.compOrCmd, cmd.attr, cmd.op, cmd.args[0].num);
+		}
+		else{
+			sprintf(msg, "%s.%s%s%s", cmd.compOrCmd, cmd.attr, cmd.op, cmd.args[0].str);
+		}
+	}
+	else if(cmd.op == NULL && cmd.attr == NULL){	// Operational commands have no attribute and no operator, just a command and >= 0 arguments
+		sprintf(msg, "%s", cmd.compOrCmd);
```
Cam0Cow, line 110: I'm worried about buffer overflow here, since we only have a 32 byte buffer for the string. We should either send the command chunk-by-chunk, or error out if commands get too long.

Drivers/Src/Display.c
```diff
@@ -1,325 +1,280 @@
...
-ErrorStatus setComponentVisibility(enum CommandString_t comp, bool vis)
-{
-    char out[24];
-    sprintf(out, "%s %s,%d%s", CommandStrings[VIS], CommandStrings[comp], vis ? 1 : 0, TERMINATOR);
+ErrorStatus Display_SetComponent(Component_t comp, uint8_t val){
+	// For components that are on/off
+	if(comp <= MOTOR && val <= 1){
+		// If blinkers, set the blink toggle of the components instead of the visibility
+		if(comp == LEFT || comp == RIGHT || comp == HZD){
+			Display_Cmd_t toggleCmd = {(char*)compStrings[comp],"en","=",1,{true},
```
Cam0Cow, line 150: Using a designated initializer might make this a bit clearer.

Drivers/Src/Display.c
```diff
@@ -1,325 +1,280 @@
...
-        if(on==ON)
-            return updateIntValue(PAGE, BCO, NEXTION_RED);
-        return updateIntValue(PAGE, BCO, NEXTION_DARK_GREY);
-    }
-        
-    setComponentVisibility(light+19, on==ON);
-    return SUCCESS;
+ErrorStatus Display_Fault(os_error_loc_t osErrCode, fault_bitmap_t faultCode){
+	char faultPage[7] = "page 2";
+	BSP_UART_Write(DISP_OUT, faultPage, strlen(faultPage));
```
Cam0Cow, line 254: This might fail if the display is in the middle of receiving a command. It might be a good idea to send a terminator on its own to end the (potential) previous command and put the display in a consistent state we can talk to.

Cam0Cow (changes requested):
Two more general comments:
1. The Display driver has a lot of functions that aren't found in Display.h. Those should probably be marked as static.
2. When setting the speed to be a two digit number, the m in mph is partially overwritten.

Apps/Src/SendDisplay.c
```diff
@@ -6,25 +6,14 @@
...
-        Display_SetGear(Switches_Read(FOR_SW), Switches_Read(REV_SW));
-        Display_SetLight(Headlight_ON, Switches_Read(HEADLIGHT_SW));
-
+        if(Display_SendNext() == ERROR){
+            FaultBitmap |= FAULT_DISPLAY;
+            
+            OSSemPost(&FaultState_Sem4, OS_OPT_POST_1, &err);
+        }
+        
         OSTimeDlyHMSM(0, 0, 0, 100, OS_OPT_TIME_HMSM_NON_STRICT, &err); // Update screen at roughly 10 fps
```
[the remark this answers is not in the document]
IshDeshpa: So no delay at all then?

Drivers/Inc/Display.h
IshDeshpa: I made sure that none of the numbers we were sending over are >255. The percentage we're displaying is an integer from 0-100 and the voltage is 0-12.

Drivers/Src/Display.c
IshDeshpa: Right, it was more for readability (to group the argument stuff together)

Apps/Src/SendDisplay.c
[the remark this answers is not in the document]
Cam0Cow: I don't think so, but that can wait until the task is fully implemented.

Drivers/Src/Display.c
IshDeshpa: Right, I did that for the union but you're saying for the struct as a whole? Sounds like a good idea to me

Drivers/Inc/Display.h
Cam0Cow: What about the velocity? I thought we were sending it with a granularity of 0.1 mph. Or is that also changed? Either way, we have a 32-bit space for an integer there, so we can have it as an int32_t for the future.

Drivers/Src/Display.c
IshDeshpa: Or just increase the buffer, right?

Drivers/Inc/Display.h
IshDeshpa: Ooh I forgot about the velocity. I'll change it to a 32 bit then.

Drivers/Src/Display.c
Cam0Cow: The task stack size is pretty small, so that's something to watch out for. That's part of the reason I elected to send the commands in pieces instead of assembling them fully in memory before sending.

Apps/Inc/SendDisplay.h
```diff
@@ -8,4 +8,95 @@
...
+ * @returns ErrorStatus: ERROR or SUCCESS
+ */
+ErrorStatus Display_SetRightBlink(bool state);
+
+/**
+ * @brief Sets the headlight state on the display
+ * @param state on or off
+ * @returns ErrorStatus: ERROR or SUCCESS
+ */
+ErrorStatus Display_SetHeadlight(bool state);
```
manthanand, line 100: I think it would be neat to also display power draw/power delivered. If it makes the display too cluttered then ig it's not important, but just a thought I had. Idk if array sends voltage/current data, but BPS sends voltage/current for power draw.

Drivers/Inc/Display.h
```diff
@@ -1,48 +1,80 @@
...
-ErrorStatus Display_NoErrors(void);
+/**
+ * Packages relevant display command data for use in a fifo
+ */
+typedef struct{
+	char* compOrCmd;
+	char* attr;
+	char* op;
+	uint8_t numArgs;
+	bool argTypes[MAX_ARGS];	// TRUE for integers, FALSE for strings
```
Cam0Cow, line 47: Might want to have an enum with INT_ARG and STRING_ARG, it'll make the code using this struct a bit clearer.

Drivers/Src/Display.c
```diff
@@ -1,325 +1,107 @@
...
-    if(light == 2){  // ctrl fault
-        if(on==ON)
-            return updateIntValue(PAGE, BCO, NEXTION_RED);
-        return updateIntValue(PAGE, BCO, NEXTION_DARK_GREY);
-    }
-        
-    setComponentVisibility(light+19, on==ON);
-    return SUCCESS;
-}
+ErrorStatus Display_Init(){
```
Cam0Cow, line 19: Sending a display reset command might be good here, since resetting the Leaderboard doesn't reset the display as well.

Apps/Src/SendDisplay.c
```diff
@@ -1,30 +1,296 @@
...
+};
+
+/**
+ * @brief Pops the next display message from the queue and passes
+ * it to the display driver. Pends on semaphore and mutex to ensure that:
+ *  1) queue has messages to send (signaled by semaphore)
+ *  2) queue is not currently being written to by a separate thread (mutex)
+ * @returns ErrorStatus: ERROR or SUCCESS
+ */
+static ErrorStatus Display_PopNext(){
```
Cam0Cow, line 84: Won't these be better in the display driver itself? SendDisplay can still use them if they're in the driver.

Cam0Cow: It doesn't particularly matter, but it feels weird to have them in the Apps layer.

Apps/Src/SendDisplay.c
```diff
@@ -1,30 +1,296 @@
...
+	return Display_SetComponent(HEAD, state);
+}
+
+/**
+ * @brief Loops through the display queue and sends all messages
+ */
+void Task_SendDisplay(void *p_arg) {
+    OS_ERR err;
+
+    disp_fifo_new();
```
Cam0Cow, line 283: This should definitely be in the driver layer.

Cam0Cow (changes requested):

Apps/Src/SendDisplay.c
IshDeshpa: It does feel kinda weird, but I was talking to @babusid and we came to the conclusion that Display Driver should pretty much just be core functionality of sending a display command and a few necessary commands (reset, fault, etc) and the Display App can do more complex stuff, since most of those functions are wrappers for the send command function anyways

Apps/Src/SendDisplay.c
IshDeshpa: The queue should stay in the app layer, no? I think that if the queue was in the driver layer it would basically require an application to actually work, which reduces portability. Whereas if the queue is in the app and the driver is only sending messages, then the driver can be ported to another completely separate system

Apps/Inc/SendDisplay.h
IshDeshpa: What do you mean by power draw/delivered?

Apps/Src/SendDisplay.c
Cam0Cow: The problem with that approach is that now other tasks have to call functions from SendDisplay, instead of calling stuff from the driver. If everyone else thinks it's fine then we can keep it, but I'm still not 100% convinced.

Apps/Src/SendDisplay.c
Cam0Cow: Nothing is stopping another task from calling a display function that attempts to put an item in the queue before it's initialized. Being in the driver initialization would guarantee that.

IshDeshpa: But the queue isn't in the driver layer anymore, it's in the app

Cam0Cow: Yes, but the functions the interface exposes call the set component function, which calls send next, which attempts to push stuff into the fifo. If the fifo isn't yet initialized (which might happen if Send Display hasn't started executing yet) then bad stuff happens.

IshDeshpa: Right, so if the driver doesn't have the fifo, but the fifo needs to be initialized before any apps, then should a SendDisplay_Init be created and run along with driver initializations?

IshDeshpa:
- Active first, then enabled
- Blue for active, green for enabled

Apps/Src/SendDisplay.c
Cam0Cow: If you really want to keep those calls in SendDisplay, then there should probably be a separate function to initialize the display fifo, similar to CAN_Queue_Init; then just put it in the Init task along with all of the other driver initialization calls.

IshDeshpa:
Error checking and moving disp_fifo_new to a SendDisplay_Init() function still needs to be done.

IshDeshpa:
Test files still need to be rewritten to accommodate for new error checking functionality. @NathanielDelgado if you could handle this that would be great, but otherwise I can do it

Apps/Inc/SendDisplay.h
IshDeshpa: Will probably put this in another issue ticket.

Apps/Src/UpdateDisplay.c
```diff
@@ -0,0 +1,327 @@
...
+
+    if(!result)
+			return UPDATEDISPLAY_ERR_FIFO_POP;
+		
+		assertDisplayError(Display_Send(cmd));
+		return UPDATEDISPLAY_ERR_NONE;
+}
+
+/**
+ * @brief Pops the next display message from the queue and passes
```
Cam0Cow, line 114: The first part of the header is wrong, this pushes a command to the fifo.

Apps/Src/UpdateDisplay.c
```diff
@@ -0,0 +1,332 @@
...
+	Display_Cmd_t refreshCmd = {
+		.compOrCmd = "click",
+		.attr = NULL,
+		.op = NULL,
+		.numArgs = 2,
+		.argTypes = {INT_ARG,INT_ARG},
+		{
+			{.num=0},
+			{.num=1}
+		}
```
babusid, line 163:
```suggestion
		.args = {
			{.num=0},
			{.num=1}
		}
```

Cam0Cow (approved):

babusid:
@IshDeshpa are we passing all test cases with the new HMI file? If not confirmed we should roll back and only merge the new HMI once its been tested.

Cam0Cow (dismissed):

Cam0Cow (dismissed):
Changes verified running on Leaderboard + Display

Apps/Src/ReadSwitches.c
```diff
@@ -43,11 +43,13 @@ void Task_ReadSwitches(void* p_arg) {
         if (Switches_Read(IGN_1) == ON || Switches_Read(IGN_2) == ON) {
             Contactors_Enable(ARRAY_CONTACTOR);
             Contactors_Enable(ARRAY_PRECHARGE);
+            Lights_Set(A_CNCTR,ON);
```
babusid, line 46: Don't need the lights set, display is going to be the sole indicator of this.

babusid: Then again, none of this file really matters since it's pretty much getting nuked later by the Minions rewrite anyways.

babusid (changes requested):
Looks mostly good, didn't dive too deep into the app/driver source code since we've exhaustively tested already. Mostly did a high-level scan. Just a few quirks here and there, but I think once those have been fixed, and I get some time to do more of a dive, we should be merged before next workday.

Apps/Src/UpdateDisplay.c
```diff
@@ -0,0 +1,292 @@
...
+*/
+#include "UpdateDisplay.h"
+#include "Minions.h"
+#include "MotorController.h"
+#include <math.h>
+
+/**
+ * Creates queue for display commands.
+ */
+#define DISP_Q_SIZE 5
```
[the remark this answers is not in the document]
Cam0Cow: It seems to be enough for now, but we can certainly increase it, either now or at some later point. It caused no trouble in testing so far.

Apps/Inc/UpdateDisplay.h
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * Error types
+ */
+typedef enum{
+	UPDATEDISPLAY_ERR_NONE			= 0,
+	UPDATEDISPLAY_ERR_FIFO_PUT		=-1,	// Error putting command in fifo
+	UPDATEDISPLAY_ERR_FIFO_POP		=-2,	// Error popping command from fifo
+	UPDATEDISPLAY_ERR_PARSE_COMP	=-3,	// Error parsing component/val in SetComponent
+} UpdateDisplayError_e;
```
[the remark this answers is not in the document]
Cam0Cow: I think this was from an experiment with clang-tidy? I'd like to see it changed back too.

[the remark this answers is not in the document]
IshDeshpa: This wasn't a clang-tidy thing, I asked Arnav to make sure the code was compliant with the current style guide. I talked with Manthan a little about this on workday but I think we may end up changing it to be _t for enums as well in the style guide. For right now though this is how the style guide wanted it so that's why I did that

Drivers/Src/Display.c
```diff
@@ -1,325 +1,141 @@
...
-    }
-
-    updateIntValue(PAGE, BCO, NEXTION_DARK_GREY);
-}
-
-/**
- * Set the displayed velocity to vel
- */
-ErrorStatus Display_SetVelocity(float vel)
+	char msgArgs[32];
```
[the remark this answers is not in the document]
babusid: Is 32 == the maximum command length for the display? Could be nice to do something like `#define MAX_DISP_CMD_LEN 32` at the top of the file.

Drivers/Src/Display.c
```diff
@@ -1,325 +1,141 @@
...
+	{					  // Operational commands have no attribute and no operator, just a command and >= 0 arguments
+		msgArgs[0] = ' '; // No args
+		msgArgs[1] = '\0';
+		if (cmd.numArgs > MAX_ARGS)
+			return DISPLAY_ERR_OTHER;
+		if (cmd.numArgs >= 1)
+		{ // If there are arguments
+			for (int i = 0; i < cmd.numArgs; i++)
+			{
+				char arg[16];
```
babusid, line 58: why is this 16 bytes? Magic numbers = bad, use macros or annotate the line with why its 16.

Drivers/Src/Display.c
```diff
@@ -1,325 +1,141 @@
...
+	}
+
+	if (cmd.numArgs >= 1)
+	{ // If there are arguments
+		BSP_UART_Write(DISP_OUT, msgArgs, strlen(msgArgs));
+	}
+
+	BSP_UART_Write(DISP_OUT, (char *)TERMINATOR, strlen(TERMINATOR));
+
+	char buf[8];
```
babusid, line 90: why are we writing 8 garbage bytes? If the idea is to transmit 8 bytes of 0s, maybe initialize it specifically ie. `char buf[8] = {0};`

Drivers/Inc/Display.h
```diff
@@ -1,48 +1,122 @@
...
+	DISPLAY_ERR_INV_INSTR =-2,	// Invalid instruction passed to nextion (0x00)
+	DISPLAY_ERR_INV_COMP  =-3,	// Invalid component id passed to nextion (0x02)
+	DISPLAY_ERR_INV_PGID  =-4,	// Invalid page id passed to nextion	(0x03)
+	DISPLAY_ERR_INV_VAR   =-5,	// Invalid variable name passed to nextion	(0x1A)
+	DISPLAY_ERR_INV_VAROP =-6,	// Invalid variable operation passed to nextion	(0x1B)
+	DISPLAY_ERR_ASSIGN    =-7,	// Assignment failure nextion	(0x1C)
+	DISPLAY_ERR_PARAMS    =-8,	// Invalid number of parameters passed to nextion	(0x1E)
+	DISPLAY_ERR_MAX_ARGS  =-9,  // Command arg list exceeded MAX_ARGS elements
+	DISPLAY_ERR_OTHER     =-10  // Other nextion display error
+} DisplayError_e;
```
babusid, line 38: weird naming convention again

Drivers/Inc/Display.h
```diff
@@ -1,48 +1,122 @@
...
-ErrorStatus Display_SetError(int idx, char *err);
-ErrorStatus Display_NoErrors(void);
+/**
+ * All three pages on the HMI 
+ */
+typedef enum{
+	STARTUP	=0,
+	INFO,
+	FAULT
+} Page_e;
```
babusid, line 52: ` }Page_t;`

Apps/Inc/UpdateDisplay.h
```diff
@@ -0,0 +1,112 @@
...
+ * @returns UpdateDisplayError_e
+ */
+UpdateDisplayError_e UpdateDisplay_SetPage(Page_e page);
+
+/**
+ * @brief Sets the state of charge value on the display
+ * @param percent charge as a percent
+ * @returns UpdateDisplayError_e
+ */
+UpdateDisplayError_e UpdateDisplay_SetSOC(uint8_t percent);
```
babusid, line 54: Indicate degree of precision in the header comment here. ie, if we have 0 digits of precision (100 = 100.0%), this should be indicated.

Apps/Src/UpdateDisplay.c
```diff
@@ -0,0 +1,335 @@
...
+		}
+	};
+
+	UpdateDisplayError_e ret = UpdateDisplay_PutNext(pgCmd);
+	return ret;
+}
+
+/* WRAPPERS */
+UpdateDisplayError_e UpdateDisplay_SetSOC(uint8_t percent){	// Integer percentage from 0-100
+	UpdateDisplayError_e ret = UpdateDisplay_SetComponent(SOC, percent);
```
babusid, line 254: maybe add a static variable so that you ensure you're not constantly setting the value over and over again when it hasn't changed.

babusid: Same applies for the rest of the functions that come after this.

babusid: Not a critical change though, just a suggestion.

babusid (commented):
No real critical changes necessary, mostly stylistic things, and one suggestion to decrease repetitive component setting. @IshDeshpa, if you can get these things implemented b4 saturday, we can merge this thing. Otherwise, we can still merge, but make a new (low-priority) issue ticket to deal with them later.

Drivers/Src/Display.c
IshDeshpa: This was something we did during testing, but it's not required so I'll fix it.

Apps/Src/FaultState.c
```diff
@@ -7,6 +7,7 @@
 #include "Contactors.h"
 #include "Minions.h"
 static bool fromThread = false; //whether fault was tripped from thread
+extern const PinInfo_t PINS_LOOKARR[]; // For GPIO writes
```
babusid, line 10:
```suggestion
extern const PinInfo_t PINS_LOOKARR[]; // For GPIO writes. Externed from Minions Driver C file.
```

Apps/Src/UpdateDisplay.c
```diff
@@ -0,0 +1,396 @@
...
+*/
+#include "UpdateDisplay.h"
+#include "Minions.h"
+#include "MotorController.h"
+#include <math.h>
+
+/**
+ * Creates queue for display commands.
+ */
+#define DISP_Q_SIZE 5
```
babusid, line 22:
```suggestion
#define DISP_Q_SIZE 10
```

babusid (approved):
lgtm

Cam0Cow (approved):

PR #258, opened 2022-09-17
Added build-essential to install script

Cam0Cow:
Some people didn't have gcc installed.


babusid (approved):

install.sh
```diff
@@ -17,6 +17,7 @@ sudo ln -s /usr/share/arm-gnu-toolchain-11.3.rel1-x86_64-arm-none-eabi/bin/arm-n
 sudo ln -s /usr/share/arm-gnu-toolchain-11.3.rel1-x86_64-arm-none-eabi/bin/arm-none-eabi-size /usr/bin/arm-none-eabi-size
 
 
+
```
IshDeshpa, line 20: Why is this extra line necessary?

IshDeshpa (changes requested):

ClarkPoon (approved):

install.sh
Cam0Cow: It doesn't compile otherwise.

IshDeshpa (approved):

PR #266, opened 2022-10-07
Minion driver rewrite

babusid:


Drivers/Inc/Minions.h
```diff
@@ -1,116 +1,48 @@
...
+} PinInfo_t;
+
+PinInfo_t PINS_LOOKARR[MINIONPIN_NUM] = {
+    {GPIO_Pin_4, PORTA, INPUT},
+    {GPIO_Pin_5, PORTA, INPUT},
+    {GPIO_Pin_6, PORTA, INPUT},
+    {GPIO_Pin_7, PORTA, INPUT},
+    {GPIO_Pin_4, PORTB, INPUT},
+    {GPIO_Pin_5, PORTB, OUTPUT}
+};
```
babusid, line 31: 1) Mark as const
2) Put it in the C file so that it actually compiles

Drivers/Src/Minions.c
```diff
@@ -1,406 +1,27 @@
...
-#define MEDIAN_FILTER_NAME SwitchesFilter
-#include "MedianFilter.h"
-static SwitchesFilter_t SwitchFilter;
+#include "os.h"
+#include "BSP_GPIO.h"
 
 static OS_MUTEX CommMutex; //Mutex to lock SPI lines
 // Stores switch states:
 // IGN_1 | IGN_2 | HZD_SW | REGEN_SW | RIGHT_SW | LEFT_SW | Headlight_SW | FOR_SW | REV_SW | CRUZ_EN | CRUZ_ST
 static uint16_t switchStatesBitmap = 0;
```
babusid, line 8: no longer need this

Drivers/Src/Minions.c
```diff
@@ -1,406 +1,27 @@
...
-/**
- * @brief   Read toggle bitmap
- * @return  returns uint8_t bitmap for toggle (LEFT/RIGHT)
- */
-uint8_t Lights_Toggle_Bitmap_Read(void) {
-    return lightToggleBitmap;
+bool Minion_Read_Switch(MinionPin_t pin){
+    if(pin == BRAKELIGHT){
+        return false; //trying to read from an output pin, can't do that.
+    }
```
babusid, line 20: This isn't generic behavior. This error check should lookup the passed-in pin in the lookup table, and use the returned PinInfo_t struct to determine if the pin is an output. This way, if we add more outputs in the future, we won't have to revisit this.

Drivers/Src/Minions.c
```diff
@@ -1,406 +1,27 @@
...
     }
 }
 
-/**
- * @brief   Read toggle bitmap
- * @return  returns uint8_t bitmap for toggle (LEFT/RIGHT)
- */
-uint8_t Lights_Toggle_Bitmap_Read(void) {
-    return lightToggleBitmap;
+bool Minion_Read_Switch(MinionPin_t pin){
```
babusid, line 17:
```suggestion
bool Minion_Read_Input(MinionPin_t pin){
```

Drivers/Inc/Minions.h
```diff
@@ -1,116 +1,48 @@
...
+    REV_SW,
+    CRUZ_EN,
+    CRUZ_ST,
+    //output
+    BRAKELIGHT, 
+    MINIONPIN_NUM,
+} MinionPin_t; 
+
+typedef struct PinInfo{
+    uint16_t pinmask;
```
babusid, line 19: Calling this a Pin Mask doesn't make a ton of sense, I'd call it a `pinNumber` instead

Drivers/Src/Minions.c
```diff
@@ -1,406 +1,27 @@
...
-
-        OSMutexPost(
-            &CommMutex,
-            OS_OPT_POST_NONE,
-            &err); // Unlock mutex
-        assertOSError(OS_BLINK_LIGHTS_LOC, err);
-    }
+void Minion_Write_Brakelight(State state){
+    BSP_GPIO_Write_Pin(PINS_LOOKARR[BRAKELIGHT].port, PINS_LOOKARR[BRAKELIGHT].pinmask, state);
 }
```
babusid, line 27: Same idea as my earlier comment. This isn't generic behavior. I would suggest having a generic `Minion_Write_Output` function that can take in any pin, and any state, and set it appropriately.

Drivers/Src/Minions.c
```diff
@@ -1,406 +1,51 @@
...
     }
-    
+    OSMutexCreate(&OutputMutex, "Minions Output Mutex", &err);
+    assertOSError(OS_MINIONS_LOC, err);
 }
 
+bool Minion_Read_Input(MinionPin_t pin){
+    if((PINS_LOOKARR[pin].direction == OUTPUT)){ //trying to read from an output pin, can't do that.
+        return false; 
+    }
```
IshDeshpa, line 34: If the direction is wrong, return an error. This requires the creation of an error enum (see [here](https://github.com/lhr-solar/Controls/blob/a4242873c8441336c8aa7f9360020b5b29eadb7c/Drivers/Inc/Display.h#L25-L37)) and an assertion function. Return the error through a pointer parameter and dereference the pointer.

The reasoning behind this is the end user needs to know when a write has failed (also makes it easier to debug)

Actually, we're able to read an output pin, so you may want to return an error here into the parameter (MINION_READ_OUTPUT_ERR or smth) but **_still_** read the input and return it. So returning the error will basically be a warning or a notifier to the user saying "hey you're reading an output but it's not gonna break the car and we're gonna return a valid value anyways". The assertion function can have a switch case that doesn't necessarily go to fault state based off of this.

Drivers/Src/Minions.c
```diff
@@ -1,406 +1,51 @@
...
-            &CommMutex,
-            OS_OPT_POST_NONE,
-            &err); // Unlock mutex
-        assertOSError(OS_BLINK_LIGHTS_LOC, err);
+    if(PINS_LOOKARR[pin].direction == OUTPUT){
+        OSMutexPend(&OutputMutex, 0, OS_OPT_PEND_BLOCKING, &timestamp, &err); 
+        BSP_GPIO_Write_Pin(PINS_LOOKARR[pin].port, PINS_LOOKARR[pin].pinMask, status);
+        OSMutexPost(&OutputMutex, OS_OPT_POST_NONE, &err);
+        assertOSError(OS_MINIONS_LOC, err);
+        return true;
```
IshDeshpa, line 48: Same thing here with the error checking.

BSP/Inc/BSP_GPIO.h
```diff
@@ -55,7 +56,7 @@ uint8_t BSP_GPIO_Read_Pin(port_t port, uint16_t pinmask);
  * @param   state ON or OFF
```
IshDeshpa, line 56: param is now a true or false value so the documentation can reflect that

IshDeshpa (dismissed):
Some things to address

BSP/Inc/BSP_GPIO.h
audreygpick: Nathaniel says you're wrong (and not to change anything in BSP)

BSP/STM32F413/Src/BSP_GPIO.c
```diff
@@ -88,7 +88,7 @@ uint8_t BSP_GPIO_Read_Pin(port_t port, uint16_t pinmask){
  * @return  None
```
babusid, line 88: documentation needs to reflect that the boolean true = ON and boolean false = OFF

babusid (commented):
lgtm, @millergage, just fix that and we can merge this with the application PR once that's also approved.

Drivers/Src/Minions.c
```diff
@@ -1,406 +1,51 @@
...
-        
-        lightStatesBitmap = lightNewStates;   // Update lights bitmap
 
-        OSMutexPost(
-            &CommMutex,
-            OS_OPT_POST_NONE,
-            &err); // Unlock mutex
-        assertOSError(OS_BLINK_LIGHTS_LOC, err);
+    if(PINS_LOOKARR[pin].direction == OUTPUT){
+        OSMutexPend(&OutputMutex, 0, OS_OPT_PEND_BLOCKING, &timestamp, &err);
```
Cam0Cow, line 44: This needs a corresponding assertOSError

Cam0Cow (changes requested):

Cam0Cow (dismissed):

NathanielDelgado (approved):

Cam0Cow (approved):

PR #267, opened 2022-10-21
Changed all code and commented out code that said 'State' to 'bool, '…

audreygpick:
changed...
- 'State' to 'bool'
- 'OFF' to 'false'
- 'ON' to 'true'
... for code and commented out code. Did not change for other comments or strings

### DIFF
--- a/Apps/Src/BlinkLights.c
+++ b/Apps/Src/BlinkLights.c
@@ -11,10 +11,10 @@
 void Task_BlinkLight(void* p_arg){
     OS_ERR err;
     while(1){
-        Lights_Set(LEFT_BLINK, (State) (Lights_Toggle_Read(LEFT_BLINK) ^ Lights_Read(LEFT_BLINK)));
-        Lights_Set(RIGHT_BLINK, (State) (Lights_Toggle_Read(RIGHT_BLINK) ^ Lights_Read(RIGHT_BLINK)));
-        Display_SetLight(LEFT_BLINK, (State) (Lights_Toggle_Read(LEFT_BLINK) ^ Lights_Read(LEFT_BLINK)));
-        Display_SetLight(RIGHT_BLINK, (State) (Lights_Toggle_Read(RIGHT_BLINK) ^ Lights_Read(RIGHT_BLINK)));
+        Lights_Set(LEFT_BLINK, (bool) (Lights_Toggle_Read(LEFT_BLINK) ^ Lights_Read(LEFT_BLINK)));
+        Lights_Set(RIGHT_BLINK, (bool) (Lights_Toggle_Read(RIGHT_BLINK) ^ Lights_Read(RIGHT_BLINK)));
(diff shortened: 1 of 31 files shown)


Apps/Src/BlinkLights.c
```diff
@@ -11,10 +11,10 @@
...
     OS_ERR err;
     while(1){
-        Lights_Set(LEFT_BLINK, (State) (Lights_Toggle_Read(LEFT_BLINK) ^ Lights_Read(LEFT_BLINK)));
-        Lights_Set(RIGHT_BLINK, (State) (Lights_Toggle_Read(RIGHT_BLINK) ^ Lights_Read(RIGHT_BLINK)));
-        Display_SetLight(LEFT_BLINK, (State) (Lights_Toggle_Read(LEFT_BLINK) ^ Lights_Read(LEFT_BLINK)));
-        Display_SetLight(RIGHT_BLINK, (State) (Lights_Toggle_Read(RIGHT_BLINK) ^ Lights_Read(RIGHT_BLINK)));
+        Lights_Set(LEFT_BLINK, (bool) (Lights_Toggle_Read(LEFT_BLINK) ^ Lights_Read(LEFT_BLINK)));
+        Lights_Set(RIGHT_BLINK, (bool) (Lights_Toggle_Read(RIGHT_BLINK) ^ Lights_Read(RIGHT_BLINK)));
+        Display_SetLight(LEFT_BLINK, (bool) (Lights_Toggle_Read(LEFT_BLINK) ^ Lights_Read(LEFT_BLINK)));
+        Display_SetLight(RIGHT_BLINK, (bool) (Lights_Toggle_Read(RIGHT_BLINK) ^ Lights_Read(RIGHT_BLINK)));
```
IshDeshpa, line 17: Unnecessary cast to bool (both Lights_Toggle_Read and Lights_Read return a bool)

BSP/Inc/BSP_GPIO.h
```diff
@@ -44,18 +44,18 @@ void BSP_GPIO_Write(port_t port, uint16_t data);
  * @brief   Reads data from a specified pin (not applicalbe to output pins)
  * @param   port The port to read from
  * @param   pinmask Mask from stm header file that says which pin to read from
- * @return  State of the pin
+ * @return  bool of the pin
```
IshDeshpa, line 47:
```suggestion
 * @return (bool) state of the pin
```

BSP/STM32F413/Src/BSP_GPIO.c
```diff
@@ -69,7 +69,7 @@ void BSP_GPIO_Write(port_t port, uint16_t data){
  * @brief   Reads data from a specified input pin (not applicable to output pins)
  * @param   port The port to read from
  * @param   pin The pin to read from 
- * @return  State of the pin
+ * @return  bool of the pin
```
IshDeshpa, line 72:
```suggestion
 * @return (bool) state of the pin

```

IshDeshpa (changes requested):
Some small changes, but really I think this works as a quick fix for now especially since we're removing State throughout new code as well (so a lot of these replacements will get overwritten anyways).

PR #268, opened 2022-10-22
SendCarCan Additional Info

NathanielDelgado:
Send pedal, switch, lights, and contactor readings on the CAN line for telemetry to use


Apps/Src/SendCarCAN.c
```diff
@@ -1,20 +1,57 @@
...
+    struct{
+        uint64_t accelPedal : 8;
+        uint64_t brakePedal : 8;
+        uint64_t switchBitmap : 16;
+        uint64_t lightBitmap : 8;
+        uint64_t contactorBitmap : 8;
+        uint64_t reserved : 16;
+    };
+    uint64_t val;
+}DATAMSG_t;
```
IshDeshpa, line 21:
```suggestion
typedef union CarData_t{
    struct{
        uint64_t accelPedal : 8;
        uint64_t brakePedal : 8;
        uint64_t switchBitmap : 16;
        uint64_t lightBitmap : 8;
        uint64_t contactorBitmap : 8;
        uint64_t reserved : 16;
    };
    uint64_t val;
}CarData_t;
```

Apps/Src/SendCarCAN.c
```diff
@@ -1,20 +1,57 @@
...
 
 void Task_SendCarCAN(void *p_arg){
-    CANMSG_t msg;
+    CANMSG_t motorMSG;
+    CANMSG_t carMSG;
```
IshDeshpa, line 25: Rename to motorMsg and carMsg

Apps/Src/SendCarCAN.c
```diff
@@ -1,20 +1,57 @@
...
+        CAN_Queue_Pend(&motorMSG);
+        CANbus_Send(motorMSG.id, motorMSG.payload, CAN_BLOCKING);
+
+        // Get pedal information
+        data.accelPedal = Pedals_Read(ACCELERATOR);
+        data.brakePedal = Pedals_Read(BRAKE);
+
+        // Get switch information
+        for(int i = 0; i < NUM_SWITCHES; i++){
+            bool switchState = (bool)Switches_Read(i);
```
IshDeshpa, line 39: Casting States to bools is technically unsafe. Use a ternary:
`Switches_Read(i)==ON?true:false`

Drivers/Inc/Minions.h
```diff
@@ -27,7 +27,8 @@ typedef enum {
      REGEN_SW,
      CRUZ_ST,
      IGN_1,
-     IGN_2
+     IGN_2,
+     NUM_SWITCHES,
```
IshDeshpa, line 31: I'm not a huge fan of this, I would rather use a macro for NUM_SWITCHES, but it's fine if nobody else has a problem with it.

Apps/Src/SendCarCAN.c
```diff
@@ -1,20 +1,57 @@
-#include "os.h"
 #include "Tasks.h"
-#include "bsp.h"
+#include "common.h"
```
IshDeshpa, line 2: File header documentation

Apps/Src/SendCarCAN.c
```diff
@@ -1,20 +1,57 @@
...
+        uint64_t lightBitmap : 8;
+        uint64_t contactorBitmap : 8;
+        uint64_t reserved : 16;
+    };
+    uint64_t val;
+}DATAMSG_t;
 
 void Task_SendCarCAN(void *p_arg){
-    CANMSG_t msg;
+    CANMSG_t motorMSG;
```
IshDeshpa, line 24: Method documentation

IshDeshpa (commented):
Very minor changes

NathanielDelgado:
I need to make sure the bitmaps are zero before updating them again (would hold old data).

babusid (changes requested):
Overall, looks solid to me. Just change the struct thing, so that we don't have an unnecessarily complex data structure. Pointers cast + dereference is cleaner in my opinion because it only exists where it needs to exist. You could even break it out into a helper "conversion" function if you think that's cleaner.

Cam0Cow (approved):

PR #271, opened 2022-10-29
ReadCarCAN Rewrite

KnockbackNemo:
- Changed precharge delay to use timer and callback function
- Changed CAN Watchdog Handler from a task to a timer and callback function
- Enters fault state if CAN Watch timer fires too many times (>2)
- Made all things in callback functions non-blocking
- Disabling charging due to missed BPS messages is now handled by fault state
- Added a non-blocking option for Contactors_Set


Apps/Src/ReadSwitches.c
```diff
@@ -51,7 +51,7 @@ void Task_ReadSwitches(void* p_arg) {
         }
         
         // motor on/off
-        Contactors_Set(MOTOR_CONTACTOR, Switches_Read(IGN_2));
+        Contactors_Set(MOTOR_CONTACTOR, Switches_Read(IGN_2), SET_CONTACTORS_NON_BLOCKING);
```
Cam0Cow, line 54: Shouldn't this be a blocking call? We don't want this to fail.

Apps/Src/ReadCarCAN.c
```diff
@@ -5,47 +5,49 @@
...
 
 // Array restart thread lock
 static OS_MUTEX arrayRestartMutex;
 static bool restartingArray = false;
 
-static void CANWatchdog_Handler(); //Handler if we stop getting messages
-static void ArrayRestart(); //handler to turn array back on
+/*
+static void CANWatchdog_Handler(); //Handler if we stop getting messages */
+static void arrayRestart(); //handler to turn array back on
```
Cam0Cow, line 34: You should probably put the arguments in this forward declaration. This is technically correct, but looks confusing.

Apps/Src/FaultState.c
```diff
@@ -79,8 +85,8 @@ void EnterFaultState(void) {
          * Temp error - Motor will throttle itself, and there's nothing we can do additional to cool it down
          */
     }
-    else if(FaultBitmap & FAULT_READBPS){ //This has been put in with future development in mind, it is not currently tripped by anything.
-        nonrecoverableFaultHandler();
+    else if(FaultBitmap & FAULT_READBPS){ // Missed BPS Charge enable message
+        killContactorHandler();
```
Cam0Cow, line 89: You probably want to return here, otherwise you get stuck in the infinite loop below.

Apps/Src/ReadCarCAN.c
```diff
@@ -101,65 +89,126 @@ static inline void chargingEnable(void) {
...
+
+    //increment trip counter
+    CANWatchTripCounter++;
+    
+    // Set fault bitmaps 
+    FaultBitmap |= FAULT_READBPS;
+    OSErrLocBitmap |= OS_READ_CAN_LOC;
+
+    // If we tripped the counter too many times, enter fault state
+    if(CANWatchTripCounter > CANWATCH_TRIP_LIMIT) {
```
Cam0Cow, line 136: Calling EnterFaultState and posting to the fault state semaphore will both do the same thing, the only difference is whether the fromThread variable is set or not. You probably want the trip counter to be in fault state and increment it there, so fault state can decide when it's recoverable and when it isn't.

Cam0Cow (changes requested):

Apps/Src/ReadSwitches.c
KnockbackNemo: Yes, you are so right! Thank you for noticing that.

Apps/Src/ReadCarCAN.c
```diff
@@ -67,98 +93,145 @@ static inline void chargingEnable(void) {
...
+
+    if(!regenEnable){    // If regen enable has been disabled during precharge, we don't want to turn on the main contactor immediately after
+        restartingArray = false;
+    }
+    else {
+       
+            Contactors_Set(ARRAY_CONTACTOR, ON, SET_CONTACTORS_NON_BLOCKING);
+    }
+
+        //Lights_Set(A_CNCTR, ON); Don't know where this function is. Does it even exist?
```
Cam0Cow, line 127: It no longer exists, you should probably replace it with a call to `UpdateDisplay_SetArray(true)`.

Apps/Src/ReadCarCAN.c
```diff
@@ -67,98 +93,145 @@ static inline void chargingEnable(void) {
     // if nothing is precharging and we need to turn the array on, turn it on
     shouldRestartArray = !restartingArray && (Contactors_Get(ARRAY_CONTACTOR)==OFF) && (Contactors_Get(ARRAY_PRECHARGE)==OFF);
 
-    // start precharge for array 
+    // wait for precharge for array 
     if(shouldRestartArray){
 
-         if (Contactors_Set(ARRAY_PRECHARGE, ON) == SUCCESS) {    // only run through precharge sequence if we successfully start precharge
-            restartingArray = true;
+            // Wait to make sure precharge is finished and then restart array
```
Cam0Cow, line 99: `restartingArray` is never actually set to `true` anywhere in the code. I think this is where it should be set, since this is the point where we're sure we're restarting the array.

Apps/Src/ReadCarCAN.c
```diff
@@ -67,98 +93,145 @@ static inline void chargingEnable(void) {
...
     assertOSError(OS_READ_CAN_LOC,err);
 }
 
+/**
+ * @brief Callback function for the precharge delay timer. Turns off precharge and restarts the array.
+ * @param p_tmr pointer to the timer that calls this function, passed by timer
+ * @param p_arg pointer to the argument passed by timer
+ * 
+*/
+static void arrayRestart(void *p_tmr, void *p_arg){
```
Cam0Cow, line 117: This function can probably be simplified. `restartingArray` is set to false no matter what, so it can be done at the end and not in the `if` statement. Setting the Contactors and updating the display can be put in the `if` block together, with no `else` necessary.

Apps/Src/ReadCarCAN.c
```diff
@@ -67,98 +93,145 @@ static inline void chargingEnable(void) {
...
 
     OSMutexCreate(&arrayRestartMutex, "array restart mutex", &err);
     assertOSError(OS_READ_CAN_LOC,err);
-    OSMutexCreate(&msg_rcv_mutex, "message received mutex", &err);
-    assertOSError(OS_READ_CAN_LOC,err);
+    OSSemCreate(&restartArraySem, "array restart semaphore", 0, &err);
+    assertOSError(OS_READ_CAN_LOC, err);
+    OSSemCreate(&canWatchDisableChargingSem, "disable charging semaphore", 0, &err);
+
+    // Calculate threshold value for the charge-enable saturation buffer based on buffer length
```
Cam0Cow, line 169: No need for a `for` loop, you can just do `saturationThreshold = (SAT_BUF_LENGTH + 1) * SAT_BUF_LENGTH / 4`

Cam0Cow: It might be better to initialize this value outside of the function since it's just a simple assignment.

Cam0Cow (changes requested):

Apps/Src/ReadCarCAN.c
```diff
@@ -52,114 +76,145 @@ static inline void chargingEnable(void) {
...
-                            &err);
-                assertOSError(OS_READ_CAN_LOC,err);
 
                 if((&dataBuf.data)[0] == 0){ // If the buffer doesn't contain 1 for enable, turn off RegenEnable and turn array off
                     chargingDisable();
-                } else {
-                    //We got a message of enable with a nonzero value, turn on Regen, If we are already in precharge / array is on, do nothing. 
-                    //If not initiate precharge and restart sequence. 
-                    chargingEnable();
+                    updateSaturation(-1); // Update saturation and buffer
```
Cam0Cow, line 207: `chargeMsgSaturation` isn't initialized anywhere, so it'll start off as zero. Is this intentional, or should it start off near the threshold?

Apps/Src/ReadCarCAN.c
```diff
@@ -52,114 +76,145 @@ static inline void chargingEnable(void) {
...
+                // Restart CAN Watchdog timer
+                OSTmrStart(&canWatchTimer, &err);
+                assertOSError(OS_READ_CAN_LOC, err);
 
-                OSMutexPost(&msg_rcv_mutex,
-                            OS_OPT_NONE,
-                            &err);
-                assertOSError(OS_READ_CAN_LOC,err);
 
                 if((&dataBuf.data)[0] == 0){ // If the buffer doesn't contain 1 for enable, turn off RegenEnable and turn array off
```
Cam0Cow, line 205: Since `dataBuf.data` is an array, you can simply access it as `dataBuf.data[0]`

Cam0Cow (changes requested):
Mostly small changes/suggestions, we can probably start testing on the Leaderboard soon.

Apps/Src/ReadCarCAN.c
```diff
@@ -5,45 +5,69 @@
...
+#include "os_cfg_app.h"
 
-static bool msg_recieved = false;
-static OS_MUTEX msg_rcv_mutex;
+// Saturation threshold is halfway between 0 and max saturation value (half of summation from one to the number of positions)
+#define saturationThreshold (SAT_BUF_LENGTH + 1) * SAT_BUF_LENGTH / 4 
 
+// timer delay constants
+#define CAN_WATCH_TMR_DLY_MS 500u
+#define CAN_WATCH_TMR_DLY_TMR_TS (CAN_WATCH_TMR_DLY_MS * OS_CFG_TMR_TASK_RATE_HZ) / (1000u) //1000 for ms -> s conversion
```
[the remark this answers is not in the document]
KnockbackNemo: Ah, I see how that would be an issue. Thanks!

Apps/Src/ReadCarCAN.c
KnockbackNemo: It wasn't intentional, but I think initializing it to zero works since the array used to calculate it will start off as all zeros as well, so chargeMsgSaturation will end up becoming zero as soon as we update it. 

I went back and explicitly initialized it to zero so that it's clearer, though

Apps/Src/ReadCarCAN.c
KnockbackNemo: Ok! 

Does the current way work as well, or would it give an incorrect value?

Cam0Cow: It works because arrays in C are weird, it's just more complicated than necessary.

Cam0Cow (approved):
Still need to test on hardware

Drivers/Src/Contactors.c
```diff
@@ -90,16 +90,24 @@ State Contactors_Get(contactor_t contactor) {
...
-    // acquire lock
-    OSMutexPend(&contactorsMutex, 0, OS_OPT_PEND_BLOCKING, &timestamp, &err);
-    assertOSError(OS_CONTACTOR_LOC, err);
+        // acquire lock if its available
+        OSMutexPend(&contactorsMutex, 0, blocking ? OS_OPT_PEND_BLOCKING : OS_OPT_PEND_NON_BLOCKING, &timestamp, &err);
+        
+        if(err == OS_ERR_PEND_WOULD_BLOCK){
+            return ERROR;
+        }
+        assertOSError(OS_CONTACTOR_LOC, err);
```
babusid, line 107: indentation looks a little odd

Apps/Inc/ReadCarCAN.h
```diff
@@ -9,4 +9,12 @@
 
 #include "CANbus.h"
 
+#define SAT_BUF_LENGTH 5
+
+/**
+ * @brief Returns whether regen braking is on or off
+ * @return  Whether regen braking is on or off
```
babusid, line 16: It's more whether regen is allowed since this thread doesn't explicitly control the actuation of the system.

Apps/Src/ReadCarCAN.c
```diff
@@ -5,45 +5,69 @@
 #include "Contactors.h"
 #include "Minions.h"
 #include "CAN_Queue.h"
+#include "FaultState.h"
+#include "os_cfg_app.h"
 
-static bool msg_recieved = false;
-static OS_MUTEX msg_rcv_mutex;
+// Saturation threshold is halfway between 0 and max saturation value (half of summation from one to the number of positions)
+#define saturationThreshold (((SAT_BUF_LENGTH + 1) * SAT_BUF_LENGTH) / 4)
```
babusid, line 12: Macros are usually all CAPS, to differentiate them from variables

Apps/Src/ReadCarCAN.c
```diff
@@ -5,45 +5,69 @@
...
+static OS_TMR canWatchTimer;
 
-//CAN watchdog thread variables
-static OS_TCB canWatchTCB;
-static CPU_STK canWatchSTK[DEFAULT_STACK_SIZE];
-
-//Array restart thread variables
-static OS_TCB arrayTCB;
-static CPU_STK arraySTK[DEFAULT_STACK_SIZE];
+// prechargeDlyTimer and delay constant
```
babusid, line 22:
```suggestion
// prechargeDlyTimer
```

Apps/Src/ReadCarCAN.c
babusid:
```suggestion
#define SATURATION_THRESHOLD (((SAT_BUF_LENGTH + 1) * SAT_BUF_LENGTH) / 4) 
```

Apps/Src/ReadCarCAN.c
```diff
@@ -52,114 +76,145 @@ static inline void chargingEnable(void) {
...
-                    //If not initiate precharge and restart sequence. 
-                    chargingEnable();
+                    updateSaturation(-1); // Update saturation and buffer
+
+                } else { //We got a message of enable with a nonzero value
+                    updateSaturation(1);
+
+                    // If the charge message saturation is above the threshold, wait for restart precharge sequence then enable charging.
+                    // If we are already precharging/array is on, nothing will be done 
+                    if (chargeMsgSaturation >= saturationThreshold){
```
babusid, line 214:
```suggestion
                    if (chargeMsgSaturation >= SATURATION_THRESHOLD){
```

babusid (dismissed):
Just have a few non-critical, stylistic issues, other than that, we are good to go!

Apps/Src/ReadCarCAN.c
```diff
@@ -5,45 +5,69 @@
 #include "Contactors.h"
 #include "Minions.h"
 #include "CAN_Queue.h"
+#include "FaultState.h"
+#include "os_cfg_app.h"
 
-static bool msg_recieved = false;
-static OS_MUTEX msg_rcv_mutex;
+// Saturation threshold is halfway between 0 and max saturation value (half of summation from one to the number of positions)
+#define saturationThreshold (SAT_BUF_LENGTH + 1) * SAT_BUF_LENGTH / 4
```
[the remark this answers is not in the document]
Cam0Cow: It would actually evaluate correctly as is, since `saturationThreshold` is never used in any mathematical expression. But it's certainly good practice to put it in parentheses.

Cam0Cow (approved):

PR #272, opened 2022-10-29
Minion Driver Application Rewrite

NathanielDelgado:
Rewriting the driver means code that called the old driver might not work. This pull request solves these issues, allowing the project to compile with the new driver.


IshDeshpa (changes requested):
A few other changes here.

Apps/Src/SendDisplay.c
```diff
@@ -1,7 +1,26 @@
...
+typedef enum {  
+    A_CNCTR = 0,
+    M_CNCTR,
+    CTRL_FAULT,
+    LEFT_BLINK,
+    RIGHT_BLINK,
+    Headlight_ON,
+    BrakeLight,
+    RSVD_LED
+} light_t;
```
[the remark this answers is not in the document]
audreygpick: This is for the code to compile. We are not going to rewrite other drivers.

Apps/Src/ReadSwitches.c
```diff
@@ -1,146 +0,0 @@
-/* Copyright (c) 2022 UT Longhorn Racing Solar */
```
babusid, line 1: Deleting this file is fine, but not without an alternative in place. While the new minion driver does make polling the switches completely unnecessary, we still need something to gate the enable line of the contactors. This logic *could* be moved into the initialization task that once it finishes up just hangs in a loop however.

babusid: If you don't want to do it that way however, then, this task will need to remain, simply to poll the ignition switch and enable/disable contactors. Up to whoever is implementing it.

Apps/Src/SendDisplay.c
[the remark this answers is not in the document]
IshDeshpa: If we removed the light and switches enum from the main minions driver, all other references to light_t and switches_t throughout the codebase need to be replaced with MinionPin_t references. Even if the code compiles, it will not function as intended because of the mismatch between enums. Plus, the reason this branch exists is to replace old driver code with new driver calls, which includes using MinionPin_t.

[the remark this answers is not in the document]
babusid: Alternatively, just comment out the references to light_t and switches_t enums, as any part of the stack that deals with it is going to get pruned regardless.

Apps/Inc/IgnitionContactor.h
```diff
@@ -0,0 +1,3 @@
+#include "os.h"
+#include "common.h"
+#include "Tasks.h"
```
babusid, line 3: Don't need an application-specific header for this. Just move these into the app itself.

Apps/Inc/Tasks.h
```diff
@@ -27,6 +27,7 @@
 #define TASK_READ_SWITCHES_PRIO             7
```
babusid, line 27: Delete the ReadSwitches defns in the tasks.h file

babusid: this includes all the macros that were defined for it, as well as pruning the data structures that were defined in tasks.c and externed in tasks.h

Apps/Src/FaultState.c
```diff
@@ -8,20 +8,20 @@
...
     //turn additional lights on to indicate critical error
-    BSP_GPIO_Write_Pin(LIGHTS_PORT, LEFT_BLINK_PIN, OFF);
-    Display_SetLight(LEFT_BLINK, ON);
-    BSP_GPIO_Write_Pin(LIGHTS_PORT, RIGHT_BLINK_PIN, OFF);
-    Display_SetLight(RIGHT_BLINK, ON);
-    BSP_GPIO_Write_Pin(LIGHTS_PORT, BRAKELIGHT_PIN, OFF);
-    Display_SetLight(CTRL_FAULT,ON); //turn on fault light
+    //Display_SetLight(LEFT_BLINK, ON);
+    //Display_SetLight(RIGHT_BLINK, ON);
+    Minion_Write_Output(BRAKELIGHT, false, &mErr);
```
babusid, line 23: should probably turn the brakelight on if we fault out, if possible. However, considering there's a mutex in the call stack, and we don't know if the OS is working properly, this should probably be a raw BSP_GPIO call to that pin.

Apps/Src/IgnitionContactor.c
```diff
@@ -0,0 +1,59 @@
...
+        } else {
+            Contactors_Disable(ARRAY_CONTACTOR);
+            Contactors_Disable(ARRAY_PRECHARGE);
+            //Lights_Set(A_CNCTR,OFF);
+            //Display_SetLight(A_CNCTR,OFF);
+        }
+        assertOSError(OS_MINIONS_LOC, err);
+
+        // motor on/off
+        Contactors_Set(MOTOR_CONTACTOR, Minion_Read_Input(IGN_2, &Merr));
```
babusid, line 44:
```suggestion
        if (Minion_Read_Input(IGN_1, &Merr)) {
            Contactors_Enable(ARRAY_CONTACTOR);
        } else {
           Contactors_Disable(ARRAY_CONTACTOR);
        }
        assertOSError(OS_MINIONS_LOC, err);
        
        if(Minion_Read_Input(IGN_2, &Merr)){
            Contactors_Enable(MOTOR_CONTACTOR);
            Contactors_Set(MOTOR_CONTACTOR, Minion_Read_Input(IGN_2, &Merr));
        } else {
            Contactors_Disable(MOTOR_CONTACTOR);
            Contactors_Set(MOTOR_CONTACTOR, Minion_Read_Input(IGN_2, &Merr));
        }
        assertOSError(OS_MINIONS_LOC, err);
```

Cam0Cow (approved):

Issue #259, opened 2022-09-17
UpdateVelocity Rewrite


NathanielDelgado:
Right now, the UpdateVelocity task is confusing and has some issues (cruise sort of doesn't work). We also want to maybe add some features

Necessities:
- Accelerate
- Brake

Additional Features:
- Cruise control
- Accelerate in cruise control and return to past cruise control speed
- One-pedal drive regen braking
- Brake-pedal regen braking
- Regen braking profiles (using a button to switch through different thresholds for regen braking)

PR #274, opened 2022-11-03
Updatevelocity rewrite

diyarajon:
renamed UpdateVelocity to SendTritium, added Toggles to Cruise and Regen buttons


Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,356 @@
...
+    velocitySetpoint = MotorController_ReadVelocity();
+    brakelightState = true;
+}
+
+void RecordVelocityDecider(){
+    if(cruiseEnable && !cruiseSet){
+        state = POWERED_CRUISE;
+    }else if(!cruiseEnable){
+        state = NORMAL_DRIVE;
+    }else if(brakePedalPercent > 0){
```
Cam0Cow, line 249: Shouldn't this be the first if? No matter what, we want to go to the brake state if the brake pedal is pressed.

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,356 @@
...
+    brakelightState = false;
+}
+
+void PoweredCruiseDecider(){
+    if(MotorController_ReadVelocity() > velocitySetpoint){
+        state = COASTING_CRUISE;
+    }else if(!cruiseEnable){
+        state = NORMAL_DRIVE;
+    }else if(!cruiseEnable && !cruiseSet){
+        state = RECORD_VELOCITY;
```
Cam0Cow, line 269: How would this work exactly? We're in cruise, disable then re enable cruise? It doesn't really hurt anything, but it seems to me like we'd probably go from cruise -> normal -> record velocity whenever the driver turns cruise control off and on again.

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,356 @@
...
+void BrakeHandler(){
+    currentSetpoint = 0;
+    velocitySetpoint = 0;
+    cruiseEnable = 0;
+    cruiseSet = 0;
+    brakelightState = true;
+}
+
+void BrakeDecider(){
+    if(brakePedalPercent == 0){
```
Cam0Cow, line 307: Might want to replace this with checking if the brake pedal is below a certain threshold: I'm not sure what kind of noise we're going to see on the pedal ADCs this time around. It would be super inconvenient to go into the brake state for no reason all of the time.

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,356 @@
...
+ * the NEUTRAL_THRESHOLD, the car will accelerate as normal.
+*/
+void OnePedalDriveHandler(){
+    if(accelPedalPercent < BRAKE_THRESHOLD){
+        // Motor brake
+        currentSetpoint = percentToFloat(((BRAKE_THRESHOLD - accelPedalPercent)*100)/BRAKE_THRESHOLD);
+        velocitySetpoint = 0;
+        brakelightState = true;
+    }
+    else if(accelPedalPercent < NEUTRAL_THRESHOLD  && accelPedalPercent < BRAKE_THRESHOLD){
```
Cam0Cow, line 325: I'm assuming the second part should be `accelPedalPercent > BRAKE_THRESHOLD`

Drivers/Inc/MotorController.h
```diff
@@ -50,7 +50,7 @@ void MotorController_Init(float busCurrentFractionalSetPoint);
 
 /**
  * @brief   Sends MOTOR DRIVE command on CAN3
- * @param   newVelocity desired motor velocity setpoint in rpm
+ * @param   newVelocity desired motor velocity setpoint in m/s
```
Cam0Cow, line 53: Did this actually change? The motor controller expects the new velocity in RPM, so the that's what the old driver took as an argument. I wouldn't change the documentation until the change is actually made.

Cam0Cow (changes requested):
The current switch de-bouncing code runs every iteration of the FSM (100 ms) which is about an order of magnitude longer than the expected settling time of our buttons and switches. So it probably wouldn't really do much. I'm not convinced we really need de-bouncing since we check the switches on a time interval; but if we do add it, the de-bouncing should happen more quickly (maybe with a timer callback). @babusid do you have a preferred approach here?

Apps/Src/SendTritium.c
IshDeshpa: It was meant to be cruiseEnable && cruiseSet, so that when set is pressed again during a cruise state we go to record velocity.

babusid:
> The current switch de-bouncing code runs every iteration of the FSM (100 ms) which is about an order of magnitude longer than the expected settling time of our buttons and switches. So it probably wouldn't really do much. I'm not convinced we really need de-bouncing since we check the switches on a time interval; but if we do add it, the de-bouncing should happen more quickly (maybe with a timer callback). @babusid do you have a preferred approach here?

@Cam0Cow What if we sped up the FSM, and had a periodic timer callback running solely to handle the motor CAN messages? The FSM would push motor messages to a 1-entry buffer that overwrites itself as messages come in, and the callback just grabs the present message at the given speed.

Tests/Test_SendTritium.c
```diff
@@ -0,0 +1,337 @@
...
+    // Accelerate Cruise to Coasting Cruise
+    goToCoastingCruise();
+    accelPedalPercent = 0;
+    stateBuffer();
+
+    // Accelerate Cruise to Brake States
+    goToOnePedalDrive();
+    brakePedalPercent = 15;
+    stateBuffer();
+
```
IshDeshpa, line 285: Might wanna print something here like "done with all state transitions" or smth

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,426 @@
...
+SCOPE uint8_t brakePedalPercent = 0;
+SCOPE uint8_t accelPedalPercent = 0;
+SCOPE bool reverseSwitch = false;
+SCOPE bool forwardSwitch = false;
+
+// Outputs
+SCOPE float currentSetpoint = 0;
+SCOPE float velocitySetpoint = 0;
+SCOPE float cruiseVelSetpoint = 0;
+static bool brakelightState = false;
```
IshDeshpa, line 57: Since the printing is now self contained in SendTritium, we need to reevaluate which of these inputs/outputs actually need to be non-static (if we don't use the variable in the test file, we don't need to expose the variable, so it can stay static all the time and we can remove the SCOPE macro and just replace it with static).

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,426 @@
...
+
+// Button states
+static bool regenButton = false;
+static bool regenPrevious = false;
+
+static bool cruiseEnableButton = false;
+static bool cruiseEnablePrevious = false;
+
+#else
+float velocityObserved = 0;
```
IshDeshpa, line 73: If this is being used or changed in the test file, it can have the SCOPE macro in front of it to conditionally expose/not expose the variable based on whether we're testing or not.

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,426 @@
...
+        printf("regenToggle: %d\n\r", regenToggle);
+        printf("brakePedalPercent: %d\n\r", brakePedalPercent);
+        printf("accelPedalPercent: %d\n\r", accelPedalPercent);
+        printf("reverseSwitch: %d\n\r", reverseSwitch);
+        printf("forwardSwitch: %d\n\r", forwardSwitch);
+        printf("currentSetpoint: %f\n\r", currentSetpoint);
+        printf("velocitySetpoint: %f\n\r", velocitySetpoint);
+        printf("velocityObserved: %f\n\r", velocityObserved);
+        printf("-------------------\n\r");
+        #endif
```
IshDeshpa, line 243: This should be in a helper function ( such as `dumpInfo()` ).

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,426 @@
...
+
+#ifndef __TEST_SENDTRITIUM
+    CANPayload_t velocityPayload;
+            velocityPayload.bytes = 32;
+            velocityPayload.data.d = velocitySetpoint;
+
+    CANPayload_t currentPayload;
+            currentPayload.bytes = 32;
+            currentPayload.data.d = currentSetpoint;
+#endif
```
IshDeshpa, line 204: Fix indentation here

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,426 @@
...
+#include "Pedals.h"
+#include "ReadCarCAN.h"
+#include "Minions.h"
+#include "SendTritium.h"
+#include "ReadTritium.h"
+#include "CANbus.h"
+
+
+// Macros
+#define MAX_VELOCITY 50.0f // m/s    if rmp: 20000.0f // rpm
```
IshDeshpa, line 26: We do need to figure out whether it's RPM or m/s, and ensure that we're sending and receiving the right data (+ converting units when we need to).
[Here's the link to the Controls Sharepoint folder.](https://utexas.sharepoint.com/:f:/s/ENGR-LonghornRacing/ElA5vebiiyFNltiZD4OtSaQByIxLEAP3tlJlGAjd56zD3A?e=nYh0py)
[Here's the document with CAN messages and their units.](https://utexas.sharepoint.com/:b:/s/ENGR-LonghornRacing/EVb2-iGdYkdGkEMl44-lTvIBgV-iotwTGzEtiI2kKI31mA?e=nswyrD)
Look for the Motor Drive Command (which we send the current/vel setpoint in) and the Velocity Measurement Command (which we get the velocity in). Velocity Measurement will end up being done in ReadCarCAN, so talk to Gage/Arnav (you may have to wait until they're done and then merge their branch back into your branch, but communicate with them). At the end of the day though, we do need to be sending/receiving correct units.

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,426 @@
...
+        FSM[state].stateHandler();    // do what the current state does
+        readInputs();   // read inputs from the system
+        FSM[state].stateDecider();    // decide what the next state is
+
+        // Sets brakelight
+        Minion_Write_Output(BRAKELIGHT, brakelightState, &minion_err);
+
+        // Drive
+        #ifndef __TEST_SENDTRITIUM
+         if(MOTOR_MSG_COUNTER_THRESHOLD%MotorCounter == 0){ // if smaller numbers, MOTOR_MSG_COUNTER_THRESHOLD-MotorCounter == 0
```
IshDeshpa, line 216: If you're resetting the Motor Counter every loop, can't you just check if `MotorCounter == MOTOR_MSG_COUNTER_THRESHOLD`?

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,426 @@
...
+
+#define MOTOR_MSG_PERIOD 100
+#define FSM_PERIOD 20
+#define MOTOR_MSG_COUNTER_THRESHOLD (MOTOR_MSG_PERIOD)/(FSM_PERIOD)
+
+#ifdef __TEST_SENDTRITIUM
+#define SCOPE
+#else
+#define SCOPE static
+static int MotorCounter = 0;
```
IshDeshpa, line 41: More descriptive name

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,426 @@
...
+        printf("forwardSwitch: %d\n\r", forwardSwitch);
+        printf("currentSetpoint: %f\n\r", currentSetpoint);
+        printf("velocitySetpoint: %f\n\r", velocitySetpoint);
+        printf("velocityObserved: %f\n\r", velocityObserved);
+        printf("-------------------\n\r");
+        #endif
+
+
+        // Delay of few milliseconds (100)
+        OSTimeDlyHMSM(0, 0, 0, 20, OS_OPT_TIME_HMSM_STRICT, &err);
```
IshDeshpa, line 247: The value of delay here should be using the macro MOTOR_MSG_PERIOD.

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,426 @@
...
+void OnePedalDriveDecider(){
+    if(!regenToggle || !ChargeEnable || cruiseEnable){
+        regenToggle = 0;
+        state = NORMAL_DRIVE;
+    }else if(brakePedalPercent >= NORMAL_BRAKE_THRESHOLD){
+        state = BRAKE_STATE;
+    }
+}
+
+
```
IshDeshpa, line 407: Add a comment here to describe the accelerate cruise state (similar to what the other states have).

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,426 @@
...
+    }else if(!reverseSwitch && forwardSwitch){
+        velocitySetpoint = MAX_VELOCITY;
+        currentSetpoint = percentToFloat(accelPedalPercent);
+    }else if(!reverseSwitch && !forwardSwitch){
+        currentSetpoint = 0;
+    }
+}
+
+void AccelerateCruiseDecider(){
+    if(accelPedalPercent == 0){
```
IshDeshpa, line 421: We may need to define a threshold macro for accel pedal as well, I'm worried that it might float a little bit and be at like 0.1 or smth.

Apps/Src/Tasks.c
```diff
@@ -52,6 +52,8 @@ State RegenEnable = OFF;
 fault_bitmap_t FaultBitmap = FAULT_NONE;
 os_error_loc_t OSErrLocBitmap = OS_NONE_LOC;
 
+bool UpdateVel_ToggleCruise;
```
IshDeshpa, line 55: This should be removed, no?

Apps/Src/SendTritium.c
IshDeshpa: Maybe even with a switch case to print out the name of the state depending on the value of `state`? that might be overkill but otherwise we will need to read like "State: 9" and look up the 9th state in the enum while testing.

IshDeshpa (changes requested):
Left a bunch of stuff to fix. Haven't looked at testing or specific state conditions yet, will do that soon. Also, we will need to include calls to the display application, specifically when we enter/exit any cruise state or regen state (as there are indicators for that).

Apps/Inc/ReadCarCAN.h
```diff
@@ -8,5 +8,7 @@
 #include "Tasks.h"
 
 #include "CANbus.h"
+extern bool ChargeEnable;
```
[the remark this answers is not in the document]
babusid: Should be using RegenEnable_Get() wherever this value is being read. Once you refactor your code to get rid of the dependency on this variable, make sure to delete this extern. DO NOT mess with ReadCarCAN as part of that refactoring effort.

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,442 @@
...
+#define FSM_PERIOD 20
+#define MOTOR_MSG_COUNTER_THRESHOLD (MOTOR_MSG_PERIOD)/(FSM_PERIOD)
+
+// Inputs
+static bool cruiseEnable = false;
+static bool cruiseSet = false;
+static bool regenToggle = false;
+static uint8_t brakePedalPercent = 0;
+static uint8_t accelPedalPercent = 0;
+
```
babusid, line 43: SUGGESTION: adding a local variable to take the place of the ChargeEnable you're currently using from ReadCarCAN
```C
#ifdef __TEST_SENDTRITIUM
bool regen_enable
#elseif 
static bool regen_enable
#endif
```

babusid: Read my other comments to get context for this suggestion

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,442 @@
...
+    // DEBOUNCING
+    if(regenCounter == DEBOUNCE_PERIOD){regenButton = true;}
+    else if(regenCounter == 0){regenButton = false;}
+
+    if(cruiseEnableCounter == DEBOUNCE_PERIOD){cruiseEnableButton = true;}
+    else if(cruiseEnableCounter == 0){cruiseEnableButton = false;}
+
+    if(cruiseSetCounter == DEBOUNCE_PERIOD){cruiseSet = true;}
+    else if(cruiseSetCounter == 0){cruiseSet = false;}
+
```
babusid, line 191: SUGGESTION: reading in the regen enable value, if we are not testing
```C
#ifndef __TEST_SENDTRITIUM
regen_enable = RegenEnable_Get();
#endif
```

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,442 @@
...
+    // TOGGLE
+    if(ChargeEnable && regenButton != regenPrevious && regenPrevious){regenToggle = !regenToggle;}
```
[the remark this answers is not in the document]
babusid: If you choose to follow my suggestions, you can replace the references to `ChargeEnable` with references to `regen_enable`

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,442 @@
...
+    if(forwardSwitch){
+        cruiseSet = Minion_Read_Input(CRUZ_ST, &err);
+    }
+    else{
+        cruiseEnable = false;
+        cruiseSet = false;
+        regenToggle = false;
+    }
+
+    if(forwardGear ^ reverseGear ^ neutralGear == false)
```
babusid, line 179: I think this is trying to say that if more than 1 is set or if none are set, we have a fault. However, I don't think this will work because it will sequentially evaluate them, and it's not a true 3-way XOR. I think a more failsafe way to do it is the following:
```C
uint8_t gearFault = (uint8_t) forwardGear + (uint8_t) reverseGear + (uint8_t) neutralGear;

if(gearFault!=1){
//whatever fault routine you're going to build
}

```

babusid: Also, just as a note, if a gear cannot be determined, neutral gear is the safest.

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,442 @@
...
+
+    #ifndef __TEST_SENDTRITIUM
+    CANPayload_t velocityPayload;
+    velocityPayload.bytes = 32;
+    velocityPayload.data.d = velocitySetpoint;
+
+    CANPayload_t currentPayload;
+    currentPayload.bytes = 32;
+    currentPayload.data.d = currentSetpoint;
+    #endif
```
babusid, line 232: This is using the old can driver. I've merged in the new one, but this code needs to be refactored to utilize it. Won't compile otherwise.

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,442 @@
...
+        // Sets brakelight
+        Minion_Write_Output(BRAKELIGHT, brakelightState, &minion_err);
+
+        // Drive
+        #ifndef __TEST_SENDTRITIUM
+        if(MOTOR_MSG_COUNTER_THRESHOLD == motorMsgCounter){
+            motorMsgCounter = 0;
+
+            CANbus_Send(VELOCITY, velocityPayload, CAN_NON_BLOCKING);
+            CANbus_Send(CURRENT_VEC, currentPayload, CAN_NON_BLOCKING);
```
babusid, line 248: Old canbus driver usage here as well.

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,442 @@
...
+static bool reverseGear = false;
+
+// Outputs
+static float currentSetpoint = 0;
+static float velocitySetpoint = 0;
+static float cruiseVelSetpoint = 0;
+static bool brakelightState = false;
+
+// Current observed velocity
+static float velocityObserved = 0;
```
babusid, line 55: Do you use this anywhere? If it's only ever used in a testing context, it should probably be wrapped with the same macro guard as everything else. It could also be defined somewhere else and accessed, but that may be unnecessary overhead.

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,442 @@
...
+        currentSetpoint = percentToFloat(accelPedalPercent);
+    }else if(neutralGear){
+        velocitySetpoint = MAX_VELOCITY;
+        currentSetpoint = 0;
+    }
+}
+
+void NormalDriveDecider(){
+    if(brakePedalPercent >= NORMAL_BRAKE_THRESHOLD){
+        state = BRAKE_STATE;
```
babusid, line 288: I'm guessing it's supposed to do this
```suggestion
        state = FSM[BRAKE_STATE];
```

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,442 @@
...
+    }else if(cruiseSet && cruiseEnable && velocitySetpoint > VELOCITY_THRESHOLD){
+        state = RECORD_VELOCITY;
```
babusid, line 290:
```suggestion
        state = FSM[RECORD_VELOCITY];
```

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,442 @@
...
+    }else if(!cruiseEnable && regenToggle && ChargeEnable){
+        state = ONEPEDAL_DRIVE;
```
babusid, line 292:
```suggestion
        state = FSM[ONEPEDAL_DRIVE];
```

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,442 @@
...
+    velocitySetpoint = MotorVelocity_Get();
+    #else
+    velocitySetpoint = velocityObserved;
+    #endif
+
+}
+
+void RecordVelocityDecider(){
+    if(brakePedalPercent >= NORMAL_BRAKE_THRESHOLD){
+        state = BRAKE_STATE;
```
babusid, line 311:
```suggestion
        state = FSM[BRAKE_STATE];
```

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,442 @@
...
+    }else if(cruiseEnable && !cruiseSet){
+        state = POWERED_CRUISE;
```
babusid, line 313:
```suggestion
        state = FSM[POWERED_CRUISE];
```

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,442 @@
...
+    }else if(!cruiseEnable || !forwardGear){
+        state = NORMAL_DRIVE;
```
babusid, line 315:
```suggestion
        state = FSM[NORMAL_DRIVE];
```

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,442 @@
...
+    else if(!cruiseEnable){
+        state = NORMAL_DRIVE;
+    }else if(cruiseEnable && cruiseSet && velocitySetpoint > VELOCITY_THRESHOLD){
+        state = RECORD_VELOCITY;
+    }else if(brakePedalPercent >= NORMAL_BRAKE_THRESHOLD){
+        state = BRAKE_STATE;
+    }else if(regenToggle){
+        cruiseEnable = false;
+        state = NORMAL_DRIVE;
+    }
```
babusid, line 347:
```suggestion
        state = FSM[COASTING_CRUISE];
    }
    #else
    if(velocityObserved > velocitySetpoint){
        state = FSM[COASTING_CRUISE];
    }
    #endif
    else if(!cruiseEnable){
        state = FSM[NORMAL_DRIVE];
    }else if(cruiseEnable && cruiseSet && velocitySetpoint > VELOCITY_THRESHOLD){
        state = FSM[RECORD_VELOCITY];
    }else if(brakePedalPercent >= NORMAL_BRAKE_THRESHOLD){
        state = FSM[BRAKE_STATE];
    }else if(regenToggle){
        cruiseEnable = false;
        state = FSM[NORMAL_DRIVE];
    }
```

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,442 @@
...
+        state = POWERED_CRUISE;
+    }else if(!cruiseEnable){
+        state = NORMAL_DRIVE;
+    }else if(brakePedalPercent >= NORMAL_BRAKE_THRESHOLD){
+        state = BRAKE_STATE;
+    }else if(regenToggle){
+        cruiseEnable = false;
+        state = NORMAL_DRIVE;
+    }
+}
```
babusid, line 370:
```suggestion
void CoastingCruiseDecider(){
    if(MotorVelocity_Get() <= velocitySetpoint){
        state = FSM[POWERED_CRUISE];
    }else if(!cruiseEnable){
        state = FSM[NORMAL_DRIVE;]
    }else if(brakePedalPercent >= NORMAL_BRAKE_THRESHOLD){
        state = FSM[BRAKE_STATE];
    }else if(regenToggle){
        cruiseEnable = false;
        state = FSM[NORMAL_DRIVE];
    }
}
```

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,442 @@
...
+    currentSetpoint = 0;
+    velocitySetpoint = 0;
+    cruiseEnable = 0;
+    cruiseSet = 0;
+    brakelightState = true;
+}
+
+void BrakeDecider(){
+    if(brakePedalPercent < NORMAL_BRAKE_THRESHOLD){
+        state = NORMAL_DRIVE;
```
babusid, line 385:
```suggestion
        state = FSM[NORMAL_DRIVE];
```

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,442 @@
...
+        currentSetpoint = percentToFloat(((accelPedalPercent - NEUTRAL_THRESHOLD)*100)/(100-NEUTRAL_THRESHOLD));
+        velocitySetpoint = MAX_VELOCITY
+        brakelightState = false;
+    }
+}
+
+void OnePedalDriveDecider(){
+    if(!regenToggle || !ChargeEnable || cruiseEnable){
+        regenToggle = 0;
+        state = NORMAL_DRIVE;
```
babusid, line 417:
```suggestion
        state = FSM[NORMAL_DRIVE];
```

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,442 @@
...
+    }else if(brakePedalPercent >= NORMAL_BRAKE_THRESHOLD){
+        state = BRAKE_STATE;
```
babusid, line 419:
```suggestion
        state = FSM[BRAKE_STATE];
```

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,442 @@
...
+        velocitySetpoint = MAX_VELOCITY;
+        currentSetpoint = percentToFloat(accelPedalPercent);
+    }else if(!reverseSwitch && !forwardSwitch){
+        currentSetpoint = 0;
+    }
+}
+
+void AccelerateCruiseDecider(){
+    if(accelPedalPercent == 0){
+        state = COASTING_CRUISE;
```
babusid, line 438:
```suggestion
        state = FSM[COASTING_CRUISE];
```

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,442 @@
...
+    }else if(brakePedalPercent >= NORMAL_BRAKE_THRESHOLD){
+        state = BRAKE_STATE;
```
babusid, line 440:
```suggestion
        state = FSM[BRAKE_STATE];
```

Tests/Test_SendTritium.c
```diff
@@ -0,0 +1,337 @@
+#include "Minions.h"
+#include "Pedals.h"
+#include "FaultState.h"
+#include "CANbus.h"
+#include "MotorController.h"
+#include "ReadCarCAN.h"
+#include "BSP_UART.h"
+#include "Tasks.h"
+#include "SendTritium.h"
+
```
babusid, line 10: Should this file not define the test macro? or is that defined somewhere else?

Tests/Test_SendTritium.c
```diff
@@ -0,0 +1,337 @@
...
+}
+
+
+void goToOnePedalDrive(){
+    goToNormalState();
+
+    // One Pedal Drive
+    cruiseEnable = false;
+    regenToggle = true;
+    ChargeEnable = true;
```
babusid, line 56: I'm guessing the raw ChargeEnable value is used a lot in this file, but it shouldn't be. Assuming you implemented the suggestions i made above in SendTritium, change this to `regen_enable` and it should work pretty much the same way, but we're protecting the value inside of ReadCarCAN properly.

babusid (changes requested):

Apps/Src/SendTritium.c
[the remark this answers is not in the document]
IshDeshpa: Can we change the name to ChargeEnable in ReadCarCAN? I think it makes more sense as a naming convention so that we don't mix up the one pedal drive mode enable with the charge enable message.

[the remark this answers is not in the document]
IshDeshpa: Also yes, I will use the getter and update a local variable.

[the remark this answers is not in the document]
IshDeshpa: Also in terms of naming, I'm going to change the name of regenToggle to onePedalToggle to be more consistent in calling it "One Pedal Drive Mode" so that we don't get confused between the regen enable message from BPS and the one pedal drive enable button.

Apps/Inc/SendTritium.h
```diff
@@ -0,0 +1,6 @@
+#ifndef __SENDTRITIUM_H
+#define __SENDTRITIUM_H
+
+//#define __TEST_SENDTRITIUM
+
```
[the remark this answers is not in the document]
IshDeshpa: Currently the inputs are all externed within the test file. If I move them to the header file, they don't seem to be recognized by the test file for some reason. Do you think all inputs that we're going to mess with during testing should be moved to the header file and ifdefed in this way?
e.g. in SendTritium.h:
```#ifdef __TEST_SENDTRITIUM
// Inputs
extern bool cruiseEnable;
extern bool cruiseSet;
extern bool onePedalEnable;
extern uint8_t brakePedalPercent;
extern uint8_t accelPedalPercent;
extern bool reverseSwitch;
extern bool forwardSwitch;
extern uint8_t state;
extern float currentSetpoint;
extern float velocitySetpoint;
extern float velocityObserved;
extern float cruiseVelSetpoint;

#endif
```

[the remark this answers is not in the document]
IshDeshpa: Additionally, you may have noticed that I removed the conditionally static scope on all inputs. Should they always be static, non-static, or should I add the conditional static behavior based on if the test macro is defined?

Tests/Test_SendTritium.c
IshDeshpa: Defined in SendTritium.h. We can define here if that's more beneficial (it makes more sense to me but idk if it will work).

Apps/Inc/SendTritium.h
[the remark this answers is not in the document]
babusid: If all of those variables are modified in the test, then yes, that should be fine. Just make sure you need all of those things. As far as the static keyword, it is an additional safety guard to ensure that certain quantities are scoped to the compilation unit. In our current setup, nothing will break as a result of not having them statically scoped, but the conditional scoping was a good safeguard to have IMO.

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,585 @@
...
+#define ONEPEDAL_BRAKE_THRESHOLD 25 // percent
+#define ONEPEDAL_NEUTRAL_THRESHOLD 35 // percent
+
+#define GEAR_FAULT_THRESHOLD 3 // number of times gear fault can occur before it is considered a fault
+
+#ifdef __TEST_SENDTRITIUM
+#define STATIC(def) static def
+#else
+#define STATIC(def) def
+#endif
```
babusid, line 45: The previous SCOPE macro was probably cleaner than this, but no need to change it now.

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,585 @@
...
+
+// Inputs
+STATIC(bool cruiseEnable = false;)
+STATIC(bool cruiseSet = false;)
+STATIC(bool onePedalEnable = false;)
+STATIC(bool chargeEnable = false;)
```
babusid, line 51:
```suggestion
STATIC(bool) cruiseEnable = false;
STATIC(bool) cruiseSet = false;
STATIC(bool) onePedalEnable = false;
STATIC(bool) chargeEnable = false;
```
why not something like this? having the whole line inside the macro func is ugly

Apps/Inc/SendTritium.h
```diff
@@ -0,0 +1,28 @@
...
+#define FSM_PERIOD 20
+#define DEBOUNCE_PERIOD 2 // in units of FSM_PERIOD
+#define MOTOR_MSG_COUNTER_THRESHOLD (MOTOR_MSG_PERIOD)/(FSM_PERIOD)
+
+#ifdef __TEST_SENDTRITIUM
+// Inputs
+extern bool cruiseEnable;
+extern bool cruiseSet;
+extern bool onePedalEnable;
+extern bool chargeEnable;
```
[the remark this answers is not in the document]
KnockbackNemo: I couldn't find a chargeEnable in ReadCarCAN, but should I change the regenEnable variable/getter function name to chargeEnable so that they're more clearly separated from regen braking?

babusid (changes requested):

Apps/Inc/SendTritium.h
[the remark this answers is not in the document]
babusid: @KnockbackNemo If that's the case, then it's backward. It should be `chargeEnable` in ReadCarCAN (since ReadCarCAN is not technically supposed to know about regen, it's gating all charging). SendTritium should be the one with `regen_enable`.

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,585 @@
...
+    #ifndef __TEST_SENDTRITIUM
+    Minion_Error_t err;
+
+    // Update pedals
+    brakePedalPercent = Pedals_Read(BRAKE);
+    accelPedalPercent = Pedals_Read(ACCELERATOR);
+    
+    // Update charge enable
+    chargeEnable = RegenEnable_Get();
+
```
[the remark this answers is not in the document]
IshDeshpa: All of read inputs is already wrapped in that

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,585 @@
...
+ * @param in_min minimum value of input range
+ * @param in_max maximum value of input range
+ * @param out_min minimum value of output range
+ * @param out_max maximum value of output range 
+ * @returns integer value from out_min to out_max
+*/
+static uint8_t map(uint8_t input, uint8_t in_min, uint8_t in_max, uint8_t out_min, uint8_t out_max){
+    if(input <= in_min) return out_min;
+    else if(input >= in_max) return out_max;
+    else return ((out_max - in_max) * (input - in_min))/(out_min - in_min) + out_min;
```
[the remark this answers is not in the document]
IshDeshpa: This is applicable. Negative values are valid.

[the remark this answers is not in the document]
IshDeshpa: Although I have found a bug with in_max > in_min

IshDeshpa:
All that's left is finishing up test cases and then this should be good to merge with beta_test_0

Apps/Src/SendTritium.c
```diff
@@ -0,0 +1,621 @@
...
+ * in_min to in_max is mapped to out_min to out_max.
+ * @param input input integer value
+ * @param in_min minimum value of input range
+ * @param in_max maximum value of input range
+ * @param out_min minimum value of output range
+ * @param out_max maximum value of output range 
+ * @returns integer value from out_min to out_max
+*/
+static uint8_t map(uint8_t input, uint8_t in_min, uint8_t in_max, uint8_t out_min, uint8_t out_max){
+    if(in_max >= in_min) in_max = in_min;
```
Cam0Cow, line 258: Why is it checking that `in_max >= in_min` and then setting them equal like that? If `in_max` is larger, shouldn't it remain larger?

Cam0Cow (changes requested):

babusid:
This is in beta_test_0. Closing.

IshDeshpa:
Closes #259

PR #276, opened 2022-11-12
DebugDump

audreygpick:
The goal of this PR is to create functions to issue commands to the system for debugging, as well as create a general debugging thread to print out various variables within the system to be used for debugging or logging purposes.


NathanielDelgado:
@babusid https://stackoverflow.com/questions/9907160/how-to-convert-enum-names-to-string-in-c
This is what I am talking about. Audrey wrote these macros in one file but, in actuality, they would need to written in the driver files, something I'm a bit iffy about.

IshDeshpa:
For future reference, [this ](https://gist.github.com/awesomekling/6fec62b6a8a57264dfdc6e7e20290d93) explains what's going on with the enum autogeneration

Renode/startup.resc
IshDeshpa, line 1: We should change the name of the Renode folder to Simulator, to reduce the diff between this and master.

IshDeshpa (dismissed):
I like the new enum structure, and I like the idea behind CommandLine and DebugDump. I'd like more clarification on usecase for both, as well as the reasoning behind why CommandLine is not a regular task. This stuff should go into our ReadTheDocs documentation

Apps/Src/CommandLine.c
```diff
@@ -0,0 +1,388 @@
...
+char *save; // Save pointer for strtok_r
+
+char *help = {
+	"LHRS Controls Command Line:\n\r"
+	"	For help, enter [help]\n\r"
+	"	Format is: cmd [param, ...]\n\r"
+	"	Commands and their params are as follows:\n\r"
+};
+
+bool isWhiteSpace(char character){
```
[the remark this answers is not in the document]
NathanielDelgado: I don't know if it is better. Roie wrote that part. @Cam0Cow  is it better?

Apps/Src/CommandLine.c
```diff
@@ -0,0 +1,388 @@
...
+
+bool isNull(char character){
+	return character == '\0';
+}
+
+bool executeCommand(char *input) {
+	// The first word in the input should be a valid command
+	char *command = strtok_r(input, " ", &save);
+	// Iterate through all valid commands and check if the input matches
+	for (int i=0; cmdline_commands[i].name; i++) {
```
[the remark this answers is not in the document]
NathanielDelgado: I don't find it confusing. We are iterating a array of names and whenever it is null (last element of the command array has a null name) we stop. The alternative is to have a macro that has the number of commands. However, modifying the array would require two modifies (adding/subtracting the command and updating the number). This solution just requires you to add/subtract the command

Apps/Src/CommandLine.c
```diff
@@ -0,0 +1,388 @@
...
+	else if(strcmp(pinInput, "rev_sw") == 0){
+		pin = REGEN_SW;
+	}
+	else if(strcmp(pinInput, "cruz_en") == 0){
+		pin = REGEN_SW;
+	}
+	else if(strcmp(pinInput, "cruz_st") == 0){
+		pin = REGEN_SW;
+	}
+	// else if(strcmp(pinInput, "brakelight") == 0){	// uncomment when fix implemented
```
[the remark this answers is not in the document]
NathanielDelgado: I left that there when minions gets changed to allow the reading of outputs

Apps/Src/CommandLine.c
```diff
@@ -0,0 +1,388 @@
...
+	printf("%s is %s\n\r", pinInput, Minion_Read_Input(pin, &err) ? "on" : "off");
+	return true;
+}
+
+static bool cmd_Minion_Write_Output(void){
+	Minion_Error_t err;
+	char *pinInput = strtok_r(NULL, " ", &save);
+	MinionPin_t pin;
+	if(strcmp(pinInput, "brakelight") == 0){	// uncomment when fix implemented
+		pin = REGEN_SW;
```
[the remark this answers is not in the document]
NathanielDelgado: Typo. Fixed

Apps/Src/CommandLine.c
```diff
@@ -0,0 +1,388 @@
...
+	}
+	else if(strcmp(pinInput, "rev_sw") == 0){
+		pin = REGEN_SW;
+	}
+	else if(strcmp(pinInput, "cruz_en") == 0){
+		pin = REGEN_SW;
+	}
+	else if(strcmp(pinInput, "cruz_st") == 0){
+		pin = REGEN_SW;
```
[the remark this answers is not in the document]
NathanielDelgado: Typo. Fixed

Drivers/Inc/Minions.h
```diff
@@ -3,31 +3,34 @@
...
+    FOREACH_MinionPin(GENERATE_ENUM)
+    MINIONPIN_NUM,
+}MinionPin_t;
+
 
 //errors bc ur bad
 typedef enum{
     MINION_ERR_NONE = 0,
     MINION_ERR_YOU_READ_OUTPUT_PIN,
     MINION_ERR_YOU_WROTE_TO_INPUT_PIN,
```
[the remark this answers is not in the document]
NathanielDelgado: I'll address this in the new minions pull request

Apps/Src/DebugDump.c
```diff
@@ -0,0 +1,125 @@
...
+            uint32_t result = (OSErrLocBitmap & 1 << i) >> i;
+            if(result){
+                printf("%s ", OS_LOC_STRING[i]);
+            }
+        }
+    }
+    printf("\n");
+
+    // regen enable
+    // printf("RegenEnable: %s\n", RegenEnable == ON ? "On" : "Off");
```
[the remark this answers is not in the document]
NathanielDelgado: There is another version of DebugDump made that addresses getters for multiple functions. Not sure if this needs to be addressed here or there

Apps/Src/DebugDump.c
```diff
@@ -0,0 +1,125 @@
...
+
+    // update velocity toggle cruise
+    printf("UpdateVel_ToggleCruise: %s\n", UpdateVel_ToggleCruise == true ? "True" : "False");
+
+    // // display fifo
+    // printf("msg_queue: %d", msg_queue);
```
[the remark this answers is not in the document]
NathanielDelgado: #301 Just referencing this because I believe it should be addressed there

Tests/Test_CommandLine.c
```diff
@@ -0,0 +1,29 @@
...
+    //     }
+
+    //     // while (USART_GetFlagStatus(UART_2, USART_FLAG_RXNE) == RESET);
+    //     // printf("%c", USART_ReceiveData(UART_2));
+    // }
+    CANbus_Init(CAN_1);
+    Contactors_Init();
+    Minion_Init();
+    Pedals_Init();
+    Task_CommandLine(NULL);
```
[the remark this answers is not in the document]
NathanielDelgado: The goal of this test was just to test the logic of the command line. The command line has a while loop in it.  There could be another test which uses it as a task but this doesn't test much else. But yes, it should be a task

Tests/Test_CommandLine.c
```diff
@@ -0,0 +1,29 @@
+#include "os.h"
+#include "Tasks.h"
+#include "BSP_UART.h"
+#include "CANbus.h"
+#include "Contactors.h"
+#include "Minions.h"
+#include "Pedals.h"
+
+char input[500] = {0};
```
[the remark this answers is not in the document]
NathanielDelgado: I was debugging UART (Something was wrong with it, I remember) and it was a remnent of it. I'll delete that along with the comments

Apps/Src/CommandLine.c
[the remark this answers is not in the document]
Cam0Cow: I just think it's a bit cleaner with the switch, but it doesn't really matter. It should be made into an inline function, though.

tiandahuang (commented):
tok

NathanielDelgado:
Created a getters macro for getting private variables in drivers and apps. The start of its use can be by using it in SendTritium

Tests/Test_DebugDump.c
IshDeshpa, line 1: This still uses outdated functionality from MotorController and IgnitionContactor.

Apps/Src/DebugDump.c
```diff
@@ -0,0 +1,141 @@
...
+        printf("Cruise Set: %s\n\r", get_cruiseSet() ? "true" : "false");
+        printf("One Pedal Enable: %s\n\r", get_onePedalEnable() ? "true" : "false");
+        printf("Regen Enable: %s\n\r", get_regenEnable() ? "true" : "false");
+        printf("Pedal Brake Percent: %d\n\r", get_brakePedalPercent());
+        printf("Pedal Accel Percent: %d\n\r", get_accelPedalPercent());
+        // printf("Pedal Accel Percent: %s\n\r", get_accelPedalPercent()); // Gear
+        // printf("Current Setpoint: %f", get_currentSetpoint());   // float print not working?
+        // printf("Current Setpoint: %f", get_currentSetpoint());
+        // printf("Current Setpoint: %f", get_currentSetpoint());
+        // printf("Current Setpoint: %f", get_currentSetpoint());
```
IshDeshpa, line 96: There's a print_float function in SendTritium. We can move that to somewhere and use it.

Tests/Test_CommandLine.c
```diff
@@ -0,0 +1,18 @@
...
+#include "CANbus.h"
+#include "Contactors.h"
+#include "Minions.h"
+#include "Pedals.h"
+
+/* Test does not use OS. This is just to test the logic */
+
+int main(void){
+    BSP_UART_Init(UART_2);
+    CANbus_Init(CAN_1);
```
IshDeshpa, line 13: This is the older version of CANbus_Init. Does not compile with newer version.

Apps/Src/DebugDump.c
```diff
@@ -0,0 +1,141 @@
...
+
+        // Get pedal information
+        int8_t accelPedal = Pedals_Read(ACCELERATOR);
+        printf("ACCELERATOR: %d\n\r", accelPedal);
+
+        int8_t brakePedal = Pedals_Read(BRAKE);
+        printf("BRAKE: %d\n\r", brakePedal);
+
+        // Get minion information
+        for(MinionPin_t pin = 0; pin < MINIONPIN_NUM-1; pin++){ // Plan to change MINIONPIN_NUM-1 -> MINIONPIN_NUM after Minion function is fixed
```
IshDeshpa, line 74: What does this mean?

Apps/Inc/getter.h
```diff
@@ -0,0 +1,8 @@
+#define GETTER(type, name) \
+    type get_##name(void){ \
+        return name; \
+    } \
+
+#define GETTER_H(type, name) \
+    type get_##name(void); \
+
```
IshDeshpa, line 8: This is a cool way to set up getter functionality.

Apps/Src/CommandLine.c
```diff
@@ -0,0 +1,361 @@
...
+	}
+	else if(strcmp(contactorInput, "motor_c") == 0){
+		contactor = MOTOR_CONTACTOR;
+	}
+	else{
+		return false;
+	}
+
+	char *stateInput = strtok_r(NULL, " ", &save);
+	State state;
```
IshDeshpa, line 251: This should not use the outdated state enum. Contactors_Set uses bool.

Apps/Src/CommandLine.c
```diff
@@ -0,0 +1,361 @@
...
+const struct Command cmdline_commands[] = {
+	{.name = "help", .action = cmd_help},
+	{.name = "CANbus_Send", .action = cmd_CANbus_Send},
+	{.name = "CANbus_Read", .action = cmd_CANbus_Read},
+	{.name = "Contactors_Get", .action = cmd_Contactors_Get},
+	{.name = "Contactors_Set", .action = cmd_Contactors_Set},
+	{.name = "Minion_Read_Input", .action = cmd_Minion_Read_Input},
+	{.name = "Minion_Write_Output", .action = cmd_Minion_Write_Output},
+	{.name = "Pedals_Read", .action = cmd_Pedals_Read},
+	{.name = NULL, .action = NULL}
```
IshDeshpa, line 47: Why do we need a NULL action?

Apps/Src/CommandLine.c
```diff
@@ -0,0 +1,361 @@
...
+// Function Implementations
+// ------------------------------------------------------------------------
+
+static inline bool cmd_help(void) {
+	printf(help);
+	return true;
+}
+
+/* This has not been tested and not sure if this implementation is correct (not sure
+if we want to just transmit string data from serial) */
```
IshDeshpa, line 135: Let's get this tested on renode at least a little bit.

Drivers/Inc/Contactors.h
```diff
@@ -13,13 +13,19 @@
...
+        contactor(ARRAY_CONTACTOR)   \
+        contactor(ARRAY_PRECHARGE)  \
+        contactor(MOTOR_CONTACTOR)   \
+
+#define GENERATE_ENUM(ENUM) ENUM,
+#define GENERATE_STRING(STRING) #STRING,
+
+typedef enum contactor_ENUM {
+    FOREACH_contactor(GENERATE_ENUM)
+    CONTACTOR_NUM,
```
IshDeshpa, line 26: NUM_CONTACTORS makes more sense to me.

Drivers/Inc/Minions.h
```diff
@@ -3,28 +3,32 @@
...
+        PIN(CRUZ_EN)  \
+        PIN(CRUZ_ST)  \
+        PIN(BRAKELIGHT)  \
+
+#define GENERATE_ENUM(ENUM) ENUM,
+#define GENERATE_STRING(STRING) #STRING,
+
+typedef enum MINIONPIN_ENUM {
+    FOREACH_MinionPin(GENERATE_ENUM)
+    MINIONPIN_NUM,
```
IshDeshpa, line 23: NUM_MINIONPINS here makes more sense to me.

IshDeshpa (changes requested):
I'd like Test_CommandLine and Test_DebugDump to be compiled and tested with Renode. There are some compile errors with old/new versions of drivers as of right now.

Apps/Src/DebugDump.c
NathanielDelgado: I was thinking of changing the minion read function to include outputs (the brakelight). But I feel this doesn't matter right now. This should happen when we move over to the embedded sharepoint. I feel like we should just leave the comment in there for now so we can be reminded when we port the code.

Apps/Src/CommandLine.c
NathanielDelgado: This is so we can have one point of change when adding a function to the command line task. If we specified a size, then when adding a function we'd have to change the size. This serves as a sentinel. This way, we only need to add the function.

IshDeshpa (changes requested):
Good stuff

Apps/Src/DebugDump.c
```diff
@@ -57,6 +57,15 @@ static const char *FAULT_BITMAP_STRING[] = {
...
 /*----------------------------------------------*/
 
+static void print_f(char * str, float f) {
+    printf(str);
+
+    int n = (int) f;
+    f -= n;
+    f *= (10 << 2);
+    printf("%d.%d\n\r", n, (int) f);
+}
```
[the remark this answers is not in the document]
Cam0Cow: That function is broken, even if you change the multiply to be by 100 instead of 10 << 2 == 40. It doesn't handle negative numbers at all, for example.

Apps/Inc/getter.h
```diff
@@ -0,0 +1,8 @@
+#define GETTER(type, name) \
+    type get_##name(void){ \
```
tiandahuang, line 2: Using `##` on its own doesn't do macro expansion, you need a second layer: https://stackoverflow.com/questions/12630548/c-expand-macro-with-token-pasting

Apps/Src/CommandLine.c
```diff
@@ -0,0 +1,361 @@
...
+		case 0x09:
+		case 0x0A:
+		case 0x0C:
+		case 0x0D:
+		case 0x20: return true;
+		default: return false;
+	}
+}
+
+static inline bool isNull(char character){
```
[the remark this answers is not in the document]
NathanielDelgado: I believe not. It was in there because there was an implementation that did not use strtok. It's uneeded now. I took it out.

Apps/Src/DebugDump.c
```diff
@@ -0,0 +1,147 @@
...
+
+static const char *FAULT_BITMAP_STRING[] = { 
+    "FAULT_NONE", 
+    "FAULT_OS", 
+    "FAULT_UNREACH", 
+    "FAULT_TRITIUM",
+    "FAULT_READBPS",
+    "FAULT_DISPLAY",
+    "FAULT_BPS"
+};
```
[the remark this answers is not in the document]
NathanielDelgado: For now, I have been unable to think of how the macro should work with these, as they are bit flags (each line has "=0x01, =0x02, =0x04, etc..." I'll think about it but unsure of right now. I'll make the change if I can think of one.

Apps/Src/DebugDump.c
```diff
@@ -0,0 +1,147 @@
...
+        } 
+
+        // Send Tritium variables
+        printf("Cruise Enable: %s\n\r", get_cruiseEnable() ? "true" : "false");
+        printf("Cruise Set: %s\n\r", get_cruiseSet() ? "true" : "false");
+        printf("One Pedal Enable: %s\n\r", get_onePedalEnable() ? "true" : "false");
+        printf("Regen Enable: %s\n\r", get_regenEnable() ? "true" : "false");
+        printf("Pedal Brake Percent: %d\n\r", get_brakePedalPercent());
+        printf("Pedal Accel Percent: %d\n\r", get_accelPedalPercent());
+        // printf("Pedal Accel Percent: %s\n\r", get_accelPedalPercent()); // Gear
```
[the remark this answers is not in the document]
NathanielDelgado: I added it as a placeholder for an actual implementation of printing the current gear. I went ahead and did that in the new commit.

Tests/Test_DebugDump.c
```diff
@@ -0,0 +1,220 @@
...
+    //     (CPU_STK*)FaultState_Stk,
+    //     (CPU_STK_SIZE)WATERMARK_STACK_LIMIT,
+    //     (CPU_STK_SIZE)TASK_FAULT_STATE_STACK_SIZE,
+    //     (OS_MSG_QTY)0,
+    //     (OS_TICK)0,
+    //     (void*)NULL,
+    //     (OS_OPT)(OS_OPT_TASK_STK_CLR),
+    //     (OS_ERR*)&err
+    // );
+    // assertOSError(OS_MAIN_LOC, err);
```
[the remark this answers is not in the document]
NathanielDelgado: When I originally started testing, it was faulting due to a typo, but I had commented this out in the process. Reran the test with this enabled and works just fine. Changed in the new commit.

Apps/Src/DebugDump.c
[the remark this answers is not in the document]
NathanielDelgado: Updated to the version in motor_sim

Apps/Src/CommandLine.c
```diff
@@ -0,0 +1,357 @@
...
+// *********** Command line (shell) ************
+void Task_CommandLine(void* p_arg) { 
+	OS_ERR err;
+
+	// output welcome/help screen
+	printf(help);
+	
+	while(1){
+		printf("> ");
+		BSP_UART_Read(UART_2, input);
```
babusid, line 107: This can probably only be resolved through testing, but I would make sure that this doesn't block without being an OS-integrated blocking call. Because if it blocks and doesn't hit a scheduling point, you won't context switch out. If it's a non-blocking read ur chillin.

Apps/Src/CommandLine.c
babusid, line 1: This isn't really a review comment but more of a future consideration. It could be cool to write some frontend (python/bash/whatever) that leverages these inbuilt commands to either give a more flexible CLI environment or implements a debug UI

Apps/Src/CommandLine.c
```diff
@@ -0,0 +1,357 @@
...
+
+	if(CANbus_Send(msg, blocking, bus)){
+		printf("msg sent on %s (%s)\n\r", "can", blockInput);
+	}else{
+		printf("msg sent failed\n\r");
+	}
+	return true;
+}
+
+static bool cmd_CANbus_Read(void){
```
babusid, line 171: I'm interested to see how this will interact with the readcarcan thread. If it causes some sort of weird behavior / data races, you might have to rethink this (if debug steals a message from ReadCarCAN or vice versa). An alternate approach could be for ReadCarCAN to have a getter available that returns the most recently read CAN message, just so that you can make sure that it sees all messages on the carCAN line. This is just a curiosity though, pure functionality looks fine.

Apps/Src/CommandLine.c
```diff
@@ -0,0 +1,357 @@
...
+	}
+	else{
+		return false;
+	}
+
+	printf("%s state: %s\n", contactorInput, Contactors_Get(contactor) == ON ? "on" : "off");
+	return true;
+}
+
+static bool cmd_Contactors_Set(void){
```
babusid, line 230: Might run into competition against other threads / logic when this is being used to manually set a contactor. Be interested to see what happens in real life / testing.

Apps/Src/DebugDump.c
```diff
@@ -0,0 +1,130 @@
...
+    FOREACH_MinionPin(GENERATE_STRING)
+};
+
+static const char *CONTACTOR_STRING[] = {
+    FOREACH_contactor(GENERATE_STRING)
+};
+
+static const char *GEAR_STRING[] = {
+    FOREACH_Gear(GENERATE_STRING)
+};
```
babusid, line 29: I understand the utility for the Minion Pin one, but the other two are so small that I'm not sure it's really all that useful to have it macro-ed away. Maybe if there was a more generic foreach macro that we could pass an enum definition into alongside the operator function?

BSP/STM32F413/Src/BSP_UART.c
```diff
@@ -119,7 +119,7 @@ static void USART_USB_Init() {
     // Enable NVIC
     NVIC_InitTypeDef NVIC_InitStructure;
     NVIC_InitStructure.NVIC_IRQChannelPreemptionPriority = 1;
-    NVIC_InitStructure.NVIC_IRQChannelSubPriority = 1;
+    NVIC_InitStructure.NVIC_IRQChannelSubPriority = 0;
```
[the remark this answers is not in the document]
babusid: IMO on controls, I don't think you'll seriously notice issues with interrupt priority, considering you have like three right now (CAN1, CAN3, UART) (?) Everything else is polling based afaik.

BSP/STM32F413/Src/BSP_UART.c
```diff
@@ -166,25 +166,28 @@ uint32_t BSP_UART_Read(UART_t usart, char *str) {
...
-            recvd += rxfifo_get(fifo, (char*)str++);
-            rxfifo_peek(fifo, &data);
-        }
+    USART_TypeDef *usart_handle = handles[usart];
 
-        rxfifo_get(fifo, &data);
-        *str = 0;
-        *line_recvd = false;
-        USART_ITConfig(usart_handle, USART_IT_RXNE, SET);
+    USART_ITConfig(usart_handle, USART_IT_RXNE, RESET);
```
babusid, line 176: just test uart read to make sure it works, because afaik, we haven't actually used it anywhere important before

Drivers/Inc/Contactors.h
```diff
@@ -13,13 +13,19 @@
...
+        contactor(ARRAY_PRECHARGE)  \
+        contactor(MOTOR_CONTACTOR)   \
+
+#define GENERATE_ENUM(ENUM) ENUM,
+#define GENERATE_STRING(STRING) #STRING,
+
+typedef enum contactor_ENUM {
+    FOREACH_contactor(GENERATE_ENUM)
+    NUM_CONTACTORS,
+}contactor_t;
```
babusid, line 27: again i think this is a little unnecessary

Drivers/Inc/Minions.h
```diff
@@ -3,28 +3,32 @@
...
+        PIN(BRAKELIGHT)  \
+
+#define GENERATE_ENUM(ENUM) ENUM,
+#define GENERATE_STRING(STRING) #STRING,
+
+typedef enum MINIONPIN_ENUM {
+    FOREACH_MinionPin(GENERATE_ENUM)
+    NUM_MINIONPINS,
+} MinionPin_t;
+
```
babusid, line 25: same as before. If the only reuse of these foreach macros is in the CLI, I feel like there has to be a more elegant solution to this

babusid:
This PR is also in desperate need of a solid description

Apps/Src/CommandLine.c
```diff
@@ -0,0 +1,357 @@
...
+const struct Command cmdline_commands[] = {
+	{.name = "help", .action = cmd_help},
+	{.name = "CANbus_Send", .action = cmd_CANbus_Send},
+	{.name = "CANbus_Read", .action = cmd_CANbus_Read},
+	{.name = "Contactors_Get", .action = cmd_Contactors_Get},
+	{.name = "Contactors_Set", .action = cmd_Contactors_Set},
+	{.name = "Minion_Read_Input", .action = cmd_Minion_Read_Input},
+	{.name = "Minion_Write_Output", .action = cmd_Minion_Write_Output},
+	{.name = "Pedals_Read", .action = cmd_Pedals_Read},
+	{.name = NULL, .action = NULL}
```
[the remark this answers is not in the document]
Cam0Cow: But then it might be confused as a command with (currently) no action. I think it's better to leave the sentinel as all NULL. But it's probably a good idea to leave a comment specifically saying it's a sentinel.

Apps/Src/CommandLine.c
```diff
@@ -0,0 +1,357 @@
...
+		contactor = ARRAY_PRECHARGE;
+	}
+	else if(strcmp(contactorInput, "motor_c") == 0){
+		contactor = MOTOR_CONTACTOR;
+	}
+	else{
+		return false;
+	}
+
+	printf("%s state: %s\n", contactorInput, Contactors_Get(contactor) == ON ? "on" : "off");
```
Cam0Cow, line 226: Missing a \r, output looks weird

Apps/Inc/SendTritium.h
```diff
@@ -1,15 +1,24 @@
...
+        GEAR(FORWARD_GEAR)   \
+        GEAR(NEUTRAL_GEAR)  \
+        GEAR(REVERSE_GEAR)   \
+
+#define GENERATE_ENUM(ENUM) ENUM,
+#define GENERATE_STRING(STRING) #STRING,
+
+typedef enum GEAR_ENUM {
+    FOREACH_Gear(GENERATE_ENUM)
+    NUM_GEARS,
```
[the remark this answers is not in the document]
NathanielDelgado: The only reason for doing this macro is so that it can be printed out with the string. Without it, that couldn't be the case. I do agree that it looks complex though. I'll look into a better way to make it seem cleaner.

Apps/Inc/SendTritium.h
```diff
@@ -1,15 +1,24 @@
...
-	FORWARD_GEAR,
-	NEUTRAL_GEAR,
-	REVERSE_GEAR
+#define FOREACH_Gear(GEAR) \
+        GEAR(FORWARD_GEAR)   \
+        GEAR(NEUTRAL_GEAR)  \
+        GEAR(REVERSE_GEAR)   \
+
+#define GENERATE_ENUM(ENUM) ENUM,
+#define GENERATE_STRING(STRING) #STRING,
```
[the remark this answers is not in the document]
NathanielDelgado: I agree. It should be in common.h. Moved it to there.

Apps/Inc/getter.h
```diff
@@ -0,0 +1,8 @@
...
+        return name; \
+    } \
+
+#define GETTER_H(type, name) \
+    type get_##name(void); \
```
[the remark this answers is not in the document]
NathanielDelgado: Renamed to EXPOSE_GETTER and made a comment explaining it.

Apps/Inc/SendTritium.h
```diff
@@ -50,4 +59,17 @@ extern float velocityObserved;
...
+GETTER_H(bool, cruiseSet)
+GETTER_H(bool, onePedalEnable)
+GETTER_H(bool, regenEnable)
+GETTER_H(uint8_t, brakePedalPercent)
+GETTER_H(uint8_t, accelPedalPercent)
+GETTER_H(Gear_t, gear)
+GETTER_H(float, currentSetpoint)
+GETTER_H(float, velocitySetpoint)
+GETTER_H(float, cruiseVelSetpoint)
+GETTER_H(float, velocityObserved)
```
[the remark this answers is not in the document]
NathanielDelgado: Added a comment explaining what it is.

Apps/Inc/Tasks.h
```diff
@@ -26,6 +26,8 @@
 #define TASK_UPDATE_DISPLAY_PRIO            6
 #define TASK_SEND_CAR_CAN_PRIO              8
 #define TASK_TELEMETRY_PRIO                 9
+#define TASK_DEBUG_DUMP_PRIO                10
+#define TASK_COMMAND_LINE_PRIO              10
```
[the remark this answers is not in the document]
NathanielDelgado: I'm going to put Command Line lower because I do not want the UART prints to interfere with each other.

Apps/Inc/Tasks.h
```diff
@@ -110,11 +115,6 @@ extern OS_SEM FaultState_Sem4;
  */
 
 
-//Put all global state variables here
-extern bool UpdateVel_ToggleCruise;
-extern uint16_t SupplementalVoltage;
-extern uint32_t StateOfCharge;
```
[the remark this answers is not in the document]
NathanielDelgado: They didn't do anything. They were only mentioned there (VSCode's global searched only showed them in that file).

Apps/Inc/getter.h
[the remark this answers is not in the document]
NathanielDelgado: Agreed. Moved to common.h.

Apps/Src/CommandLine.c
```diff
@@ -0,0 +1,357 @@
...
+}
+
+static bool executeCommand(char *input) {
+	// The first word in the input should be a valid command
+	char *command = strtok_r(input, " ", &save);
+	// Iterate through all valid commands and check if the input matches
+	for (int i=0; cmdline_commands[i].name; i++) {
+		if (!strcmp(command, cmdline_commands[i].name)) {
+			return cmdline_commands[i].action(); // Execute the command
+		}
```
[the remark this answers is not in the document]
NathanielDelgado: It can be added in, I suppose, but I don't think it really matters. If we want to use lowercase for the commands, then we can just put the command to be in lowercase.

Apps/Src/CommandLine.c
NathanielDelgado: The function is blocking in that it just has a while loop waiting for a line to be received, which is signaled by the ISR. Not sure why it would be a problem if it blocks. If it blocks, then it acts like an idle task and does nothing and lets the RTOS context switch whenever it needs to. The only other thing to do would be to force a context switch while waiting to not waste CPU cycles but this is not much of an optimization.

Apps/Src/CommandLine.c
NathanielDelgado: I wrote a GUI that just executes what the CLI does. However, this didn't work with renode since the UART communication is bugged (skips every so character). However, if time permits, I can test on the hardware and see how it goes.

Apps/Src/CommandLine.c
NathanielDelgado: I think that might happen, but I will wait for testing to be done to confirm this. @Cam0Cow. I do like the alternative approach.

Apps/Src/CommandLine.c
NathanielDelgado: I don't think it would matter since the contactors is protected by a mutex, so each thread would wait. But we can confirm this in testing. @Cam0Cow

Apps/Src/DebugDump.c
NathanielDelgado: I agree that it is sort of tedious to do it on these enums since they are so small, but it is best to just use it for the sake of scalability. I'll look into a better way to make it cleaner. It currently seems difficult because you'd have to pass in a variable number of variables into a macro. I do not currently see a way to access individual variables within the macro (just can pass the variables as a whole to another function).

BSP/STM32F413/Src/BSP_UART.c
NathanielDelgado: We've tested and it does work.

Apps/Src/CommandLine.c
NathanielDelgado: Corrected

Apps/Src/CommandLine.c
babusid: Are we sure the idle task doesn't hit a scheduling point? I'd be wary of relying on the RTOS to switch without one, but maybe the priority override will handle it. This could be potentially made more "guaranteed" by using a setup similar to the CAN driver, but if it's shown to work then we're chilling.

Apps/Src/CommandLine.c
Cam0Cow: Haven't had a chance to test this yet but I agree with Sid, we should cache the last CAN message received and read that.

Apps/Src/CommandLine.c
Cam0Cow: I think this one would make sense, but I've yet to test it

IshDeshpa:
@NathanielDelgado Would a macro like [this](https://stackoverflow.com/questions/5530248/creating-a-string-list-and-an-enum-list-from-a-c-macro) maybe look cleaner?

Cam0Cow (approved):
Works on hardware

NathanielDelgado:
> @NathanielDelgado Would a macro like [this](https://stackoverflow.com/questions/5530248/creating-a-string-list-and-an-enum-list-from-a-c-macro) maybe look cleaner?

Somewhat. The problem with the top answer in the post is that the macro's arguments are not variable, so we'd have to change the macro for each enum and would not be scalable. There does not seem a way for this to be solved that I can see. The only thing is there are macros that take variable arguments but those arguments can't be accessed individually in the macro (only as a whole).

IshDeshpa (approved):

PR #277, opened 2022-11-12
Minion driver update

NathanielDelgado:
- Added a test file for testing the ignition contactor task
- Rewrote the minions read function to read both inputs and outputs
- Changed some other code due to the function rewrite


IshDeshpa:
https://github.com/lhr-solar/Controls/blob/478e97d3fdab07e5a5a11aca101fcc89591d236c/Drivers/Inc/Minions.h#L7-L13
Can we please remove this unnecessary comment and change the names of these errors to be more concise?

NathanielDelgado:
According to the documentation in stm32f4xx_gpio.c, the BSP function to write to the pins is atomic. Thus, the mutex is not needed.

NathanielDelgado:
Add a comment and documentation about why motor and array contactors are treated differently

babusid (dismissed):
lgtm, just need to run the ignition contactor test on the test candidate. We'll merge this once the tests pass.

babusid:
This is in beta_test_0. Closing.

PR #278, opened 2022-11-14
Read tritium rewrite

millergage:
Wrote functions to a) find if an error has occurred using motor controller status, and b) read which error has been triggered, if any.


Apps/Src/ReadTritium.c
```diff
@@ -5,29 +5,86 @@
...
+ * 
+ */
+tritium_error_code_t MotorController_getTritiumError(void){
+    //TODO: implement for loop to parse this
+	for(int i = 0; i < NUM_TRITIUM_ERRORS; ++i){
+		if(Motor_FaultBitmap & (1<<i)){
+			return ((tritium_error_code_t)(1<<i));
+		}
+	}
+}
```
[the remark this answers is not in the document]
babusid: Also, is this used anywhere as of right now? I assume yall are planning on using this to differentiate fault behaviors inside the fault assertion thread, because if not, then it doesn't have many purposes.

Apps/Src/ReadTritium.c
```diff
@@ -5,27 +5,86 @@
...
-	while (1)
-	{
-		CANbuff buf;
-		ErrorStatus status = MotorController_Read(&buf);
+		if (status == SUCCESS){
+			switch(dataBuf.ID){
+				case MOTOR_STATUS:{
+					// motor status error flags is in bytes 4-5
+					uint16_t errorFlags = *((uint16_t*)(&dataBuf.data[4]));
+					Motor_FaultBitmap = errorFlags;
```
babusid, line 75: Why not condense these two lines into one? If it is just for highlighting that we are storing error flags, a comment explaining that is a better way to do this.

Apps/Src/ReadTritium.c
```diff
@@ -5,27 +5,86 @@
...
+					assertTritiumError(errorFlags);
+					break;
+				}
+				default:{
+					break; //for cases not handled currently
+				}
```
babusid, line 81: Make sure to explicitly handle (at the minimum) the case describing our speed. That is the second most important message after the status, as we have to display speed on the display and store it internally somewhere.

babusid (changes requested):
Currently, looks like it's trending in the right direction. Can't fully review this until the Fault thread updates are made, as fault assertion is this thread's most important responsibility. Additionally, ensure that you handle the velocity message asap.

Apps/Src/ReadTritium.c
```diff
@@ -3,30 +3,156 @@
...
+
+        Motor_FaultBitmap = T_INIT_FAIL;
+        assertTritiumError(Motor_FaultBitmap);
+        return;
+		}
+
+    OSMutexPost(&restartFinished_Mutex, OS_OPT_POST_NONE, &err);
+    assertOSError(0, err);
+
+}
```
babusid, line 149:
```C
void MotorController_Restart(void){
	CPU_TS ts;
	OS_ERR err;

	OSMutexPend(&restartFinished_Mutex, 0, OS_OPT_POST_NONE, &ts, &err);
    assertOSError(0, err);
    uint8_t data[8] = {0};
    restartFinished = false;

 	OSSemPend(&MotorController_MailSem,
            0,
            OS_OPT_PEND_BLOCKING,
            &ts,
            &err);
	assertOSError(0, err);

	ErrorStatus initCommand = BSP_CAN_Write(CAN_3, MOTOR_RESET, data, MAX_CAN_LEN); //send motor controller reset command
    if (initCommand == ERROR) { //error found, post mutex and shut off motor controller in FaultState.c
        restartFinished = true;
		MotorController_Release();

		OSMutexPost(&restartFinished_Mutex, OS_OPT_POST_NONE, &err);
        assertOSError(0, err);

        Motor_FaultBitmap = T_INIT_FAIL;
        assertTritiumError(Motor_FaultBitmap);
        return;
		}

    OSMutexPost(&restartFinished_Mutex, OS_OPT_POST_NONE, &err);
    assertOSError(0, err);

}
```


This function is what is blocking this PR from being testable, and since I want to have a test candidate before the end of the break. I'm going to go ahead and fix it myself. However, I'm going to write a detailed explanation here of what is wrong, and what my fixes are so that they can be used for future reference. @ASurjan1  and @millergage, make sure to read over these notes, and I do expect a question or two from each of you.

I'm happy to give a mini-lecture on the CAN peripheral and the nuances of the technology (both in our chip and in our software) if you think it will help you understand it better. I will likely do this near the end of the year regardless, but I can adjust it to be a priority if necessary.

# What is restart meant for?
The restart function is intended to be used a few times in the event that we cannot initialize the motor properly, or get some other type of fault where we decide the best approach is to try to "turn it off and back on again"

# What did the old version do?  What's wrong with the current implementation? 

To understand the way the old version worked, one must first understand the underlying architecture of the old system. In the previous system, this function was contained within the **Motor Controller Driver**. This driver's purpose was to communicate and interface with the Motor Controller. However, at a lower level, it was a driver meant to manage the CAN3 CAN bus peripheral. We also had the CAN Driver, which at that time, only managed the CAN1 CAN bus peripheral. This approach was redundant, which is why we changed to a single driver that manages both CAN1 and CAN3. 

However, since the functionality in the previous version was less generic, the motor controller driver (and thus the restart function) had to do much of the low-level management itself. This is why it interacted with Mailbox semaphores and the BSP - layer can functions. The semaphores were used internally to synchronize the actual hardware resources of the CAN3 peripheral.

So why does this no longer work? Well, now we have a generic CANBus driver, that manages both the CAN1 and CAN3 peripherals. **It internally handles the resource management for both peripherals**. Why does this matter? Well, the hardware mailboxes are already set up and synchronized by the driver. So putting this same functionality in a higher-level restart function doesn't make sense, and it means that function isn't actually in sync with the rest of the system. 

# What is the new version supposed to do? 
Now that we've covered why the previous version was the way that it was, and why it will no longer work with our new drivers, let's go over what the new version is supposed to do. The new version is meant to accomplish the same functionality (restarting the motor), but now uses the new CANbus driver.

# My solution

```C
void MotorController_Restart(void){
	CANDATA_t resetmsg = {0};
	resetmsg.ID = MOTOR_RESET;
	CANbus_Send(resetmsg, true, MOTORCAN);
}
```

This version constructs an all-zero can packet, sets the id to reset, and sends it using the CAN driver. The CAN driver internally does all of the resource management for the peripheral. You can also see that it is significantly shorter :)

babusid:
@millergage I'm going to close this, since we've already merged it into beta_test_0.

PR #280, opened 2022-12-10
BPS Fault Handling

KnockbackNemo:
BPS Fault Handling PR. Split into three parts:
- ReadCarCAN handling reading BPS trip message & fault behavior on BPS trip read
- Display handling evacuation message on BPS trip
- Hardware filtering in BSP_CAN to speed up CAN message reading


babusid:
@KnockbackNemo  I'm going to include this in the first fullsys test candidate, and we'll go from there.

babusid:
@KnockbackNemo I'm going to create a new issue ticket due to some things we identified at Electrical Design Review. It pertains to readcarcan, but I don't want you to create a whole new PR. Just continue this one for that issue ticket, and edit the title / description.

babusid:
😅 why does this PR touch so many files, there's no way that's right

babusid:
@KnockbackNemo i think the git history on this branch is messed up let's try to resolve it maybe Friday

KnockbackNemo:
Ah, I think the file changes came from merging beta_test_0 into ReadCarCAN-rewrite2. Should I have done that?

The history is a bit messy from me accidentally reverting something I didn't mean to, but I'm not sure what the best way to clean it is. I know commits can be squashed- would that be an option?

IshDeshpa:
I have changed the branch it's merging into from master into beta_test_0, since this is merging into the test candidate. That's why all the file changes were there, because it was comparing with the wrong branch.

Apps/Src/FaultState.c
```diff
@@ -93,8 +94,18 @@ void EnterFaultState(void) {
...
+        /* TODO: set display to show BPS has been tripped */
+
+        nonrecoverableFaultHandler();
+        return;
+    }
     else if(FaultBitmap & FAULT_READBPS){ // Missed BPS Charge enable message
         readBPS_ContactorHandler();
+
+        // Reset FaultBitmap since this is a recoverable fault
+        FaultBitmap = FAULT_NONE; // Clears everything because only one error can be set at a time
```
IshDeshpa, line 108: Fault bitmap can actually hold multiple errors, since it's a bitmap and each bit represents a different error, so you may only want to clear one bit (the FAULT_READBPS bit in this case)

Apps/Src/FaultState.c
```diff
@@ -62,6 +62,7 @@ void EnterFaultState(void) {
             } else {
                 hall_fault_cnt++;
                 MotorController_Restart(); //re-initialize motor
+                FaultBitmap = FAULT_NONE; // Clearing entire bitmap because it can only hold one error
```
IshDeshpa, line 65: Same thing here as the other comment. We only need to clear one bit as opposed to the entire bitmap.

Apps/Src/FaultState.c
```diff
@@ -107,6 +118,7 @@ void EnterFaultState(void) {
         } else {
             disp_fault_cnt++;
             Display_Reset();
+            FaultBitmap = FAULT_NONE; // Clearing bitmap is ok because it will only hold one error
```
IshDeshpa, line 121: See other comments.

IshDeshpa (changes requested):
Just a few suggestions, otherwise looks good. I'll start merging these changes along with mine and Gage's on workday.

Apps/Src/FaultState.c
Cam0Cow: Sorry, that one's on me. My impression was that if we return from fault state, then it means we shouldn't have any faults still asserted in the bitmap. @babusid thoughts?

IshDeshpa: If two faults trigger at the same time then we need to go back into fault state to handle the second fault right? Speaking of which, is there an edge case there where a higher priority fault doesn't get handled quickly enough because of something like a display error?

IshDeshpa:
There are some things I don't like about CANFiltering, mainly the fact that the ID list is included in the BSP and not passed in through the driver layers. Also a separate function for setting up the filters would be good, so that you can just call that through BSP_CAN Init. Somebody will do a full review at some point and let y'all know exactly what needs fixing.

Cam0Cow:
These changes have already been incorporated into beta_test_0.

PR #284, opened 2023-01-05
Hashtable Utility library

babusid:
I figured that this could become useful in the future, especially considering we may want to add additional telemetry systems moving forward. Not an important PR, but it would be nice to have this merged in for use at some point.

### DIFF
--- a/Apps/Inc/hashtable.h
+++ b/Apps/Inc/hashtable.h
@@ -1,0 +1,283 @@
+// Copyright (c) 2023 Longhorn Racing Solar
+/*
+ * Author: Sidharth N. Babu, sidharth.n.babu@gmail.com
+ * 
+ * This file implements a hashtable.
+ * 
+ * In order to use it in another file, you must import it in 
+ * a particular way.
+ * 
+ * 1. Define your data and key type, like so
+ *    #define HASH_TABLE_DATA_TYPE int
+ *    #define HASH_TABLE_KEY_TYPE int
+ * 2. Define your hashtable size, like so NOTE: this must be less than or equal to 256.
+ *    #define HASH_TABLE_SIZE 256
+ * 3. Name your hashtable, like so
+ *    #define HASH_TABLE_NAME my_fifo
+ * 4. Import this file
+ *    #include "hashtable.h"
+ * 
+ * This file includes some defaults, but they might not work for
+ * your case!
+ * 
+ * Also, this file undef's everything at the end, so you can import
+ * multiple times if you need.
+ * 
+ * If HASH_TABLE_NAME == my_hashtable, then your new data structure will be
+ * called my_hashtable_table_t.
+ * 
+ * NOTE: importantly, this does not currently support usage from
+ * header files. That is, all these types/functions are statically
+ * declared, so there cannot be a non-static hashtable at the moment.
+ */
+
+
+#ifndef _HASHTABLE_COMMON
+#define _HASHTABLE_COMMON
+//This section contains resources common to any hashtable initialized
+#include "common.h"
+//This is the lookup table for the hash function
+//It is a 256 element array, consisting of a random permutation of numbers from 0-255
+uint8_t lookup[256] =  {232, 36, 152, 47, 166, 250, 86, 137, 184, 65, 144, 
+                    175, 79, 221, 228, 109, 73, 185, 133, 58, 57, 242,
+                    84, 225, 34, 156, 123, 174, 121, 4, 241, 224, 106,
+                    124, 17, 111, 42, 155, 246, 50, 170, 159, 255, 139,
+                    223, 76, 60, 29, 8, 173, 193, 214, 52, 122, 163, 0,
+                    113, 53, 191, 77, 80, 162, 72, 88, 231, 85, 108, 234,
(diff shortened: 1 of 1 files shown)


Apps/Inc/hashtable.h
```diff
@@ -0,0 +1,283 @@
...
+#define CONCAT(A, B) _CONCAT(A, B)
+
+//Type names
+#define HASH_ENTRY_STRUCT_NAME CONCAT(HASH_TABLE_NAME, _entry_s)
+#define HASH_ENTRY_TYPE_NAME CONCAT(HASH_TABLE_NAME, _entry_t)
+#define HASH_TABLE_STRUCT_NAME CONCAT(HASH_TABLE_NAME, _table_s)
+#define HASH_TABLE_TYPE_NAME CONCAT(HASH_TABLE_NAME, _table_t)
+
+//the struct def for the entries in the hash table
+typedef struct HASH_ENTRY_STRUCT_NAME {
```
Cam0Cow, line 100: Since we don't know the alignment of the types used here, it might be better to restructure this as a struct of arrays instead of array of structs, so that we don't (potentially) lose a bunch of space to padding.

Apps/Inc/hashtable.h
```diff
@@ -0,0 +1,283 @@
...
+ * 
+ * If the type of the fifo is new_hashtable, then this function
+ * will be called new_hashtable_init()
+ * 
+ * @return an empty hashtable
+*/
+
+#define INIT CONCAT(HASH_TABLE_NAME, _init)
+static inline HASH_TABLE_TYPE_NAME INIT(){
+    HASH_TABLE_TYPE_NAME table;
```
Cam0Cow, line 125: The table will go out of scope at the end of the function. You probably want the user to pass in a pointer to an already allocated table struct.

Apps/Inc/hashtable.h
```diff
@@ -0,0 +1,283 @@
...
+ */
+
+
+#ifndef _HASHTABLE_COMMON
+#define _HASHTABLE_COMMON
+//This section contains resources common to any hashtable initialized
+#include "common.h"
+//This is the lookup table for the hash function
+//It is a 256 element array, consisting of a random permutation of numbers from 0-255
+uint8_t lookup[256] =  {232, 36, 152, 47, 166, 250, 86, 137, 184, 65, 144,
```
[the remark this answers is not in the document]
Cam0Cow: We might also want to have the ability to use different hash functions for each hashmap type we create.

Apps/Inc/hashtable.h
```diff
@@ -0,0 +1,283 @@
...
+static inline bool LOOKUP(HASH_TABLE_TYPE_NAME *table, HASH_TABLE_KEY_TYPE key, HASH_TABLE_DATA_TYPE *data){
+    //get the hash of the key using a pearson hash
+    uint8_t hash = 0;
+    uint8_t *keyref = (uint8_t*)&key;
+    for(int i = 0; i < sizeof(key); ++i){
+        hash = lookup[hash ^ keyref[i]];
+    }
+    hash = hash % HASH_TABLE_SIZE;
+
+    for(int i = 0;i<HASH_TABLE_SIZE;++i){
```
Cam0Cow, line 189: This means that a failed lookup is O(n), which is probably not desirable. There's ways to avoid this, such as adding a `deleted` boolean to each bucket; the lookup can halt whenever it reaches a slot that isn't in use and not marked as deleted. The usefulness of this trick however depends on how often we delete elements.

Apps/Inc/hashtable.h
```diff
@@ -0,0 +1,283 @@
...
+static inline bool DELETE(HASH_TABLE_TYPE_NAME *table, HASH_TABLE_KEY_TYPE key){
+    //get the hash of the key using a pearson hash
+    uint8_t hash = 0;
+    uint8_t *keyref = (uint8_t*)&key;
+    for(int i = 0; i < sizeof(key); ++i){
+        hash = lookup[hash ^ keyref[i]];
+    }
+    hash = hash % HASH_TABLE_SIZE;
+
+    for(int i = 0;i<HASH_TABLE_SIZE;++i){
```
Cam0Cow, line 221: Same idea as lookup.

Cam0Cow (commented):
I have a few comments, but I think we should hold off on this. Hashtables are complicated, and implementation details are really heavily dependent on how exactly we intend to use them. If you think we really need an implementation in the code base, then we should nail down exactly we want the hashtables to be capable of before testing and merging.

babusid:
> I have a few comments, but I think we should hold off on this. Hashtables are complicated, and implementation details are really heavily dependent on how exactly we intend to use them. If you think we really need an implementation in the code base, then we should nail down exactly we want the hashtables to be capable of before testing and merging.

I was thinking of having a solution for a future telemetry interface. It would primarily be utilized in the ReadCarCAN and ReadTritium contexts, so that we could store information in a unified data structure and have other elements of the system access by known keys. I figured it may be cleaner than having independent getters for every variable, and more scalable than having an array (in both of those tasks) with macro-defined indices to read and write to for each data point.

babusid:
@Cam0Cow :(

manthanand:
:}

PR #285, opened 2023-01-21
Add arch linux support for install script

Champers5000:
I don't wanna search up which packages I need again when I setup a new laptop


babusid:
@Champers5000 lmao arch strikes back

PR #286, opened 2023-01-28
Motor sim

IshDeshpa:


Tests/Test_Renode_MotorSimulator.c
```diff
@@ -0,0 +1,136 @@
...
+#include "Tasks.h"
+
+// Global variables
+static float velocity; 
+static float current; 
+static float total_accel;
+static CANDATA_t oldCD;
+static int ctr = 0;
+static float motor_force;
+static int forceLUTIndex;
```
IshDeshpa, line 13: These should be initialized to a default value

Tests/Test_Renode_MotorSimulator.c
```diff
@@ -0,0 +1,136 @@
...
+#include "CANbus.h"
+#include <stdio.h>
+#include "Tasks.h"
+
+// Global variables
+static float velocity; 
+static float current; 
+static float total_accel;
+static CANDATA_t oldCD;
+static int ctr = 0;
```
IshDeshpa, line 11: Make sure all ints are used with the fixed width integer types (uint8_t, uint16_t, etc.) so that we don't reserve more space than we need.

Tests/Test_Renode_MotorSimulator.c
```diff
@@ -0,0 +1,136 @@
...
+static CPU_STK Task1Stk[DEFAULT_STACK_SIZE];
+
+// Helper function to convert from current to desired FORCELUT index
+inline int currentPercentToIndex(float currentPercent){
+        return ((int)(currentPercent*FORCELUT_SIZE));
+    }
+
+// Helper function to convert time delay from ms to s
+inline float millisToS(int timeInMS) {
+    return (((float)MS_TIME_DELAY_MS) / 1000);
```
IshDeshpa, line 58: There might be a better way to do this with fixed point

Tests/Test_Renode_MotorSimulator.c
```diff
@@ -0,0 +1,136 @@
...
+        current = newCD.data[4]; // Percent from 0.0 to 1.0
+        
+        
+        forceLUTIndex = currentPercentToIndex(current); // Converts current to index for FORCELUT
+        motor_force = FORCELUT[forceLUTIndex];
+
+        // Net acceleration is dependent on the force from the motor (based on current), mass of the car, 
+        // and resistive forces which are being substituted with a constant 2m/s^2 negative acceleration
+        total_accel = ((motor_force / CAR_MASS_KG) - DECELERATION);
+        velocity += (total_accel * mmillisToS(MS_TIME_DELAY_MS));
```
IshDeshpa, line 85: Typo

Tests/Test_Renode_MotorSimulator.c
```diff
@@ -0,0 +1,136 @@
...
+        velocity = newCD.data[0]; // m/s
+        current = newCD.data[4]; // Percent from 0.0 to 1.0
+        
+        
+        forceLUTIndex = currentPercentToIndex(current); // Converts current to index for FORCELUT
+        motor_force = FORCELUT[forceLUTIndex];
+
+        // Net acceleration is dependent on the force from the motor (based on current), mass of the car, 
+        // and resistive forces which are being substituted with a constant 2m/s^2 negative acceleration
+        total_accel = ((motor_force / CAR_MASS_KG) - DECELERATION);
```
IshDeshpa, line 84: You'll need a cast to a float here to avoid integer division

Tests/Test_Renode_MotorSimulator.c
IshDeshpa, line 11: Rename this to something like message_ctr

Tests/Test_Renode_MotorSimulator.c
```diff
@@ -0,0 +1,136 @@
...
+
+    //start_OS
+    OSStart(&err);
+    assertOSError(OS_MAIN_LOC, err);
+
+    while (1);
+
+    return 0;
+
+    }
```
IshDeshpa, line 136: Indentation

Tests/Test_Renode_MotorSimulator.c
```diff
@@ -0,0 +1,136 @@
...
+#define DECELERATION 2.0
+#define CAR_MASS_KG 270
+
+static OS_TCB Task1TCB;
+static CPU_STK Task1Stk[DEFAULT_STACK_SIZE];
+
+// Helper function to convert from current to desired FORCELUT index
+inline int currentPercentToIndex(float currentPercent){
+        return ((int)(currentPercent*FORCELUT_SIZE));
+    }
```
IshDeshpa, line 54: Indentation

Tests/Test_Renode_MotorSimulator.c
ASurjan1: no. 😏

Tests/Test_Renode_MotorSimulator.c
```diff
@@ -72,22 +73,48 @@ void Task1(void *arg)
...
 
-        ++ctr;
-        if (ctr == 3)
+            } else if (setpointVelocity < velocity){
+
+                velocity -= ((float)(linear_accel * MS_TIME_DELAY_MS) / 1000); //Divide by 1000 to turn into seconds from ms
+
+            }
+
+            // NEED TO RETURN VELOCITY AS PART OF NEWCD
```
IshDeshpa, line 109: Do this

Tests/Test_Renode_MotorSimulator.c
```diff
@@ -72,22 +73,48 @@ void Task1(void *arg)
...
+            // and resistive forces which are being substituted with a constant 2m/s^2 negative acceleration
+            if (setpointVelocity < 0) {
+                total_accel = ((float)motor_force / CAR_MASS_KG - DECELERATION);
+            } else{
+                total_accel = ((float)(motor_force * -1) / CAR_MASS_KG + DECELERATION);
+            }
+
+            velocity += ((total_accel * MS_TIME_DELAY_MS) / 1000); //Divide by 1000 to turn into seconds from ms
+
+            // NEED TO RETURN VELOCITY AS PART OF NEWCD
```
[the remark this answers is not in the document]
IshDeshpa: Also do this

Tests/Test_Renode_CurVelSend.c
```diff
@@ -0,0 +1,105 @@
...
+
+            CANbus_Send(message, CAN_BLOCKING, MOTORCAN);   
+            CANbus_Read(&response, CAN_BLOCKING, MOTORCAN);
+            
+            printf("Sending (Current: %f, Velocity %f), Receiving: (Current: %f, Velocity %f)",
+            *((float*)&message.data[4]), *((float*)&message.data[0]), *((float*)&response.data[4]), 
+            *((float*)&response.data[0]));          
+
+
+            for(int d = 0; d < 1000000; d++){} //delay
```
IshDeshpa, line 35: Ideally this would be in a helper function

Tests/Test_Renode_CurVelSend.c
```diff
@@ -0,0 +1,105 @@
...
+    //initialize CANBus message
+    CANDATA_t response;
+    CANDATA_t message;
+    message.ID = MOTOR_DRIVE;
+    message.idx = 0;
+    float vel = 0.0f;
+    float i = 0.0f;
+
+
+    //velocity -> 735
```
IshDeshpa, line 19: comment should be updated

Tests/Test_Renode_CurVelSend.c
```diff
@@ -0,0 +1,105 @@
...
+        memcpy(&message.data[0], &vel, sizeof(vel));
+        memcpy(&message.data[4], &i, sizeof(i));
+        CANbus_Send(message, CAN_BLOCKING, MOTORCAN);  
+        CANbus_Read(&response, CAN_BLOCKING, MOTORCAN); 
+
+        printf("Sending (Current: %f, Velocity %f), Receiving: (Current: %f, Velocity %f)",
+        *((float*)&message.data[4]), *((float*)&message.data[0]), *((float*)&response.data[4]), 
+        *((float*)&response.data[0]));  
+
+        for(int d = 0; d < 1000000; d++){} //delay
```
IshDeshpa, line 84: Maybe a longer delay is a good idea here to let the motor sim get up to speed before making the jump.

Tests/Test_Renode_CurVelSend.c
```diff
@@ -0,0 +1,105 @@
...
```
IshDeshpa, line 101: Same idea here as above.

Tests/Test_Renode_MotorSimulator.c
```diff
@@ -2,60 +2,61 @@
...
 };
 
 // Macros for calculation the velocity of the car
 #define MS_TIME_DELAY_MS 100
 #define DECELERATION 2.0
 #define CAR_MASS_KG 270
+#define MAX_VELOCITY 20000.0f
+
+//linear acceleration used for velocity control mode
+static int linear_accel = FORCELUT[20]/(CAR_MASS_KG * 5);
```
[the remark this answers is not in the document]
KnockbackNemo: It's for generating a constant acceleration to use when modeling velocity-controlled mode. Multiplying by five makes the acceleration value 1/5 the total possible acceleration, but the choice of 1/5 was somewhat arbitrary, so we should probably explain that.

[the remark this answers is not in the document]
IshDeshpa: Some macro for that is a good idea (linear acceleration factor or smth like that). We don't like arbitrary or "magic" numbers in the code as far as possible.

Tests/Test_Renode_MotorSimulator.c
```diff
@@ -0,0 +1,163 @@
...
+    4941,
+    5250,
+    5559,
+    5868,
+    6176
+};
+
+// Macros for calculating the velocity of the car
+#define MS_TIME_DELAY_MS 100
+#define DECELERATION 2.0 // In m/s^2
```
Cam0Cow, line 47: If last year's car is anything to go by, the deceleration when coasting is less than this. Not sure if it'll make that big of a difference in the simulations though.

Tests/Test_Renode_MotorSimulator.c
```diff
@@ -0,0 +1,163 @@
...
+        
+        if (abs(setpointVelocity) == MAX_VELOCITY){
+
+            forceLUTIndex = currentPercentToIndex(current); // Converts current to index for FORCELUT
+            motor_force = FORCELUT[forceLUTIndex];
+
+            // Net acceleration is dependent on the force from the motor (based on current), mass of the car, 
+            // and resistive forces which are being substituted with a constant 2m/s^2 negative acceleration
+            if (setpointVelocity < 0) {
+                total_accel = ((float)motor_force / CAR_MASS_KG - DECELERATION);
```
Cam0Cow, line 87: `total_accel` might be negative when `motor_force` is low, which probably isn't desirable. Since the deceleration force is being modeled as a constant, it's probably best to clamp `total_accel` to be at least zero.

Tests/Test_Renode_MotorSimulator.c
```diff
@@ -0,0 +1,163 @@
...
+            } else{
+                total_accel = ((float)(motor_force * -1) / CAR_MASS_KG + DECELERATION);
```
Cam0Cow, line 89: Same as above but with the signs flipped. Also, you can just write `-motor_force` instead of multiplying it by -1.

Tests/Test_Renode_MotorSimulator.c
```diff
@@ -0,0 +1,163 @@
...
+            velocity += ((total_accel * MS_TIME_DELAY_MS) / 1000); //Divide by 1000 to turn into seconds from ms
+
+            // NEED TO RETURN VELOCITY AS PART OF NEWCD
+
+        } else {
+            setpointVelocity = ((setpointVelocity * WHEEL_DIAMETER * M_PI) / 60); // Converts setpointVelocity from RPM to m/s
+            
+            if (setpointVelocity > velocity){
+
+                velocity += ((float)(linear_accel * MS_TIME_DELAY_MS) / 1000); //Divide by 1000 to turn into seconds from ms
```
Cam0Cow, line 101: Is there a reason why the deceleration parameter doesn't appear in this calculation?

Cam0Cow (changes requested):

Tests/Test_Renode_MotorSimulator.c
ASurjan1: how exactly would one go about doing so?

IshDeshpa: Roie is saying that total_accel shouldn't be negative, so basically if it is negative just set the value of total_accel to 0. However, don't we need to allow for reverse drive? @Cam0Cow

Cam0Cow: They have a separate case for that. I think I've phrased it poorly, what I meant is that total_accel shouldn't be large enough to make the car go in reverse if it's currently going forward. total_accel will be negative when the car is just coasting.

babusid (commented):
I think this one is on me, I didn't take the time to properly explain how velocity control should really work. I'll describe a high-level overview in an issue ticket, and we can go over it on Saturday.

Tests/Test_Renode_MotorSimulator.c
```diff
@@ -0,0 +1,172 @@
...
+#include "Tasks.h"
+#include <math.h>
+
+// Global variables
+static float setpointVelocity = 0.0f; 
+static float velocity = 0.0f;
+static float current = 0.0f; 
+static float total_accel = 0.0f;
+static CANDATA_t oldCD;
+static CANDATA_t currCD; // CANmsg we will be sending
```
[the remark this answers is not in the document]
ASurjan1: No thank you. 😊

PR #288, opened 2023-01-30
Test display functionality using Renode

NathanielDelgado:
- Requires the use of com0com (online version doesn't work because of drivers. Included a working version's 7z file)
- Instructions for Windows/Linux UART communication included in sender.py in root


NathanielDelgado:
If Nextion currently can read display.tft, don't update Nextion. I updated it and can no longer read display.tft

babusid:
> If Nextion currently can read display.tft, don't update Nextion. I updated it and can no longer read display.tft

As in we can't flash it?

NathanielDelgado:
> > If Nextion currently can read display.tft, don't update Nextion. I updated it and can no longer read display.tft
> 
> As in we can't flash it?

Nevermind. Issue was fixed. Just didn't load the debugger properly

babusid:
@NathanielDelgado @IshDeshpa Has this been tested / confirmed to work? If so we should be able to merge this. Also, if this requires any other software (i remember yall saying something about this) it should be documented.

NathanielDelgado:
> @NathanielDelgado @IshDeshpa Has this been tested / confirmed to work? If so we should be able to merge this. However, I'm not sure we want to delete all that stuff out of the .resc file. What's the rationale there?

Yes, the nextion debugger can communicate with renode (Ishan uploaded a video earlier this month showing it working on Slack). We can merge it in and the documentation is in the sender.py script.

babusid:
Honestly, I don't think we need to care about this at this stage. I'm just gonna close the PR, but we can revisit it later if we need to.

PR #289, opened 2023-01-31
New Simulator 

babusid:
This PR contains the new stm32f413 simulator built on Renode, customized for Controls usage. It contains three major components:
- A BPS Simulator
- Controls Simulator
- Motor Simulator

The Motor Simulator and BPS simulator are currently pre-compiled test files. The Motor will remain this way, however, the BPS simulator in the future could become a faithful representation of the actual BPS system. This will require additional peripheral development. 

The Controls simulator is currently running the same .elf file as the actual system. We are aiming to simulate all input/output behavior of the system in an extensible manner, so as to allow for quick recreations of future control interfaces.


Simulator/startup.resc
```diff
@@ -0,0 +1,93 @@
+$bin?=@Objects/controls-leader.elf
+$bps_bin?=@Objects/car-sim.elf
+$motor_bin?=@Objects/motor-sim.elf
+$platform=@Simulator/stm32f413.repl
+
+macro reset
+"""
+    sysbus LoadELF $bin
```
NathanielDelgado, line 8: does reset actually reset the emulation by just loading in the elf again? I would think that `Clear` does the same

CarSim/car-sim.c
NathanielDelgado, line 1: This should be deleted or at least not called car-sim.c

MotorSim/motor-sim.c
NathanielDelgado, line 1: This should not be called motor-sim or should be deleted.

Simulator/startup.resc
```diff
@@ -0,0 +1,93 @@
...
+"""
+    emulation CreateServerSocketTerminal 49154 "motor-uart-con" false
+    connector Connect sysbus.usart2 motor-uart-con
+"""
+
+macro show_uart_bps
+"""
+    emulation CreateServerSocketTerminal 49155 "bps-uart-con" false
+    connector Connect sysbus.usart2 bps-uart-con
+"""
```
NathanielDelgado, line 33: Take out `-con` on the names (more typing for no reason)

Simulator/startup.resc
```diff
@@ -0,0 +1,93 @@
...
+print 'Offset: %s'%hex(offset)
+print "\n"
+"""
+
+set periphReadHook """
+print "START READ HOOK"
+print 'Periph: %s'%self
+print 'Value: %s'%hex(value)
+print 'Offset: %s'%hex(offset)
+print "\np"
```
NathanielDelgado, line 48: \np -> \n

IshDeshpa (approved):
LGTM. Some stuff like ReadCarCAN test file will be overwritten after ReadCarCAN+newsim is merged.

Cam0Cow (approved):

NathanielDelgado (approved):

babusid (commented):
let's do dis

PR #295, opened 2023-02-11
Removed Ignition Contactor task, updated contactor driver, moved ignition AND into readcarcan.

babusid:
Pretty self explanatory, but here's a brief list of changes
- Removed enable/disable logic from contactor driver
- Removed ign contactor task
- Added logical AND between bps msg and ign switch in ReadCarCAN to handle array mgmt without driver enable/disable
- moved motor contactor mgmt to init task since its low prio


JimothyGreene (approved):

babusid:
Just realized this doesn't satisfy motor precharge if we turn it off and back on again while leaving the system running. Not sure if this is an issue the same way the array is. @manthanand @ClarkPoon Does the motor need to re-precharge if we open the bypass contactor? Or is it fine because it's never disconnected from the battery after the initial precharge?

manthanand:
It isn't disconnected from battery unless BPS trips. Not sure why you would turn off the contactor tho since the motor could still be powered (but lose more power through the 47 ohm resistor).

ClarkPoon:
Yeah, the precharge resistor will connect the motor to the battery unless BPS disconnects it. 

@manthanand I think they are disconnecting it if the driver tells them to. The motor won't be able to go much (if at all) when it's connected through precharge so it probably a decent feature to have if we want to stop the driver from accidentally flooring it into someone/something

Apps/Src/ReadCarCAN.c
```diff
@@ -119,13 +119,13 @@ static inline void chargingEnable(void) {
  * 
 */
 static void arrayRestart(void *p_tmr, void *p_arg){
-
+    Minion_Error_t Merr;
     if(regenEnable){    // If regen has been disabled during precharge, we don't want to turn on the main contactor immediately after
-        Contactors_Set(ARRAY_CONTACTOR, ON, false);
+        Contactors_Set(ARRAY_CONTACTOR, (true & Minion_Read_Input(IGN_1, &Merr)), false); //turn on array contactor if the ign switch lets us
```
Cam0Cow, line 124: Why are you bitwise anding a boolean with true? Just use `Minion_Read_Input` there.

Apps/Src/main.c
```diff
@@ -11,9 +11,10 @@
 #include "MotorController.h"
 #include "Pedals.h"
 #include "CAN_Queue.h"
-
 #include "UpdateDisplay.h"
 
+#define IGN_CONT_PERIOD 10
```
Cam0Cow, line 16: Any particular reason why we do this every 10 milliseconds? That seems excessive.

Cam0Cow (changes requested):
Pretty good overall, just a few small comments

KnockbackNemo (dismissed):

Apps/Src/main.c
babusid: artefact of old code, my b. I'll change it to 100ms

Apps/Src/ReadCarCAN.c
babusid: lolzies

JimothyGreene (approved):
Reapproving

Cam0Cow (approved):
Amazing commit names

PR #296, opened 2023-02-25
Read car can+new sim

millergage:
# CAN Hardware Filtering

Implemented hardware filtering to more efficiently read necessary CAN IDs. There are many CAN IDs that are sent to the leader board that are unnecessary for it to read, so by using hardware filtering we can read only the IDs that are important to the leader board. This cuts down on computing time and power since we do not have to manually disregard unneeded messages via software. 

## General Changes / Implementation
* Wrote hardware filtering into CAN initialization. This can be found in BSP_CAN.c
* Created two test files "Test_CANFilter_BPS.C" and "Test_CANFilter_ReadBPS.c" to simulate BPS sending messages to the leaderboard (simulated in Renode).
* For CAN1, a whitelist was specified in CANConfig.c which contains four IDs we have deemed important (we can add/remove as needed).
*  Created a helper function, called before BASP_CAN_Init in CANbus.c that iterates through a given whitelist and changes every ID not found in the CANLUT to value NULL.


babusid:
This PR includes a lot of changes that aren't relevant to the PR. I think we should break it up using `cherry-pick` or `checkout` to make the PR more atomic.  @millergage let's talk about this asap.

Cam0Cow (approved):

PR #298, opened 2023-03-01
Changes to map function and scope macro

IshDeshpa:
Big changes: the map function should now actually function correctly.
Small changes: macros for 0 and 100 to remove magic numbers (ugh) as well as fixes to the conditional scope macro.

### DIFF
--- a/Apps/Src/SendTritium.c
+++ b/Apps/Src/SendTritium.c
@@ -37,24 +37,29 @@
 #define ONEPEDAL_BRAKE_THRESHOLD 25 // percent
 #define ONEPEDAL_NEUTRAL_THRESHOLD 35 // percent
 
+#define PEDAL_MIN 0  // percent
+#define PEDAL_MAX 100 // percent
+#define CURRENT_SP_MIN 0 // percent
+#define CURRENT_SP_MAX 100 // percent
+
 #define GEAR_FAULT_THRESHOLD 3 // number of times gear fault can occur before it is considered a fault
 
 #ifndef __TEST_SENDTRITIUM
-#define STATIC(def) static def
+#define SCOPE static
 #else
-#define STATIC(def) def
+#define SCOPE 
 #endif
 
 // Inputs
-STATIC(bool) cruiseEnable = false;
-STATIC(bool) cruiseSet = false;
-STATIC(bool) onePedalEnable = false;
-STATIC(bool) regenEnable = false;
+SCOPE bool cruiseEnable = false;
+SCOPE bool cruiseSet = false;
+SCOPE bool onePedalEnable = false;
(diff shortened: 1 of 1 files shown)


Apps/Src/SendTritium.c
```diff
@@ -280,10 +285,22 @@ static void readInputs(){
...
+    } else if(input >= in_max){
+        // Upper bound the input to the maximum output
+        return out_max;
+    } else{
+        // Linear mapping between ranges
+        uint8_t offset_in = input - in_min;  // If input went from A -> B, it now goes from 0 -> B-A
+        uint8_t in_range = in_max - in_min;   // Input range
+        uint8_t out_range = out_max - out_min;    // Output range
+        uint8_t offset_out = out_min;
+        return (offset_in * out_range)/in_range + offset_out;   // slope = out_range/in_range. y=mx+b so output=slope*offset_in+offset_out
```
babusid, line 302: Overall looks fine. Just have one question. If `(offset_in * out_range)` is less than `in_range`, the first term of the last line will zero out, and only offset_out will be returned. Is that desired behavior?

babusid: The reason I ask is that if `out_range` is sufficiently smaller than `in_range`, it would return the same output for a range of input values, which I just want to confirm is what it is meant to do.

Apps/Src/SendTritium.c
```diff
@@ -535,7 +552,7 @@ static void OnePedalDriveHandler(){
     if(accelPedalPercent <= ONEPEDAL_BRAKE_THRESHOLD){
         // Regen brake: Map 0 -> brake to 100 -> 0
         velocitySetpoint = 0;
-        currentSetpoint = percentToFloat(map(accelPedalPercent, 0, ONEPEDAL_BRAKE_THRESHOLD, 100, 0));
+        currentSetpoint = percentToFloat(map(accelPedalPercent, PEDAL_MIN, ONEPEDAL_BRAKE_THRESHOLD, CURRENT_SP_MAX, CURRENT_SP_MIN));
```
babusid, line 555: Does using pedal_min satisfy any noise tolerance concerns? IIRC, we have the ACCEL_PEDAL_THRESHOLD at 10% to ensure that ppl are pressing it for sure. May want to adjust the boundaries somewhat. May also need testing to confirm this.

babusid:
Overall, looks fine. One suggestion is that you could make a macro for the three use cases of the map function, just to tidy it up a bit, as you sometimes (lines 328, 388, 515, 555, 567) have very long lines that could be minimized w use of a macro. Not necessary, just an observation.

JimothyGreene (approved):

PR #301, opened 2023-03-25
New debug dump

millergage:
#Updated DebugDump

This PR added more values to be printed from DebugDump while also making sure DebugDump.c does not query UART. Getter functions were added in Apps/Drivers files that contained useful static values. Added a while(1) loop in Debug_Dump(.) with a delay of 5 seconds between each iteration of the loop. Nathaniel is going to finish implementing a string conversion function/macro for a current_state enum to print the current state of the car in Debug_Dump(.).


NathanielDelgado:
Closing this due to the changes being made here are implemented in #276

PR #305, opened 2023-04-02
Added new HAL layer and deleted old code

NathanielDelgado:
Uploaded code generated by STM32Cube MX for others to see for now.

The goal of this is to deprecate our BSP layer and import the newer HAL layer.

Not sure if this should be a branch (but is as of now so that others can see the new code) or a new repo


IshDeshpa:
Does this also incorporate FreeRTOS? I know it was mentioned in the issue ticket as being involved in some way. Can somebody explain how exactly FreeRTOS is integrated with HAL?

Issue #320, opened 2023-05-31
Testing for M2 Batch 1


IshDeshpa:
- Evac Screen Testing
- BPS Fault Testing (already tested, want to reverify)
- Contactors Driver Testing
- Contactors Integration Testing
- SendTritium Testing w/ Motor Controller
- Full System Test (will happen at beginning of Fall 2023)
- New Fault State Stress Testing
- DebugDump and Telemetry Testing
- Ignition Switch Array Delay Testing (once complete - #308 )
- Minion Driver (check order of IGN1 and IGN2, it looks flipped)
- CANBus w/ whitelist hardware testing

Please add more in the comments below if needed.

PR #321, opened 2023-06-05
M2 testing Batch 1

Cam0Cow:
#320 
- [x] - Evac Screen Testing
- [x] - BPS Fault Testing (already tested, want to reverify)
- [x] - Contactors Driver Testing
- [x] - Contactors Integration Testing
- [x] - SendTritium Testing w/ Motor Controller
- [x] - New Fault State Stress Testing
- [x] - DebugDump and Telemetry Testing
- [x] - Minion Driver (check order of IGN1 and IGN2, it looks flipped)
- [x] - CANBus w/ whitelist hardware testing


Cam0Cow (commented):
Verified on hardware

Apps/Src/SendTritium.c
```diff
@@ -634,7 +635,9 @@ void Task_SendTritium(void *p_arg){
     prevState = FSM[NEUTRAL_DRIVE];
 
     #ifndef __TEST_SENDTRITIUM
-    CANbus_Init(MOTORCAN);
+    // The CANBus initialization shouldn't be in the task! ReadTritium is higher priority
+    // and will try to read it
+    //CANbus_Init(MOTORCAN, motorCANFilterList, sizeof(motorCANFilterList) / sizeof(CANId_t));
```
IshDeshpa, line 640: Remove

Apps/Src/Tasks.c
```diff
@@ -25,6 +25,8 @@ OS_TCB CommandLine_TCB;
 
 task_trace_t PrevTasks;
 
+task_trace_t PrevTasks;
+
```
IshDeshpa, line 29: Why do we have two

Apps/Src/UpdateDisplay.c
```diff
@@ -73,7 +73,7 @@ const char* compStrings[15]= {
 static void assertUpdateDisplayError(UpdateDisplayError_t err){
 	OS_ERR os_err;
 
-	if(err != UPDATEDISPLAY_ERR_NONE){
+	if(err != UPDATEDISPLAY_ERR_NONE || err != UPDATEDISPLAY_ERR_NO_CHANGE){
```
IshDeshpa, line 76: Why do we fault on No Change? I thought it just meant that the value did not change since when we last updated it.

Tests/Test_MotorControl.c
```diff
@@ -0,0 +1,141 @@
...
+#define STACK_SIZE 256
+static CPU_STK Task1_Stk[STACK_SIZE];
+
+void p_float(float f) {
+    bool neg = f < 0;
+    if (neg) f = -f;
+    int n = (int) f;
+    f -= n;
+    printf("%s%d.%d\n\r", neg ? "-" : "", n, (int) (f*100));
+}
```
IshDeshpa, line 21: Use common.c implementation instead?

IshDeshpa (changes requested):
Small changes here and there. If all tests in this branch seem like they should work then it should be fine.

Apps/Src/UpdateDisplay.c
Cam0Cow: This is to make sure we don't fault on no change anymore.

Tests/Test_Driver_CANbus.c
```diff
@@ -0,0 +1,106 @@
...
+
+static void assert(bool);
+
+CANDATA_t dataBuf, resultBuf;
+uint64_t data = 0xdeadbeef12345678;
+
+/*
+ * NOTE: This test must be run with car CAN in loopback mode
+ * TODO: automate this, either with arguments to BSP or #define
+ */
```
NathanielDelgado, line 21: What does this mean?

Tests/Test_Driver_CANbus.c
```diff
@@ -0,0 +1,106 @@
...
+    OS_CPU_SysTickInit(SystemCoreClock / (CPU_INT32U) OSCfg_TickRate_Hz);
+
+    CANbus_Init(CARCAN, carCANFilterList, NUM_CARCAN_FILTERS);
+
+    dataBuf.ID = CHARGE_ENABLE; // First, send a value that we want to be able to receive
+    memcpy(&dataBuf.data, &data, sizeof data);
+
+    assert(CANbus_Send(dataBuf, true, CARCAN) == SUCCESS);
+    assert(CANbus_Read(&resultBuf, true, CARCAN) == SUCCESS);
+    assert(memcmp(&dataBuf, &resultBuf, sizeof dataBuf) == 0); // TODO: make sure that all 8 data bytes are copied
```
NathanielDelgado, line 36: Maybe take out the TODO if it is verified?

NathanielDelgado (commented):
LGTM. Maybe add copyright to top of tests but doesn't matter

Tests/Test_Driver_CANbus.c
Cam0Cow: Every time we want to run CAN in loopback mode we need to go to the BSP layer and change the setting. It might be nice to have a macro to set it, so the test can decide if it wants loopback enabled or not

Apps/Src/UpdateDisplay.c
IshDeshpa: Makes sense, don't know what i was thinking.

Tests/Test_Driver_CANbus.c
IshDeshpa: Can you make this an issue ticket?

IshDeshpa (approved):

NathanielDelgado (approved):

PR #322, opened 2023-06-16
ReadCarCAN Restructure

diyarajon:


Drivers/Src/Minions.c
```diff
@@ -9,8 +9,8 @@ static OS_MUTEX OutputMutex; //Mutex to lock GPIO writes to input pins
 
 //should be in sync with pin enum
 const PinInfo_t PINS_LOOKARR[MINIONPIN_NUM] = {
-    {GPIO_Pin_0, PORTA, INPUT},
     {GPIO_Pin_1, PORTA, INPUT},
+    {GPIO_Pin_0, PORTA, INPUT},
```
KnockbackNemo, line 13: Do we need to change the enum to match this?

KnockbackNemo (changes requested):
This looks great! Just added some questions about changes to consider

Apps/Src/ReadCarCAN.c
```diff
@@ -189,21 +125,25 @@ void Task_ReadCarCAN(void *p_arg)
...
-        if(status != SUCCESS) {
+        // BPS sent a message
+        ErrorStatus status = CANbus_Read(&dataBuf, true, CARCAN);
+        if(status != SUCCESS){
             continue;
         }
 
-        switch(dataBuf.ID){ //we got a message
+        switch(dataBuf.ID){
             case BPS_TRIP: {
```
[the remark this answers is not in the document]
Cam0Cow: Manthan confirmed that all BPS_TRIP messages should be treated as a trip, they don't send the message otherwise

Apps/Src/ReadCarCAN.c
```diff
@@ -212,28 +152,38 @@ void Task_ReadCarCAN(void *p_arg)
...
-                    updateSaturation(1);
+                    // Enable regen and update saturation
+                    updateSaturation(1);                               
+                    
+                    if(Minion_Read_Pin(IGN_1, &Merr) == ON                                          // Ignition is ON
+                        && chargeMsgSaturation >= SATURATION_THRESHOLD                              // Saturation Threshold has be met
+                        && (Contactors_Get(ARRAY_CONTACTOR)==OFF)                                   // Array Contactor is OFF
+                        && OSTmrStateGet(&prechargeDlyTimer, &err) != OS_TMR_STATE_STOPPED){        // Precharge Delay is not runnning and is not executing one-shot 
+
+                            assertOSError(OS_MINIONS_LOC, err);
```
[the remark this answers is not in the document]
IshDeshpa: No, assertOSError is only for OS errors, not our custom error types.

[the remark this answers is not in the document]
diyarajon: @IshDeshpa In what cases should we be using Minion_Error_t besides for it being a parameter? Would it only be for debugging reasons when testing the code?

Drivers/Src/Minions.c
diyarajon: Yes! I will change that. Also in reference to mirroring enums, I believe that the CANId_t enums do not completely mirror the CAN LUT. Should I also update the CAN LUT in this issue ticket?

diyarajon: Also, NUM_CAN_IDS is currently 1410, which is the incremented number following CARDATA. Is that intended or should it be the number of CAN IDs as enums (21 IDs)?

Apps/Src/ReadCarCAN.c
```diff
@@ -212,28 +152,38 @@ void Task_ReadCarCAN(void *p_arg)
...
+                            assertOSError(OS_READ_CAN_LOC, err);
```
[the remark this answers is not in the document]
diyarajon: Sounds good, I will do that in an if-else. 

(1) If it passes the conditional statement, then the minions and read can assertions happen before and after the OS timer start. 
(2) if it doesn't pass the conditional statement, then it asserts for both minions and read can. 

The LOCs wouldn't actually be useful for this if statement. If both locations have os errors, then the error's location will be directed to the first assert. I will make a note of this in the comments in case we do decide to use LOCs in the future.

Apps/Src/ReadCarCAN.c
```diff
@@ -212,28 +152,38 @@ void Task_ReadCarCAN(void *p_arg)
...
 
-                } else { //We got a message of enable with a nonzero value
-                    updateSaturation(1);
+                    // Enable regen and update saturation
+                    updateSaturation(1);                               
+                    
+                    if(Minion_Read_Pin(IGN_1, &Merr) == ON                                          // Ignition is ON
+                        && chargeMsgSaturation >= SATURATION_THRESHOLD                              // Saturation Threshold has be met
+                        && (Contactors_Get(ARRAY_CONTACTOR)==OFF)                                   // Array Contactor is OFF
+                        && OSTmrStateGet(&prechargeDlyTimer, &err) != OS_TMR_STATE_STOPPED){        // Precharge Delay is not runnning and is not executing one-shot
```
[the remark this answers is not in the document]
diyarajon: Thank you :) I was debating between using either (1) STOPPED or (2) RUNNING and COMPLETED. They should have opposite functionality. I ultimately decided on STOPPED because it would be one statement but I forgot to change the negation.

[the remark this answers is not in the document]
diyarajon: I believe that STOPPED should encompass that the delay is not running and it is not in one-shot. Testing *should* confirm this.  If not, this will be changed to check if it is not RUNNING or COMPLETED.

Apps/Src/ReadCarCAN.c
```diff
@@ -70,99 +64,41 @@ static void updateSaturation(int8_t chargeMessage){
...
- * @param p_arg pointer to the argument passed by timer
- * 
 */
+Minion_Error_t Merr;
 static void arrayRestart(void *p_tmr, void *p_arg){
-    Minion_Error_t Merr;
     if(chargeEnable){    // If regen has been disabled during precharge, we don't want to turn on the main contactor immediately after
-        Contactors_Set(ARRAY_CONTACTOR, (Minion_Read_Pin(IGN_1, &Merr)), false); //turn on array contactor if the ign switch lets us
+        Contactors_Set(ARRAY_CONTACTOR, (Minion_Read_Pin(IGN_1, &Merr)), false); // Turn on array contactor if the ign switch lets us
         UpdateDisplay_SetArray(true);
```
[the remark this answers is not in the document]
diyarajon: Sounds good. Thank you. This also made me realize that the display needs to be updated when ignition is turned off.

Apps/Src/ReadCarCAN.c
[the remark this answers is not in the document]
IshDeshpa: Ideally we would be making sure that Minion_Error_t is indeed equal to none upon calling the function. However, I don't think there's a need for Minion_Error_t anymore so we'd probably remove it in the minion driver fixes issue ticket. For now, don't worry about doing anything with it.

Apps/Src/ReadCarCAN.c
```diff
@@ -70,99 +64,41 @@ static void updateSaturation(int8_t chargeMessage){
...
+    callback_chargingDisable();
 }
 
 /**
  * @brief Callback function for the precharge delay timer. Waits for precharge and then restarts the array.
- * @param p_tmr pointer to the timer that calls this function, passed by timer
- * @param p_arg pointer to the argument passed by timer
- * 
 */
+Minion_Error_t Merr;
```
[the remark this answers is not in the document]
IshDeshpa: This should be inside the function, since it's a local and we don't need to do anything with it later.

Drivers/Src/Minions.c
Cam0Cow: It's intended to be a big array, since it makes looking up CAN IDs quicker

Apps/Src/ReadCarCAN.c
```diff
@@ -70,99 +64,41 @@ static void updateSaturation(int8_t chargeMessage){
...
+static void chargingDisable(void *p_tmr, void *p_arg){
+    callback_chargingDisable();
 }
 
 /**
  * @brief Callback function for the precharge delay timer. Waits for precharge and then restarts the array.
- * @param p_tmr pointer to the timer that calls this function, passed by timer
- * @param p_arg pointer to the argument passed by timer
- * 
 */
```
IshDeshpa, line 87: Let's keep the @param tags in the method comment.

Apps/Src/ReadCarCAN.c
```diff
@@ -70,99 +64,41 @@ static void updateSaturation(int8_t chargeMessage){
...
-            OSTmrStart(&prechargeDlyTimer, &err);
-            assertOSError(OS_READ_CAN_LOC, err);
-            
-        }
-    
-
-    OSMutexPost(&arrayRestartMutex, OS_OPT_NONE, &err);
-    assertOSError(OS_READ_CAN_LOC,err);
+// Nested function because same function needs to be executed
+// however the function declarations require different arguments
```
IshDeshpa, line 80: This should be a doxygen style comment (@brief, @param).

Apps/Src/ReadCarCAN.c
```diff
@@ -70,99 +64,41 @@ static void updateSaturation(int8_t chargeMessage){
     chargeMsgSaturation = newSaturation;
 }
 
-// exception struct callback for charging disable, kills contactors and turns of display
-static void callback_disableContactors(void){
-    // Kill contactors 
+// exception struct callback for charging disable, kills contactor and turns off display
+static void callback_chargingDisable(void){
```
IshDeshpa, line 68: This should be a doxygen style comment (@brief, @param).

Apps/Src/ReadCarCAN.c
```diff
@@ -23,10 +22,6 @@ static OS_TMR canWatchTimer;
 // prechargeDlyTimer
```
IshDeshpa, line 22: This comment doesn't tell us anything new.

Apps/Src/ReadCarCAN.c
```diff
@@ -40,23 +35,22 @@ static uint8_t oldestMsgIdx = 0;
...
  * @brief adds new messages by overwriting old messages in the saturation buffer and then updates saturation
  * @param chargeMessage whether bps message was charge enable (1) or disable (-1)
 */
 static void updateSaturation(int8_t chargeMessage){
+    if(chargeMessage == 1){
+        chargeEnable = true;
+    }else{
+        chargeEnable = false;
+    }
+
```
IshDeshpa, line 53: I don't think this has the intended effect. This will turn on chargeEnable if we get a chargeMessage of 1 and turn it off if we get a chargeMessage of 0. Ideally, we want chargeEnable to be turned on only if the saturation buffer is above the saturation threshold. If we get a chargeMessage of 0, we want to decrement the buffer and immediately turn chargeEnable off. If we get a chargeMessage of 1, we want to increment the buffer and only turn on chargeEnable if we're past the threshold.

Apps/Src/ReadCarCAN.c
IshDeshpa, line 179: This check works for turning on the car at startup, but what about a scenario where chargeEnable is off but the car is on? We should still increment the saturation buffer by 1 and turn on chargeEnable if the saturation buffer has passed the threshold. Otherwise, this does nothing when the Array Contactor is already on.

Apps/Src/ReadCarCAN.c
```diff
@@ -70,99 +64,41 @@ static void updateSaturation(int8_t chargeMessage){
...
- * @param p_tmr pointer to the timer that calls this function, passed by timer
- * @param p_arg pointer to the argument passed by timer
- * 
 */
+Minion_Error_t Merr;
 static void arrayRestart(void *p_tmr, void *p_arg){
-    Minion_Error_t Merr;
     if(chargeEnable){    // If regen has been disabled during precharge, we don't want to turn on the main contactor immediately after
-        Contactors_Set(ARRAY_CONTACTOR, (Minion_Read_Pin(IGN_1, &Merr)), false); //turn on array contactor if the ign switch lets us
+        Contactors_Set(ARRAY_CONTACTOR, (Minion_Read_Pin(IGN_1, &Merr)), false); // Turn on array contactor if the ign switch lets us
```
IshDeshpa, line 91: I would really like to get rid of the Minion_Read_Pin here, that was the main idea behind this issue ticket in the first place. Can we use a static variable and only read the pin in the main loop once, so that we avoid certain edge cases?

Drivers/Src/Minions.c
IshDeshpa: Oh you're right. I completely forgot about that.
Intended behavior is that each CAN ID acts as an index to the CAN LUT array, so it makes sense that the CAN LUT size is one more than the maximum possible CAN ID. This way we can just look up the CAN spec we want using the CAN ID as an index, as opposed to making it an array of structs each with its own CAN ID (in which case we'd have to linear search for the CAN ID). However, proper naming might be good here. How about MAX_CAN_ID?

IshDeshpa (changes requested):
Left some comments on better structure and documentation.

Apps/Src/ReadCarCAN.c
```diff
@@ -189,21 +125,25 @@ void Task_ReadCarCAN(void *p_arg)
...
-    
 
-    while (1)
-    {
+    while(1){
+
+        // Array Contactor is turned off if Ignition 1 is off
+        if(Minion_Read_Pin(IGN_1, &Merr) == OFF){
+            Contactors_Set(ARRAY_CONTACTOR, OFF, true);
+            assertOSError(OS_MINIONS_LOC, err);
```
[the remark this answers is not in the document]
diyarajon: @KnockbackNemo @IshDeshpa Got it. 
I had assumed that I need to use an assertOSError after every minion pin read because the main loop in the Task_Init had one after reading the minion pin. Is the one in Task_Init intended to be there?

Apps/Src/ReadCarCAN.c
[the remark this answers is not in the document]
diyarajon: Sounds good

Apps/Src/ReadCarCAN.c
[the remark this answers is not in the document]
IshDeshpa: assertOSError is only for when we run an OS function and want to make sure that it doesn't return an error. If there's any OS function being run, most (if not all) of the time it should have an assertOSError after it.

Apps/Src/ReadCarCAN.c
```diff
@@ -171,7 +107,7 @@ void Task_ReadCarCAN(void *p_arg)
         0,
         CAN_WATCH_TMR_DLY_TMR_TS,
         OS_OPT_TMR_PERIODIC,
```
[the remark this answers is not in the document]
IshDeshpa: @diyarajon Me, Maddie and Roie had a discussion about this briefly. We actually do want this to be periodic, since we need the saturation buffer to be decremented every single time we miss a charge enable message (i.e. whenever we reset the watchdog timer). Make sure you are also decrementing the saturation buffer in the callback for this.

IshDeshpa:
@diyarajon Please resolve the conversations on here that have already been addressed by you as well

Apps/Src/ReadCarCAN.c
```diff
@@ -189,53 +147,75 @@ void Task_ReadCarCAN(void *p_arg)
...
+        #ifndef __TEST_READCARCAN
+        arrayIgnitionStatus = Minion_Read_Pin(IGN_1, &Merr); // Don't check ignition if just testing
+        #endif
+
+        // Array Contactor is turned off if Ignition 1 is off
+        if(arrayIgnitionStatus == false){
+            Contactors_Set(ARRAY_CONTACTOR, OFF, true);
+
+            // Update Display
+            UpdateDisplay_SetArray(false);
```
IshDeshpa, line 169: We don't want to spam the display with messages. Let's try to do some previous/current checking here so we only send the UpdateDisplay message on the falling edge.

Apps/Src/ReadCarCAN.c
```diff
@@ -189,53 +147,75 @@ void Task_ReadCarCAN(void *p_arg)
...
+        }
```
IshDeshpa, line 170: As said above, this logic (along with the "prechargeComplete" logic) can all be put in a helper.

Apps/Src/ReadCarCAN.c
```diff
@@ -189,53 +147,75 @@ void Task_ReadCarCAN(void *p_arg)
...
-
-                    Display_Evac(SOC, SBPV);    // Display evacuation message
 
-                    // Create an exception and assert the error
-                    // kill contactors and enter a nonrecoverable fault
-                    exception_t tripBPSError = {.prio=1, .message="BPS has been tripped", .callback=callback_disableContactors};
-                    assertExceptionError(tripBPSError);
-                }
+        switch(dataBuf.ID){
+            case BPS_TRIP: { // BPS has a fault and we need to enter fault state (probably)
```
IshDeshpa, line 180:
```suggestion
            case BPS_TRIP: { // BPS has a fault and we need to enter fault state
```

Apps/Src/ReadCarCAN.c
```diff
@@ -252,8 +232,88 @@ void Task_ReadCarCAN(void *p_arg)
...
+    // Kill contactor using a direct write to avoid blocking calls when the scheduler is locked
+    BSP_GPIO_Write_Pin(CONTACTORS_PORT, ARRAY_CONTACTOR_PIN, false);
 
+    // Check that the contactor was successfully turned off
+    bool ret = (bool)Contactors_Get(ARRAY_CONTACTOR);
+
+    if(ret == false) { // Array contactor was turned off
+
+        // Turn off the array contactor display light
+        UpdateDisplay_SetArray(false);
```
IshDeshpa, line 263: This call may result in an OS error, since the scheduler is locked and UpdateDisplay makes calls to Mutex Pend/Post. Instead, let's move this logic somewhere else somehow. I'll leave that for you to figure out.

Apps/Src/TempSave_ReadCarCAN
IshDeshpa, line 1: Remove this file from git.

Tests/Test_ReadCarCANrewrite.c
```diff
@@ -0,0 +1,195 @@
...
+#define SATURATION_THRESHOLD_TEST (((SAT_BUF_LENGTH + 1) * SAT_BUF_LENGTH) / 4) 
+
+uint8_t supp_voltage_can_data = 0;
+uint32_t state_of_charge_can_data = 0;
+
+CANDATA_t bps_trip_msg = {.ID=BPS_TRIP, .idx=0, .data={1}};
+CANDATA_t charge_enable_msg = {.ID=CHARGE_ENABLE, .idx=0, .data={1}};
+CANDATA_t charge_disable_msg = {.ID=CHARGE_ENABLE, .idx=0, .data={0}};
+CANDATA_t supp_voltage_msg = {.ID=SUPPLEMENTAL_VOLTAGE, .idx=0, .data={0}};
+CANDATA_t state_of_charge_msg = {.ID=STATE_OF_CHARGE, .idx=0, .data={100}};
```
IshDeshpa, line 42:
```suggestion
static uint8_t supp_voltage_can_data = 0;
static uint32_t state_of_charge_can_data = 0;

static CANDATA_t bps_trip_msg = {.ID=BPS_TRIP, .idx=0, .data={1}};
static CANDATA_t charge_enable_msg = {.ID=CHARGE_ENABLE, .idx=0, .data={1}};
static CANDATA_t charge_disable_msg = {.ID=CHARGE_ENABLE, .idx=0, .data={0}};
static CANDATA_t supp_voltage_msg = {.ID=SUPPLEMENTAL_VOLTAGE, .idx=0, .data={0}};
static CANDATA_t state_of_charge_msg = {.ID=STATE_OF_CHARGE, .idx=0, .data={100}};
```

Tests/Test_ReadCarCANrewrite.c
```diff
@@ -0,0 +1,195 @@
...
+
+#define CARCAN_FILTER_SIZE (sizeof carCANFilterList / sizeof(CANId_t))
+
+void Task1(){
+    OS_ERR err;
+
+    CPU_Init();
+    OS_CPU_SysTickInit(SystemCoreClock / (CPU_INT32U) OSCfg_TickRate_Hz);
+    Contactors_Init();
+    Contactors_Init();
```
IshDeshpa, line 52: Why twice?

Tests/Test_ReadCarCANrewrite.c
```diff
@@ -0,0 +1,195 @@
...
+#include "CANbus.h"
+#include "CAN_Queue.h"
+#include "ReadCarCAN.h"
+#include "Contactors.h"
+#include "Display.h"
+#include "UpdateDisplay.h"
+#include "CANConfig.h"
+
+static OS_TCB Task1_TCB;
+#define STACK_SIZE 128
```
IshDeshpa, line 30: Just use DEFAULT_STACK_SIZE defined in Tasks.h. @Cam0Cow also mentioned that setting STACK_SIZE to 128 causes problems especially when the stack becomes too big.

Apps/Src/ReadCarCAN.c
```diff
@@ -20,43 +27,43 @@
...
+uint8_t SOC = 0;
+uint32_t SBPV = 0;
 
+// Error assertion function prototype
+static void assertReadCarCANError(ReadCarCAN_error_code_t rcc_err);
 
 // Getter function for chargeEnable
-bool ChargeEnable_Get() 
-{
+bool ChargeEnable_Get(void){
```
IshDeshpa, line 57: We can replace this with GETTER from common.h on master. Merge master into this branch to get the changes from there.

Apps/Src/ReadCarCAN.c
```diff
@@ -189,53 +147,75 @@ void Task_ReadCarCAN(void *p_arg)
...
+                    updateSaturation(1);                               
+                    
+                    if(arrayIgnitionStatus == true                                                  // Ignition is ON
+                        && chargeMsgSaturation >= SATURATION_THRESHOLD                              // Saturation Threshold has be met
+                        && (Contactors_Get(ARRAY_CONTACTOR)==OFF)                                   // Array Contactor is OFF
+                        && ((OSTmrStateGet(&prechargeDlyTimer, &err) == OS_TMR_STATE_STOPPED)       // Precharge Delay is not running
+                        || (OSTmrStateGet(&prechargeDlyTimer, &err) == OS_TMR_STATE_COMPLETED))){   // or has already completed
+
+                            // NOTE: An assert error for OSTmrStateGet isn't needed here because the 
+                            //       conditional should fail if there is an error.
```
IshDeshpa, line 209: I feel like we should still assert the error since a new error could emerge after running OSTmrStart.

Apps/Src/ReadCarCAN.c
```diff
@@ -20,43 +27,43 @@
...
+static bool arrayIgnitionStatus = true; // Can't test logic if ignition is always off
+#else
+static bool arrayIgnitionStatus = false;
+#endif
 
-// Handler to turn array back on
-static void arrayRestart(void *p_tmr, void *p_arg); 
+// SOC and Supp V
+uint8_t SOC = 0;
+uint32_t SBPV = 0;
```
IshDeshpa, line 51: Could make static with GETTERs

Apps/Src/ReadCarCAN.c
```diff
@@ -1,13 +1,20 @@
 /* Copyright (c) 2020 UT Longhorn Racing Solar */
 
+#define __TEST_READCARCAN
+// Uncomment the #define above when testing Renode or on hardware. Also uncomment #define in the h file
+
```
IshDeshpa, line 5: If we're going to use this to test (which I already explained why we shouldn't, but if we need to on Saturday) don't define it in both places. By including ReadCarCAN.h, it will be defined in this file as well since it is defined there.
```suggestion
```

Apps/Src/ReadCarCAN.c
```diff
@@ -68,110 +75,61 @@ static void updateSaturation(int8_t chargeMessage){
...
-
-
     // Create the CAN Watchdog (periodic) timer, which disconnects the array and disables regenerative braking
     // if we do not get a CAN message with the ID Charge_Enable within the desired interval.
     OSTmrCreate(
         &canWatchTimer,
         "CAN Watch Timer",
-        0,
-        CAN_WATCH_TMR_DLY_TMR_TS,
+        CAN_WATCH_TMR_DLY_TMR_TS, // Initial delay equal to the period since 0 doesn't seem to work
```
IshDeshpa, line 129: Zero should work. Are you sure about this?

Apps/Src/ReadCarCAN.c
```diff
@@ -189,53 +147,75 @@ void Task_ReadCarCAN(void *p_arg)
...
+            case BPS_TRIP: { // BPS has a fault and we need to enter fault state (probably)
+                
+                // Assert the error to kill contactors, display evacuation message, and enter a nonrecoverable fault
+                assertReadCarCANError(READCARCAN_ERR_BPS_TRIP);
+                
             }
-            case CHARGE_ENABLE: { 
+            case CHARGE_ENABLE:{
 
                 // Restart CAN Watchdog timer
```
IshDeshpa, line 188:
```suggestion
                // Acknowledge CAN Watchdog timer
```

IshDeshpa (dismissed):
Test the other things we need to test on Saturday first. If time permits, you can address these changes. If you don't get to testing this on Saturday, we can push it back to next Saturday. If you change it after we already verify it on hardware, we will need to verify it again, so keep that in mind.

Tests/Test_ReadCarCANrewrite.c
diyarajon: twas a mistake

Apps/Inc/ReadCarCAN.h
```diff
@@ -11,10 +11,28 @@
...
+#endif
+
+/**
+ * Error types
+ */
+typedef enum{
+	READCARCAN_ERR_NONE			    = 0,
+	READCARCAN_ERR_CHARGE_DISABLE   =-1,	// Received a charge disable msg
+	READCARCAN_ERR_MISSED_MSG		=-2,	// Didn't receive a BPS charge msg in time
+	READCARCAN_ERR_BPS_TRIP		    =-3		// Received a BPS trip msg (0 or 1)
```
NathanielDelgado, line 29: Just a stylistic change. You probably just want the enum to assign the default values, instead of incrementing them by their negative numbers. Error values don't have to be negative numbers. Instead, just let the NONE error code be default (which is at, 0) and let the others be 1, 2, ... and so on. This way, you don't have to keep assigning values if you add more errors

NathanielDelgado, line 29: Just a stylistic change. You probably just want the enum to assign the default values, instead of incrementing them by their negative numbers. Error values don't have to be negative numbers. Instead, just let the NONE error code be default (which is at, 0) and let the others be 1, 2, ... and so on. This way, you don't have to keep assigning values if you add more errors

Apps/Src/ReadCarCAN.c
```diff
@@ -189,53 +147,75 @@ void Task_ReadCarCAN(void *p_arg)
...
+        }
         
-        //Get any message that BPS Sent us
-        ErrorStatus status = CANbus_Read(&dataBuf,true,CARCAN);  
-        if(status != SUCCESS) {
+        // BPS sent a message
+        ErrorStatus status = CANbus_Read(&dataBuf, true, CARCAN);
+        if(status != SUCCESS){
             continue;
         }
```
NathanielDelgado, line 176: I don't think this does anything

Apps/Src/ReadCarCAN.c
```diff
@@ -189,53 +147,75 @@ void Task_ReadCarCAN(void *p_arg)
...
-    while (1)
-    {
+    #ifndef __TEST_READCARCAN
+    Minion_Error_t Merr; // Not used when testing since ignition won't be changed
+    #endif
+
+    while(1){
+
+        #ifndef __TEST_READCARCAN
+        arrayIgnitionStatus = Minion_Read_Pin(IGN_1, &Merr); // Don't check ignition if just testing
```
NathanielDelgado, line 161: I know the ignition inputs are inverted. Just double check the status is what it should be

Apps/Src/ReadCarCAN.c
```diff
@@ -68,110 +75,61 @@ static void updateSaturation(int8_t chargeMessage){
...
- * @param p_arg pointer to the argument passed by timer
-*/
-void canWatchTimerCallback (void *p_tmr, void *p_arg){
-    chargingDisable();
-}
+        // Update display based on if array contactor is on/off
+        if(Contactors_Get(ARRAY_CONTACTOR) == ON){
+            UpdateDisplay_SetArray(true);
+        }else{
+            UpdateDisplay_SetArray(false);
```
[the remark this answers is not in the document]
NathanielDelgado: Just curious, what is the need for having them in one place? @IshDeshpa

Apps/Src/ReadCarCAN.c
NathanielDelgado: I think the UpdateDisplay functions already do this (good point though) @IshDeshpa

Apps/Src/ReadCarCAN.c
diyarajon: @IshDeshpa @Cam0Cow and I tested it with 0 on hardware yesterday but it didn't seem to work. It only worked with that specific delay.

Apps/Src/ReadCarCAN.c
diyarajon: Thanks! This PR should also fix the inverted issue by directly switching the GPIO pins in the look up array.

Apps/Src/ReadCarCAN.c
diyarajon: @NathanielDelgado Got it. It looks like whenever there isn't a success, the ID is set to null so it wouldn't enter the switch case. I'll go ahead and remove it, unless there is something else to consider.

diyarajon: I deleted it, but it's being caught as an unused variable. Could it be helpful in the long run to keep it?

Apps/Inc/ReadCarCAN.h
diyarajon: @KnockbackNemo Could we do this?

Apps/Src/ReadCarCAN.c
NathanielDelgado: Nevermind, keep it. It's better to have that case be more apparent @diyarajon

Apps/Src/ReadCarCAN.c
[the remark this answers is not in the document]
IshDeshpa: It's far more confusing for a new person to read how the contactors are updated if one of the sets is in a callback for a timer delay and the other is in the main while loop, plus it could lead to some synchronization issues (what if the precharge timer ends, we decide to turn on the contactor, and then we decide to turn it off in the next iteration of the while loop?), plus it's easier to debug and see if the Contactors_Set call is in one spot with some boolean condition determining whether to turn it off/on.

Apps/Inc/ReadCarCAN.h
KnockbackNemo: @diyarajon Yup, that sounds good

IshDeshpa:
Just realized we never handled this here: let's change Contactors to only include the two contactors we actually have control on, ARRAY_PRECHARGE and MOTOR_CONTACTOR.

IshDeshpa:
> Just realized we never handled this here: let's change Contactors to only include the two contactors we actually have control on, ARRAY_PRECHARGE and MOTOR_CONTACTOR.

@Cam0Cow Can you address this before testing as well?

diyarajon:
A placeholder for BPS_CONTACTOR_STATUS is being used right now, and the ID has not been confirmed. BPS All Clear and BPS Contactor State have been commented out because BPS Contactor Status will take responsibility for those messages instead.

IshDeshpa:
Task linked: [CU-8685u78h3 ReadCarCAN Restructure](https://app.clickup.com/t/8685u78h3)

KnockbackNemo (changes requested):
I only was able to look through ReadCarCAN and leave some small comments, but it looks pretty good to me! I'm sorry I wasn't able to leave a more thorough review and that the comments are short- I was running low on time and wanted to get something out by this workday- but from what I've seen, it looks like nice work, Diya!

Apps/Src/ReadCarCAN.c
```diff
@@ -1,259 +1,466 @@
...
+// BPS Array and Plus/Minus saturation threshold is halfway between 0 and max saturation value 
+#define ARRAY_SATURATION_THRESHOLD (((SAT_BUF_LENGTH + 1) * SAT_BUF_LENGTH) / 4) 
+#define PLUS_MINUS_SATURATION_THRESHOLD (((SAT_BUF_LENGTH + 1) * SAT_BUF_LENGTH) / 4) 
+
+// Timer delay constants
+#define CAN_WATCH_TMR_DLY_MS 500u // 500 ms
 #define CAN_WATCH_TMR_DLY_TMR_TS ((CAN_WATCH_TMR_DLY_MS * OS_CFG_TMR_TASK_RATE_HZ) / (1000u)) //1000 for ms -> s conversion
-#define PRECHARGE_DLY_TMR_TS (PRECHARGE_ARRAY_DELAY * OS_CFG_TMR_TASK_RATE_HZ)
+
+// Precharge Delay times in seconds
```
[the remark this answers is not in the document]
diyarajon: Good catch, thanks!

Apps/Src/ReadCarCAN.c
```diff
@@ -3,257 +3,397 @@
...
-#include "Minions.h"
 #include "Minions.h"
+#include "os.h"
 #include "os_cfg_app.h"
-#include "Display.h"
 
-// Saturation threshold is halfway between 0 and max saturation value (half of summation from one to the number of positions)
-#define SATURATION_THRESHOLD (((SAT_BUF_LENGTH + 1) * SAT_BUF_LENGTH) / 4) 
+// Array saturation threshold is halfway between 0 and max saturation value (half of summation from one to the number of positions)
+#define ARRAY_SATURATION_THRESHOLD (((SAT_BUF_LENGTH + 1) * SAT_BUF_LENGTH) / 4)
```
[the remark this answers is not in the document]
diyarajon: Got it.

Apps/Src/ReadCarCAN.c
```diff
@@ -1,259 +1,466 @@
...
+    OS_ERR err;
+    if((arrayIgnitionStatus == true  || motorControllerIgnitionStatus == true  )                    // Ignition is ON
+        && HVArrayMsgSaturation >= ARRAY_SATURATION_THRESHOLD                                       // Saturation Threshold has be met
+        && (Contactors_Get(ARRAY_PRECHARGE_BYPASS_CONTACTOR) == OFF)                                // Array PBC is OFF
+        && (OSTmrStateGet(&arrayBypassPrechargeDlyTimer, &err) != OS_TMR_STATE_RUNNING)){           // and precharge is currently not happening  
+            // Asserts error for OS timer start above if conditional was met
+            assertOSError(OS_READ_CAN_LOC, err);
+            // Wait to make sure precharge is finished and then restart array
+            OSTmrStart(&arrayBypassPrechargeDlyTimer, &err);
+    }
```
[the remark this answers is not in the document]
diyarajon: Thanks, I'll add it. I also realized the comment for the first assertOSError is wrong so I'll fix that, too.

Apps/Src/ReadCarCAN.c
```diff
@@ -1,259 +1,466 @@
...
 */
-void canWatchTimerCallback (void *p_tmr, void *p_arg){
-    chargingDisable();
-}
+ void updatePrechargeContactors(void){
 
+    arrayIgnitionStatus = (!Minions_Read(IGN_1));
+    motorControllerIgnitionStatus = (!Minions_Read(IGN_2));
 
+    if((motorControllerIgnitionStatus == false && arrayIgnitionStatus == false) || (motorControllerIgnitionStatus == true && arrayIgnitionStatus == true)){
```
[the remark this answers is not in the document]
diyarajon: I renamed it to READCARCAN_ERR_DISABLE_CONTACTORS_MSG that the contactors should be turned off if array and motor controller are both off. LV won't be responsible for turning them off unless the disable messages are sent. So Controls should probably turn it off through the assert error. I hope that's what you were asking.

Apps/Src/ReadCarCAN.c
```diff
@@ -1,259 +1,466 @@
...
+// Precharge delay timer variable
+static OS_TMR motorControllerBypassPrechargeDlyTimer;
 
 // NOTE: This should not be written to anywhere other than ReadCarCAN. If the need arises, a mutex to protect it must be added.
 // Indicates whether or not regenerative braking / charging is enabled.
-static bool chargeEnable = false;
+static bool chargeEnable = false; // Enable (High message) of BPS high voltage (HV) array contactor
+
+// BPS HV Array saturation buffer variables
+static int8_t HVArrayChargeMsgBuffer[SAT_BUF_LENGTH];
```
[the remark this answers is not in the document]
diyarajon: Yes, that sounds like a great idea. For clarity purposes, I set the all to disable messages as opposed to 0 so it takes longer for it to meet threshold. The initial time it takes to meet the threshold probably doesn't matter, but just to be safe I made it longer. Thanks for pointing this out!

Apps/Src/ReadCarCAN.c
```diff
@@ -1,259 +1,466 @@
...
 
-        switch(dataBuf.ID){ //we got a message
-            case BPS_TRIP: {
-                // BPS has a fault and we need to enter fault state (probably)
-                if(dataBuf.data[0] == 1){ // If buffer contains 1 for a BPS trip, we should enter a nonrecoverable fault
+        switch(dataBuf.ID){  // Switch case based on BPS msg received
+            case BPS_TRIP: { // BPS has a fault and we need to enter fault state 
 
-                    Display_Evac(SOC, SBPV);    // Display evacuation message
+                Display_Evac(SOC, SBPV); // Display evacuation message
```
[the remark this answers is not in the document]
diyarajon: Thanks! I'll remove it.

Apps/Inc/ReadCarCAN.h
```diff
@@ -14,16 +14,24 @@
...
+/**
+ * Error types
+ */
+typedef enum{
+	READCARCAN_ERR_NONE			    		= 0,
+	READCARCAN_ERR_CHARGE_DISABLE   		=-1,	// Received a charge disable msg
+	READCARCAN_ERR_MISSED_MSG				=-2,	// Didn't receive a BPS charge msg in time
+	READCARCAN_ERR_DISABLE_CONTACTORS_MSG   =-3,		// Ignition is turned to neither (off due to LV) or both at the same time (impossible) are on at 
+	READCARCAN_ERR_BPS_TRIP		    		=-4		// Received a BPS trip msg (0 or 1)
+} ReadCarCAN_error_code_t;
```
NathanielDelgado, line 28: Shouldn't be negative

Tests/Test_Minions.c
NathanielDelgado, line 1: Can take this out I think

NathanielDelgado (dismissed):
seems fine

Apps/Inc/ReadCarCAN.h
diyarajon: fixed now

KnockbackNemo (changes requested):
Looks great! I just wanted to check if a couple things were intentional or not before we merge it in, but otherwise it looks solid to me!

Apps/Src/ReadCarCAN.c
```diff
@@ -5,256 +5,482 @@
...
             }
+
             case SUPPLEMENTAL_VOLTAGE: {
                 SBPV = (*(uint16_t *) &dataBuf.data);
                 UpdateDisplay_SetSBPV(SBPV); // Receive value in mV
                 break;
             }
             case STATE_OF_CHARGE:{
-                SOC = (*(uint32_t*) &dataBuf.data)/(100000);  // Convert to integer percent
+                SOC = (*(uint32_t*) &dataBuf.data)/(1000000);  // Convert to integer percent
```
IshDeshpa, line 378: Make this number a macro and call it SOC_SCALER or something

Apps/Src/ReadCarCAN.c
```diff
@@ -5,256 +5,482 @@
...
+    }
+}
+
+/**
+ * @brief error handler function to display the evac screen if we get a BPS trip message.
+ * Callbacks happen after displaying the fault, so this screen won't get overwritten
+ */ 
+static void handler_ReadCarCAN_BPSTrip(void) {
+    chargeEnable = false; // Not really necesssary but makes inspection less confusing
+	Display_Evac(SOC, SBPV);   // Display evacuation screen
```
IshDeshpa, line 449: Tabbing looks weird on my end. Also 3 s's are in necessary when there should only be 2.

Apps/Src/main.c
```diff
@@ -177,8 +177,6 @@ void Task_Init(void *p_arg){
 
 
     while(1){
-        Contactors_Set(MOTOR_CONTACTOR, Minions_Read(IGN_2), true); //turn on the contactor if the ign switch lets us
-        assertOSError(err);
-        OSTimeDlyHMSM(0, 0, 0, IGN_CONT_PERIOD, OS_OPT_TIME_HMSM_NON_STRICT, &err);
+        CPU_WaitForInt();
```
IshDeshpa, line 180: Why is this here

BSP/STM32F413/Src/BSP_CAN.c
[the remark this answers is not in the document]
IshDeshpa: ^. remove this from PR if possible.

Simulator/stm32f413.repl
```diff
@@ -60,8 +60,16 @@ timer9: Timers.STM32_Timer @ sysbus <0x40014000, +0x400>
...
+    numberOfOutputLines: 24
     [0-9, 13] -> nvic@[6-10, 23, 40, 1, 41, 42, 3]
 
+nvicInput23: Miscellaneous.CombinedInput @ none
+    numberOfInputs: 5
+    -> nvic@23
+
+nvicInput40: Miscellaneous.CombinedInput @ none
+    numberOfInputs: 6
+    -> nvic@40
```
IshDeshpa, line 72: What does this do

Tests/Test_App_Telemetry.c
```diff
@@ -59,7 +59,7 @@ void Task1(void *arg)
 
         // Check contactors (HAVE NOTHING HOOKED UP TO CONTACTORS)
         contactorState = contactorState == OFF ? ON : OFF;
-        Contactors_Set(MOTOR_CONTACTOR, contactorState, true);
+        Contactors_Set(MOTOR_CONTROLLER_BYPASS_PRECHARGE_CONTACTOR, contactorState, true);
         Contactors_Set(ARRAY_CONTACTOR, contactorState, true);
```
IshDeshpa, line 63: Shouldn't this also be renamed?

Tests/Test_Minions.c
```diff
@@ -0,0 +1,14 @@
...
+int main(void){
+    Minion_Init();
+    BSP_UART_Init(UART_2);
+    Minion_Error_t e;
+
+    while(1){
+        printf("IGN_1: %d\n\r", Minion_Read_Pin(IGN_1, &e));
+        printf("IGN_2: %d\n\r", Minion_Read_Pin(IGN_2, &e));
+
+        for(int i =0; i < 999999; i++){}
```
IshDeshpa, line 12: This is a fine test file to use to make sure ignition stuff is working, but I would either rename the test file to be more descriptive of what's actually in it or actually flesh out the test file to test all minions, or remove it from the PR.

Tests/Test_Renode_CANBPSFault.c
```diff
@@ -28,7 +28,7 @@ void Task1(void *p_arg){
     // Enable contactors for ReadCarCAN to flip them
     Contactors_Init();
     Contactors_Enable(ARRAY_CONTACTOR);
```
IshDeshpa, line 30: Shouldn't this be the motor precharge bypass contactor?

IshDeshpa (changes requested):
🦖🦖🦖🦖

Apps/Src/ReadCarCAN.c
```diff
@@ -5,256 +5,482 @@
...
+// Precharge delay timer variable
+static OS_TMR motorControllerBypassPrechargeDlyTimer;
 
 // NOTE: This should not be written to anywhere other than ReadCarCAN. If the need arises, a mutex to protect it must be added.
 // Indicates whether or not regenerative braking / charging is enabled.
-static bool chargeEnable = false;
+static bool chargeEnable = false; // Enable (High message) of BPS high voltage (HV) array contactor
+
+// BPS HV Array saturation buffer variables
+static int8_t HVArrayChargeMsgBuffer[SAT_BUF_LENGTH];
```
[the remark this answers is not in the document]
diyarajon: Struct would be much better

Apps/Src/ReadCarCAN.c
```diff
@@ -5,256 +5,482 @@
...
-// Getter function for chargeEnable
-bool ChargeEnable_Get() 
-{
-    return chargeEnable;
+/**
+ * @brief Callback function for the precharge delay timer. Waits for precharge and then sets arrayBypassPrechargeComplete to true.
+ * @param p_tmr pointer to the timer that calls this function, passed by timer
+ * @param p_arg pointer to the argument passed by timer
+*/
+static void setArrayBypassPrechargeComplete(void *p_tmr, void *p_arg){
```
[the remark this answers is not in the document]
diyarajon: We discussed this, and the reason we cannot have a OSTmrStateGet is because the timer and the array/mc PBC complete boolean do not reflect each other. This is because array/mc PBC complete bool is set to true when precharge timer is complete, but unlike precharge timer, the bool is then set to false when array/mc contactor is on. When precharge happens again, the array/mc PBC complete boolean is set to ON again, if conditions are met.

Tests/Test_Renode_CANBPSFault.c
diyarajon: I think this test can be deleted because it references fault state.

Drivers/Inc/CANbus.h
```diff
@@ -48,8 +41,12 @@ typedef enum {
 	TEMPERATURE 					= 0x24B,
 	ODOMETER_AMPHOURS 				= 0x24E,
 	ARRAY_CONTACTOR_STATE_CHANGE 	= 0x24F,
+	MOTOR_DRIVE 					= 0x221,
```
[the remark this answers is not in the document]
diyarajon: I had changed the order to mirror the CAN LUT but I'll change it to be least to greatest and change it on the LUT as well.

Cam0Cow (changes requested):
Mostly small things, but looks good for the most part.

Apps/Src/ReadCarCAN.c
```diff
@@ -5,256 +5,484 @@
...
-    
 
-    while (1)
-    {
-        
-        //Get any message that BPS Sent us
-        ErrorStatus status = CANbus_Read(&dataBuf,true,CARCAN);  
-        if(status != SUCCESS) {
+    // Fills buffers with disable messages
+    memset(HVArrayChargeMsgBuffer, DISABLE_SATURATION_MSG, sizeof(HVArrayChargeMsgBuffer));
```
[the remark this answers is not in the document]
diyarajon: I'll add that as a comment in case we use a type bigger than a byte.

Apps/Src/ReadCarCAN.c
```diff
@@ -5,256 +5,484 @@
...
+                assertOSError(err); 
+                
+                // Retrieving HV contactor statuses using bit mapping
+                // Bitwise to get HV Plus and Minus, and then &&ing to ensure both are on
+                bool HVPlusMinusStatus = (bool)((dataBuf.data[0] & HV_PLUS_CONTACTOR_BIT) && (dataBuf.data[0] & HV_MINUS_CONTACTOR_BIT));
+                // Bitwise to get HV Array
+                bool HVArrayStatus = (bool)(dataBuf.data[0] & HV_ARRAY_CONTACTOR_BIT);
+
+                // Update HV Array and HV Plus/Minus saturations based on the respective statuses
+                HVArrayStatus ? updateHVArraySaturation(ENABLE_SATURATION_MSG) : disableArrayPrechargeBypassContactor();
```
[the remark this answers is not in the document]
diyarajon: It did not. Should I change it to be an if else instead?

Apps/Src/ReadCarCAN.c
```diff
@@ -5,256 +5,484 @@
...
+#define CAN_WATCH_TMR_DLY_TMR_TS ((CAN_WATCH_TMR_DLY_MS * OS_CFG_TMR_TASK_RATE_HZ) / (1000u)) // 1000 for ms -> s conversion
+
+// Precharge Delay times in milliseconds
+#define PRECHARGE_PLUS_MINUS_DELAY 100u	// 100 ms, as this the smallest time delay that the RTOS can work with
+#define PRECHARGE_ARRAY_DELAY 100u     	// 100 ms
+#define ARRAY_PRECHARGE_BYPASS_DLY_TMR_TS ((PRECHARGE_ARRAY_DELAY * OS_CFG_TMR_TASK_RATE_HZ) / (1000u)) 
+#define MOTOR_CONTROLLER_PRECHARGE_BYPASS_DLY_TMR_TS ((PRECHARGE_PLUS_MINUS_DELAY * OS_CFG_TMR_TASK_RATE_HZ) / (1000u))
+
+// High Voltage BPS Contactor bit mapping
+#define HV_ARRAY_CONTACTOR_BIT 0b001
```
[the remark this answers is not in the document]
diyarajon: I'll change them to be ints and comment the binary just in case

Apps/Src/ReadCarCAN.c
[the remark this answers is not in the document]
Cam0Cow: Eh, if it works it works.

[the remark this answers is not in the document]
diyarajon: 🧙‍♀️

Cam0Cow (changes requested):

Cam0Cow (approved):

Apps/Src/ReadCarCAN.c
```diff
@@ -40,22 +33,25 @@
...
-#define HV_PLUS_CONTACTOR_BIT 0b100
+#define HV_ARRAY_CONTACTOR_BIT 1  //0b001
+#define HV_MINUS_CONTACTOR_BIT 2  //0b010
+#define HV_PLUS_CONTACTOR_BIT  4  //0b100
 
 // Saturation messages
 #define DISABLE_SATURATION_MSG -1
 #define ENABLE_SATURATION_MSG 1
 
+// State of Charge scalar to scale it to correct fix point
```
IshDeshpa, line 44:
```suggestion
// State of Charge scalar to scale it to correct fixed point
```

Apps/Src/ReadCarCAN.c
```diff
@@ -214,30 +213,28 @@ static void updateHVPlusMinusSaturation(int8_t messageState){
...
         }
  }
 
 /**
  * @brief Helper to turn motorControllerPBCOn if motorControllerBypassPrecharge is completed and threshold is reached
  * @param None
 */
- void turnMotorControllerPBCOn(void){
-    if(motorControllerBypassPrechargeComplete == true){
+ void attempTurnMotorControllerPBCOn(void){
```
IshDeshpa, line 236:
```suggestion
 void attemptTurnMotorControllerPBCOn(void){
```

Apps/Src/ReadCarCAN.c
IshDeshpa, line 44:
```suggestion
// State of Charge scalar to scale it to correct fixed point
```

IshDeshpa (commented):
LGTM, just typos.

Cam0Cow (approved):

IshDeshpa (approved):
LGTM.

IshDeshpa (approved):
LGTM

Cam0Cow (approved):

Issue #323, opened 2023-06-17
Keep track of previous task


Cam0Cow:
It's sometimes difficult to know which task triggered fault state, and we currently have no real good way to tell. Modify the OS_AppTaskSwHookPtr to point to a function that will store the task being switched out.

PR #324, opened 2023-06-17
Enabled hooks, defined task switch hook

Cam0Cow:
#323


Cam0Cow:
Tested on hardware, seems to work just fine

Apps/Inc/Tasks.h
```diff
@@ -152,6 +158,20 @@ typedef enum{
...
+/**
+ * Task trace
+ * 
+ * Stores the last 8 tasks that were run
+ * The most recent task is at tasks[index], the one before at tasks[index-1],
+ * wrapping back around at the beginnning
+ */
+typedef struct {
+    OS_TCB *tasks[8];
+    uint32_t index;
```
NathanielDelgado, line 170: Doesn't matter but change to uint8_t to save space.

Apps/Inc/Tasks.h
```diff
@@ -152,6 +158,20 @@ typedef enum{
...
 
+/**
+ * Task trace
+ * 
+ * Stores the last 8 tasks that were run
+ * The most recent task is at tasks[index], the one before at tasks[index-1],
+ * wrapping back around at the beginnning
+ */
+typedef struct {
+    OS_TCB *tasks[8];
```
NathanielDelgado, line 169: Maybe make the 8 a macro?

NathanielDelgado (approved):
Looks good

Apps/Inc/Tasks.h
Cam0Cow: Due to padding it actually wouldn't make a difference.

Apps/Inc/Tasks.h
Cam0Cow: Good idea, I'll do that

IshDeshpa (approved):
Looks good. It may be worth noting that we could just store previous task if we enter a fault state or some built-in fault handler, but assuming that this does not add much overhead to our context switches, this is cool.

NathanielDelgado (approved):
✔

PR #325, opened 2023-06-27
Implement Breathe into documents for doxygen & sphinx integration

NathanielDelgado:
Doxygen has a cool way to document code with comments. Currently we use sphinx to create our documentation. It would be cool if we could use both. That is, write doxygen documentation in our code and be able to leverage that information within the sphinx documentation. 

To do this, I chose Breathe (another Python package) to accomplish this. Basically, you use doxygen to create the xml files and then use Breathe to call those xml files and render them in the rst files sphinx creates. It is simple and does what we need.


NathanielDelgado:
To view the docs, you have to:

1. Install breathe (pip install breathe) and install doxygen (look at doxygen's documentation. Basically have to build it somwhere on the machine)
2. Run doxygen doxyfile in the root directory (creates the xml files used by breathe to render in sphinx)
3. Run make html in Docs/

IshDeshpa:
Leaving this here: https://devblogs.microsoft.com/cppblog/clear-functional-c-documentation-with-sphinx-breathe-doxygen-cmake/

IshDeshpa (approved):
LGTM

Cam0Cow (approved):
LGTM

PR #337, opened 2023-07-08
SendCarCAN Restructure

NathanielDelgado:
Just merging SendCarCAN into Telemetry due to them both just sending data on the CarCAN line.

The only thing I worry about is if sending the motor feedback blocks (CarQueuePost is a blocking call. No non-blocking call available) which stops the other data being sent as well. But this probably doesn't matter much


IshDeshpa:
I would like to handle the restructuring of SendCarCAN (in terms of merging in CANQueue) in this PR as well. I will start work on that.

IshDeshpa:
Also, I'd like to keep the name SendCarCAN instead of Telemetry. It fits more with our naming scheme (SendCarCAN, ReadCarCAN, SendTritium, ReadTritium).

NathanielDelgado (commented):
Make sure you also change the names of the Tests and the tasks in the tests

Apps/Inc/fifo_protected.h
```diff
@@ -0,0 +1,337 @@
...
+    assertOSError(OS_NONE_LOC, err);
+
+    OSMutexPend(&(fifo->FIFO_MUTEX_NAME), 0, OS_OPT_PEND_BLOCKING, &ticks, &err);
+    assertOSError(OS_NONE_LOC, err);
+    
+    if(!IS_EMPTY(fifo)) {
+        *elem = fifo->buffer[fifo->get];
+        fifo->get = (fifo->get + 1) % FIFO_SIZE;
+        success = true;
+    }
```
NathanielDelgado, line 209: Do we need to check if full. Shouldn't the semaphore keep track of the resources inside?

Apps/Inc/fifo_protected.h
```diff
@@ -0,0 +1,337 @@
...
+ * If the type of the fifo is myfifo_t, then this function
+ * will be called myfifo_popback().
+ * 
+ * @param fifo A pointer to the fifo
+ * @param elem A pointer to space to put the element
+ * @return true if successful
+ * @return false if unsuccessful
+ */
+static bool __attribute__((unused))
+POPBACK (FIFO_TYPE_NAME *fifo, FIFO_TYPE *elem) {
```
NathanielDelgado, line 299: Just a prefrence, but I'd like it if it was called just POP instead.

Apps/Src/SendCarCAN.c
```diff
@@ -1,35 +1,89 @@
...
-        CANbus_Send(motorMsg, true, CARCAN);
-
-        // Delay of few milliseconds (100)
-        OSTimeDlyHMSM(0, 0, 0, 100, OS_OPT_TIME_HMSM_STRICT, &err);
-        if (err != OS_ERR_NONE){
-            assertOSError(OS_UPDATE_VEL_LOC, err);
-        }
+        SendCarCAN_Q_get(&CANFifo, &message);
+        CANbus_Send(message, true, CARCAN);
+
```
NathanielDelgado, line 60: Do we want a delay here? I'm guessing it's fine if we always try to send when something gets in the FIFO

NathanielDelgado (commented):
Looks good. Just some questions and prefrences

Apps/Src/SendCarCAN.c
```diff
@@ -1,35 +1,89 @@
...
+    OSTmrCreate(&IOStateTmr,
+                "IO State CAN Message Timer",
+                0,
+                IO_STATE_TMR_DLY_TS,
+                OS_OPT_TMR_PERIODIC,
+                putIOState,
+                NULL,
+                &err
+                );
+}
```
babusid, line 42: Why is this being exported and called elsewhere instead of inside the Task's initialization sequence? Doesn't really make sense for the application not to initialize it's core data structure.

Apps/Inc/fifo_protected.h
babusid: I'm with Nathaniel, popback on a fifo sounds like you're popping the back of it which would be the most recent element (ie. stack type behavior). Conceivably the fifo could be generalized into a deque which could make this more important but thats out of the scope of this pr.

Cam0Cow: Popback makes sense to me, but back and front are confusing when it comes to fifos and stacks. That's why you'll see a lot of fifo implementations call them get and put instead.

IshDeshpa: Popback was the original name in fifo.h 🤣 but I can change it in both if y'all want.

Apps/Src/SendCarCAN.c
IshDeshpa: I figured we don't need a delay, since SendCarCAN_Q does a semaphore pend which will act as a scheduling point anyways.

Apps/Inc/fifo_protected.h
babusid: Maybe just conform to the C++ std queue api def

Apps/Src/SendCarCAN.c
IshDeshpa: I thought we might need it considering that other tasks may be initialized before SendCarCAN and might try to put something in the queue before the semaphore & mutex are initialized. If we initialize all of our tasks before starting the OS though, it's not a problem and I can initialize directly in the task.

IshDeshpa:
Task linked: [CU-8685u78f5 SendCarCAN Restructure](https://app.clickup.com/t/8685u78f5)

IshDeshpa:
Is this ready for review? Also, merge in master.

Apps/Src/SendTritium.c
```diff
@@ -42,6 +45,12 @@
 
 #define GEAR_FAULT_THRESHOLD 3 // number of times gear fault can occur before it is considered a fault
 
+#ifndef SENDTRITIUM_EXPOSE_VARS
+#define SCOPE static
+#else
+#define SCOPE 
+#endif
+
```
NathanielDelgado, line 53: Why is all the old send tritium stuff here?

NathanielDelgado (commented):
### 

> `_**Code seems fine but there is so much old shit in here. I can help you fix this **_`

Apps/Src/SendCarCAN.c
```diff
@@ -1,36 +1,187 @@
...
+
+#define IO_STATE_DLY_MS 250u 
+
+// Task_PutIOState
+OS_TCB putIOState_TCB;
+CPU_STK putIOState_Stk[TASK_SEND_CAR_CAN_STACK_SIZE];
+
+//fifo
+#define FIFO_TYPE CANDATA_t
+#define FIFO_SIZE 256
```
Cam0Cow, line 29: Might want to make this smaller, maybe 16 or 32 elements.

Apps/Src/SendCarCAN.c
```diff
@@ -1,36 +1,187 @@
...
+#define FIFO_NAME SendCarCAN_Q
+#include "fifo.h"
+
+// Motor message
+#define NUM_MOTOR_MSGS 15 // Messages received from the motor controller
```
Cam0Cow, line 34: Might want to leave some comments about how this is used for a sparse array.

Apps/Src/SendCarCAN.c
```diff
@@ -1,36 +1,187 @@
...
+    OS_ERR err;
+    CPU_TS ticks;
+
+    // Reduce the frequency at which we forward motor messages to CarCAN
+    uint8_t msgIdx = message.ID - MOTOR_MSG_BASE_ADDRESS;
+
+    if (msgIdx >= 0 && msgIdx < NUM_MOTOR_MSGS) {  // Check if the message is from the motor controller
+        if (++motorMsgCount[msgIdx] < motorMsgThreshold[msgIdx]) {
+            return;  
+        } else {
```
Cam0Cow, line 71: Don't need an else if you're returning, it can just be

`if (++motorMsgCount[msgIdx] < NUM_MOTOR_MSGS) {  
    return;  
}  
motorMsgCount[msgIdx] = 0;`

Apps/Src/SendCarCAN.c
```diff
@@ -1,36 +1,187 @@
...
+/**
+ * @brief Grabs the latest messages from the queue and sends over CarCAN
+*/
 void Task_SendCarCAN(void *p_arg){
-    CANDATA_t motorMsg;
     OS_ERR err;
+    CPU_TS ticks;
+
+    CANDATA_t message;
+    memset(&message, 0, sizeof(CANDATA_t));
```
Cam0Cow, line 111: A nice way to do this (which is also type-agnostic) is `memset(&message, 0, sizeof message);`

Apps/Src/SendCarCAN.c
```diff
@@ -1,36 +1,187 @@
...
+
+    // PutIOState
+    OSTaskCreate(
+        (OS_TCB*)&putIOState_TCB,
+        (CPU_CHAR*)"PutIOState",
+        (OS_TASK_PTR)Task_PutIOState,
+        (void*)NULL,
+        (OS_PRIO)TASK_SEND_CAR_CAN_PRIO,
```
Cam0Cow, line 119: Maybe add a comment to specify that this is intentional?

Apps/Src/SendCarCAN.c
```diff
@@ -1,36 +1,187 @@
...
+        assertOSError(err);
+
+        if(res) CANbus_Send(message, true, CARCAN);
+        
+    }
+}
+
+static void putIOState(void){
+    CANDATA_t message;
+    memset(&message, 0, sizeof(message));
```
Cam0Cow, line 153: Same here, no need for parentheses for `sizeof` if it's used on a variable name.

Apps/Src/SendCarCAN.c
```diff
@@ -1,36 +1,187 @@
...
+    message.ID = IO_STATE;
+    
+    // Get pedal information
+    message.data[0] = Pedals_Read(ACCELERATOR);
+    message.data[1] = Pedals_Read(BRAKE);
+
+    // Get minion information
+    message.data[2] = 0;
```
Cam0Cow, line 161: Redundant, since you've already memset it to 0.

Apps/Src/SendCarCAN.c
```diff
@@ -1,36 +1,187 @@
...
+    for(pin_t pin = 0; pin < NUM_PINS; pin++){
+        bool pinState = Minions_Read(pin);
+        message.data[2] |= pinState << pin;
+    }
+    
+    // Get contactor info
+    message.data[3] = 0;
```
Cam0Cow, line 168: Same as above.

Apps/Src/common.c
```diff
@@ -9,3 +9,11 @@ void print_float(char * str, float f) {
     int32_t d = (f<0)?-f:f;
     printf("%d.%02d\n\r", (int)n, (int)d);
 }
+
+void print_bin(char * str, uint32_t i){
+    if(str) printf(str);
```
Cam0Cow, line 14: `printf` should always take a string literal, maybe write this as `printf("%s", str);`

Apps/Src/common.c
```diff
@@ -9,3 +9,11 @@ void print_float(char * str, float f) {
...
+
+    for(uint32_t mask=0x80000000L; mask > 0L; mask >>= 1){
+        printf("%d",(mask & i)?1:0);
+    }
```
Cam0Cow, line 18: Should this function print a newline, to match the float printing function?

Docs/source/index.rst
```diff
@@ -37,7 +37,6 @@ Keep the docs clear and readible. Reach out to more experienced team members wit
    Apps/Main
    Apps/ReadCarCAN
    Apps/ReadTritium
-   Apps/SendCarCAN
```
Cam0Cow, line 40: Why is this removed?

arm-gnu-toolchain-11.3.rel1-x86_64-arm-none-eabi.tar.xz?rev=95edb5e17b9d43f28c74ce824f9c6f10&hash=176C4D884DBABB657ADC2AC886C8C095409547C4
Cam0Cow, line 1: This probably shouldn't be here.

Cam0Cow (changes requested):
Didn't review the test file.

Apps/Src/SendCarCAN.c
KnockbackNemo: Got it! I'll see if 16 will work.

Apps/Src/SendCarCAN.c
KnockbackNemo: Yes, thanks for reminding me! Though I was wondering- since the queue isn't actually filling up, do we still want to keep the array? Only half of our messages go out more frequently than 1/sec, and those get sent every 200ms, so I'm not sure how necessary it is to scale them back.

Apps/Src/SendCarCAN.c
KnockbackNemo: Ah very true!

Apps/Src/SendCarCAN.c
KnockbackNemo: Good idea- I'll do that

Apps/Src/common.c
KnockbackNemo: Got it!

Apps/Src/common.c
KnockbackNemo: I think that makes sense, so I'll add \n\r, but @IshDeshpa let me know if that was intentional and I'll change it back to how it is

arm-gnu-toolchain-11.3.rel1-x86_64-arm-none-eabi.tar.xz?rev=95edb5e17b9d43f28c74ce824f9c6f10&hash=176C4D884DBABB657ADC2AC886C8C095409547C4
KnockbackNemo: Yes, thank you for catching that!

Apps/Src/SendCarCAN.c
Cam0Cow: If the queue works fine when scaled down (16-32 elements) then might as well only keep the queue. We can always re-add this if there's bandwidth issues on the CAN line.

KnockbackNemo: Sounds good

Docs/source/index.rst
KnockbackNemo: Whoops, I think that was a mistake. Thanks for catching that!

diyarajon (changes requested):
I like how the code was written -- made it easy to understand.

Apps/Inc/Tasks.h
```diff
@@ -33,7 +33,6 @@
 #define TASK_READ_CAR_CAN_PRIO              5
 #define TASK_UPDATE_DISPLAY_PRIO            6
 #define TASK_SEND_CAR_CAN_PRIO              8
-#define TASK_TELEMETRY_PRIO                 9
 #define TASK_DEBUG_DUMP_PRIO                10
```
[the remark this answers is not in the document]
KnockbackNemo: Yes! Thanks for mentioning that.

Apps/Src/SendCarCAN.c
```diff
@@ -1,36 +1,166 @@
...
- * @param p_arg 
- */
+ * @brief return the space left in SendCarCAN_Q for debug purposes
+*/
+uint8_t get_SendCarCAN_Q_Space(void) {
+    return (CANFifo.get - CANFifo.put - 1) % (sizeof CANFifo.buffer / sizeof CANFifo.buffer[0]);
+}
+
+/**
+ * @brief Wrapper to put new message in the CAN queue
```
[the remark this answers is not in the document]
KnockbackNemo: Yes, good catch!

Apps/Src/SendCarCAN.c
```diff
@@ -1,36 +1,166 @@
...
+    message.ID = IO_STATE;
+    
+    // Get pedal information
+    message.data[0] = Pedals_Read(ACCELERATOR);
+    message.data[1] = Pedals_Read(BRAKE);
+
+    // Get minion information
+    for(pin_t pin = 0; pin < NUM_PINS; pin++){
+        bool pinState = Minions_Read(pin);
+        if(pin == IGN_1 || pin == IGN_2) pinState = !pinState;
```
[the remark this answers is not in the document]
KnockbackNemo: This accounts for the inverted logic of the ignition switch; even though this message is IO State, we decided during integration that sending the logical state would be more useful for BPS than the physical state. This part of the for-loop inverts the recorded state of the ignition pins: if the pin we read is IGN_1 or IGN_2, then we want to negate the state of the pin before adding it to the message data.

Apps/Src/SendCarCAN.c
```diff
@@ -1,36 +1,166 @@
...
+        message.data[2] |= pinState << pin;
+    }
+    
+    // Get contactor info
+    for(contactor_t contactor = 0; contactor < NUM_CONTACTORS; contactor++){
+        bool contactorState = (Contactors_Get(contactor) == ON) ? true : false;
+        message.data[3] |= contactorState << contactor;
+    }
+    // Tell BPS if the array contactor should be on
+    message.data[3] |= (!Minions_Read(IGN_1) || !Minions_Read(IGN_2)) << 2;
```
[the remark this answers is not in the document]
KnockbackNemo: Sorry for the confusion! This is something new we added when integrating with BPS. BPS only cares about the array contactor, so they were only looking at whether or not the array state of the ignition switch was on. We wanted the array contactor to be on whenever we're in array or motor state, but our ignition switch only allows either motor or array to be on at a time (turning on IGN_2 for motor turns off IGN_1 for array).  To ensure the array contactor gets closed in both array and motor state, we made a new data bit (this one) that will tell BPS if the array contactor should be closed (i.e. if either IGN_1 or IGN_2 is true).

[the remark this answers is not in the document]
KnockbackNemo: I forgot to finish updating the CANBusID sheet, though, so thank you for checking that!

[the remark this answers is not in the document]
diyarajon: Thanks! That makes sense. Does this mean that BPS references byte 2 to check for the motor ignition status?

Apps/Src/SendCarCAN.c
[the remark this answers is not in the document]
diyarajon: That makes sense, thanks so much for explaining it!

Apps/Src/SendCarCAN.c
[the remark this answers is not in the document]
KnockbackNemo: As it turns out, BPS doesn't do anything with the motor contactor, so they actually don't look at the motor ignition status at all.

diyarajon (approved):
Looks good to me!

Apps/Inc/SendCarCAN.h
```diff
@@ -0,0 +1,21 @@
...
+
+/**
+ * @brief Wrapper to put new message in the CAN queue
+*/
+void SendCarCAN_Put(CANDATA_t message);
+
+/**
+ * @brief return the space left in SendCarCAN_Q for debug purposes
+*/
+uint8_t get_SendCarCAN_Q_Space(void);
```
IshDeshpa, line 19: Do we need this anymore? If no, can we remove it? If yes, can we surround it with a DEBUG ifdef?

Apps/Inc/fifo.h
```diff
@@ -42,6 +42,8 @@
 #ifndef __FIFO_H
 #define __FIFO_H
 #include <stdbool.h>
+#include <string.h>
+#include "common.h"
```
IshDeshpa, line 46: Why do we need these includes?

IshDeshpa (changes requested):
Just a couple of changes

IshDeshpa (approved):
LGTM!

diyarajon (approved):

Apps/Inc/SendCarCAN.h
diyarajon: I haven't worked/debugged this PR before, but I like the idea of keeping this function just in case 👍

Apps/Src/SendCarCAN.c
```diff
@@ -1,36 +1,166 @@
...
+}
+
+static void putIOState(void){
+    CANDATA_t message;
+    memset(&message, 0, sizeof message);
+    message.ID = IO_STATE;
+    
+    // Get pedal information
+    message.data[0] = Pedals_Read(ACCELERATOR);
+    message.data[1] = Pedals_Read(BRAKE);
```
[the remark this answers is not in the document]
IshDeshpa: Created an issue for it. #379

PR #344, opened 2023-07-16
Rewrote and cleanup Minions

NathanielDelgado:
- Mutex taken out
- Can read from input and output pin
- Took out minion error (unneeded)
- Added doxygen
- Made names better (imo)


NathanielDelgado:
Noticed that SendTritium and Telemtry tests don't have the updated CANbus_Init. @IshDeshpa Let me know if you want me to update that here. Also Test_Ign_Cont seems to be something to be deleted @IshDeshpa Let me know if you want me to delete that here as welll

IshDeshpa:
@NathanielDelgado The tests will be updated in Roie's M2 testing PR anyways, so don't bother here.

Apps/Src/CommandLine.c
```diff
@@ -30,7 +30,7 @@ static bool cmd_Contactors_Set(void);
 
 static bool cmd_Minion_Read_Input(void);
 
-static bool cmd_Minion_Write_Output(void);
+static bool cmd_Minion_write(void);
```
IshDeshpa, line 33: Why lowercase on the write/read? All of our other drivers are uppercase after the driver name. We should be consistent

NathanielDelgado: I think it makes sense for functions to be camelCase with the first letter being undercase. But if it is difference in other drivers then I'll make the change

IshDeshpa (approved):
LGTM.

Drivers/Inc/Minions.h
```diff
@@ -17,47 +28,41 @@
...
 /**
- * @brief Writes given status to a specified output pin. Locks writing to all output pins
+ * @brief Updates the status of a pin
  * 
- * @param pin specific pin to be written to 
- * @param status state of pin (0 or 1)
- * @return false if pin is not an output pin, true if it is
+ * @param pin 
+ * @param status 
+ * @return true is fail (wrote to an input)
```
Cam0Cow, line 61: Any particular reason why this isn't the other way around?

Cam0Cow (approved):
This might also be a good time to figure out if the name of the driver is Minion or Minions, since the filenames and function names don't agree.

Drivers/Inc/Minions.h
NathanielDelgado: If we want to error check the function, it is cleaner to me to write `if (Minion_Read(...)) return error;` than `if (!Minion_Read(...)) return error;`

Just one less character to worry about I guess.

Cam0Cow: We don't do that anywhere else, so it just feels a bit weird. But I guess it's fine

IshDeshpa: I think returning false on fail makes a lot more sense. if (Minion_Read(...)) return error; looks like we return an error on a valid read

IshDeshpa:
Is Minion or Minions more appropriate?

Apps/Src/SendTritium.c
```diff
@@ -225,18 +224,18 @@ static void readInputs(){
...
-    if(Minion_Read_Pin(REGEN_SW, &err) && onePedalCounter < DEBOUNCE_PERIOD){onePedalCounter++;}
+    if(Minions_Read(REGEN_SW, &err) && onePedalCounter < DEBOUNCE_PERIOD){onePedalCounter++;}
     else if(onePedalCounter > 0){onePedalCounter--;}
 
-    if(Minion_Read_Pin(CRUZ_EN, &err) && cruiseEnableCounter < DEBOUNCE_PERIOD){cruiseEnableCounter++;}
+    if(Minions_Read(CRUZ_EN, &err) && cruiseEnableCounter < DEBOUNCE_PERIOD){cruiseEnableCounter++;}
     else if(cruiseEnableCounter > 0){cruiseEnableCounter--;}
 
-    if(Minion_Read_Pin(CRUZ_ST, &err) && cruiseSetCounter < DEBOUNCE_PERIOD){cruiseSetCounter++;}
+    if(Minions_Read(CRUZ_ST, &err) && cruiseSetCounter < DEBOUNCE_PERIOD){cruiseSetCounter++;}
```
IshDeshpa, line 233: Why do we still have an error argument here?

Apps/Src/SendTritium.c
```diff
@@ -225,18 +224,18 @@ static void readInputs(){
...
     else if(cruiseSetCounter > 0){cruiseSetCounter--;}
     
     // Update gears
-    bool forwardSwitch = Minion_Read_Pin(FOR_SW, &err);
-    bool reverseSwitch = Minion_Read_Pin(REV_SW, &err);
+    bool forwardSwitch = Minions_Read(FOR_SW, &err);
+    bool reverseSwitch = Minions_Read(REV_SW, &err);
```
IshDeshpa, line 238: Same here. No error argument necessary.

IshDeshpa (changes requested):
LGTM other than the things I mentioned.

Apps/Src/SendTritium.c
NathanielDelgado: My bad, resolved those mistakes

Cam0Cow (approved):
Works on hardware

IshDeshpa (approved):

PR #345, opened 2023-07-22
Rewrite Test_SendTritium

NathanielDelgado:
- Deprecated TEST_SENDTRITIUM macro
- Moved test specific code to DEBUG macro
- Moved setting of variables to setters dependent on DEBUG macro


NathanielDelgado:
Note: In the future, when Renode is fully setup, SendTritium test can be done through Renode, negating the need for setters, since Renode can directly manipulate the inputs.

Apps/Src/SendTritium.c
```diff
@@ -638,29 +645,25 @@ void Task_SendTritium(void *p_arg){
     state = FSM[NEUTRAL_DRIVE];
     prevState = FSM[NEUTRAL_DRIVE];
 
-    #ifndef __TEST_SENDTRITIUM
-    CANbus_Init(MOTORCAN);
+    CANbus_Init(MOTORCAN, NULL, 0);
```
Cam0Cow, line 648: The initialization shouldn't be done in the task

Apps/Src/SendTritium.c
```diff
@@ -638,29 +645,25 @@ void Task_SendTritium(void *p_arg){
...
         state.stateDecider();    // decide what the next state is
 
         // Drive
-        #ifdef __TEST_SENDTRITIUM
+        #ifdef DEBUG
         dumpInfo();
-        #else
+        #endif
         if(MOTOR_MSG_COUNTER_THRESHOLD == motorMsgCounter){
             memcpy(&driveCmd.data[0], &currentSetpoint, sizeof(float));
```
Cam0Cow, line 668: These should be flipped. No need to change it here (the fix is in M2-testing) but I just want to make sure that the change doesn't disappear during a merge, so keep it in mind if you have to merge in master

Cam0Cow (changes requested):
LGTM other than that

Apps/Src/SendTritium.c
NathanielDelgado: Took it out

Cam0Cow (approved):

Apps/Src/SendTritium.c
```diff
@@ -155,7 +164,7 @@ static float percentToFloat(uint8_t percent){
 /**
  * @brief Dumps info to UART during testing
 */
-#ifdef __TEST_SENDTRITIUM
+#ifdef DEBUG
```
IshDeshpa, line 167: Can we move all this stuff to the test file as well? Since we are creating getters for the variables anyways, we should be able to implement this in the test file.

IshDeshpa (changes requested):

Apps/Inc/common.h
```diff
@@ -28,6 +28,19 @@
...
+ * SETTER is used for creating the function and EXPOSE_SETTER is used for creating the declaration in the header file
+ */
+#define SETTER(type, name) \
+    void set_##name(type val){ \
+        name = val; \
+    } \
+
+#define EXPOSE_SETTER(type, name) \
+    void set_##name(type val); \
+
```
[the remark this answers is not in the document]
NathanielDelgado: Could you define mock & stub macro?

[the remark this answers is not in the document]
NathanielDelgado: Also, I don't know if it makes much sense to macro out setters because you might want to use setters in general. We don't do so now but in general, maybe in the future, a module could us a setter as its normal functionality. But if you're adamant on macro-ing them out, that's fine

Apps/Src/SendTritium.c
NathanielDelgado: These debug functions are called in the FSM sequence (it's in the task's while loop). They are not called in the test file. So we can't move them, since they would do nothing in the test file (and cause warnings). It dumps debug info every FSM transition. I guess the real question is if you want to have a a debugging instrument within the FSM code. I think it is fine as is. I don't think having debugging instruments in our code is a bad thing, as long as we macro them out, such as here

IshDeshpa: I think it's fine for M2 but I'd eventually like to isolate all test/debug instruments to the test files specifically. That's more of a long-term goal though.

IshDeshpa (commented):
Has this been tested on Renode/hardware?

NathanielDelgado:
> Has this been tested on Renode/hardware?

@IshDeshpa Yes, I think Roie did so the Saturday after I made this PR. I think he said it worked fine

IshDeshpa (approved):
LGTM as long as we're sure it's been verified on hardware (@Cam0Cow )

IshDeshpa:
@NathanielDelgado Also, make sure you fix any merge conflicts with master

NathanielDelgado:
@IshDeshpa Merged in master

IshDeshpa (dismissed):
LGTM

Cam0Cow (dismissed):

Cam0Cow (dismissed):
LGTM

Simulator/startup.resc
```diff
@@ -67,27 +67,8 @@ machine LoadPlatformDescription $platform
...
-connector Connect sysbus.can1 carCan
-
-mach create "motor-sim"
-mach set "motor-sim"
-machine LoadPlatformDescription $platform
-sysbus LoadELF $motor_bin
-logLevel 3
-runMacro $show_uart_motor
-machine StartGdbServer 3335
-connector Connect sysbus.can3 motorCan
```
IshDeshpa, line 91: Why remove?

Simulator/stm32f413.repl
```diff
@@ -148,8 +148,8 @@ spi2i2s2: SPI.STM32SPI @ sysbus 0x40003800
 
 i2s2ext: SPI.STM32SPI @ sysbus 0x40003400
 
-iwdg: Timers.STM32F4_IndependentWatchdog @ sysbus 0x40003000
-    frequency: 32000
```
IshDeshpa, line 152: Why comment this out?

IshDeshpa (changes requested):
Good stuff. Multiple macros makes more sense. I like.

Simulator/startup.resc
NathanielDelgado: It's annoying when debugging so I took it out, I can put it back in

Simulator/stm32f413.repl
IshDeshpa: It might be because of the version of renode you're using, although i'm not entirely sure. Try running the install script and uninstalling + reinstalling renode

NathanielDelgado: It is because the version of Renode being run but it's whatever at this point. I uncommented the lines

IshDeshpa (approved):
LGTM!

Cam0Cow (approved):

PR #347, opened 2023-07-28
Embedded Sharepoint integration

IshDeshpa:
Ready for integration. Added option to decide which BSP to compile with, so that we maintain compatibility with our completed BSP on StdPeriph.


Makefile
```diff
@@ -21,6 +21,16 @@ LEADER = controls-leader
...
 CARSIM = car-sim
 
+# TODO: Remove BSP option once HAL is fully integrated
+# --------------------------------------------------------------
+BSP ?= StdPeriph
+BSP_PATH = BSP/STM32F413
+
+ifeq ($(BSP), HAL)
+	BSP_PATH := Embedded-Sharepoint/$(BSP_PATH)
+endif
```
NathanielDelgado, line 31: Maybe add a flag to pass to the make command to switch to the HAL path?

Config/Inc/common.h
NathanielDelgado, line 1: Why move it to the config? common.h is not really a config file and common.c is still in the Apps/Src/ directory, so I think there would be some confusion there

NathanielDelgado (changes requested):
Not too sure about the changes for .github so I'd still like @Cam0Cow to review. But looks mostly fine otherwise.

Makefile
IshDeshpa: The BSP variable is the flag, so if BSP=StdPeriph then it will choose the existing BSP, if BSP=HAL it will choose the Embedded-Sharepoint version of it.

Config/Inc/common.h
IshDeshpa: I didn't even realize I had done this. Good catch.

Makefile
NathanielDelgado: Yeah, but I mean like making a flag to pass in to the make command like 'make -HAL stm32 ...' so  can use the hal without having to edit the variable in the file

babusid:
Haven't looked at this too thoroughly, but are you establishing some sort of guarantee that the HAL and STDperiph versions present the same API surface to the higher layers? If not, over time, this may degrade somewhat. Maybe there's some sort of file marker you can add to throw compiler deprecation warnings for stuff you're changing in HAL but preserving backwards compatibility with in STDPeriph.

Makefile
Cam0Cow: You can always run make BSP=HAL to set the BSP variable to HAL

Cam0Cow (approved):
LGTM, just make the changes Nathaniel pointed out.

NathanielDelgado:
> Haven't looked at this too thoroughly, but are you establishing some sort of guarantee that the HAL and STDperiph versions present the same API surface to the higher layers? If not, over time, this may degrade somewhat. Maybe there's some sort of file marker you can add to throw compiler deprecation warnings for stuff you're changing in HAL but preserving backwards compatibility with in STDPeriph.

@babusid We are trying to make the BSP layer in the Embedded-Sharepoint (uses HAL) compatible with our current codebase. There hasn't been too much development but you can see one of the BSP drivers I wrote in the Embedded-Sharepoint under the branch 'BSP-ADC-integration'.

IshDeshpa:
@NathanielDelgado Are you still planning to use this branch to integrate the embedded-sharepoint, or should I close this PR?

IshDeshpa:
Stale PR.

PR #348, opened 2023-08-05
Motor watchdog

### LINKED ISSUE #307: Motor CAN Message Watchdog
I'm not sure if this is already implemented, but if we don't get a message from the motor within a certain interval, we should build a fault handling routine for that.

Cam0Cow:
Closes #307


Cam0Cow:
No point in merging this until update_faultstate is merged in to master

Apps/Src/ReadTritium.c
```diff
@@ -150,7 +167,7 @@ static void assertTritiumError(tritium_error_code_t motor_err){
 	Error_ReadTritium = (error_code_t)motor_err; // Store error code for inspection info
 	if(motor_err == T_NONE) return; // No error, return
 	
-	if(motor_err != T_HALL_SENSOR_ERR){
+	if(motor_err != T_HALL_SENSOR_ERR && motor_err != T_WATCHDOG_LAST_RESET_ERR){
 		// Assert a nonrecoverable error with no callback function- nonrecoverable will kill the motor and infinite loop
 		assertTaskError(OS_READ_TRITIUM_LOC, Error_ReadTritium, NULL, OPT_LOCK_SCHED, OPT_NONRECOV);
```
KnockbackNemo, line 172: Also, I should have thought of this when I wrote the assert earlier, but do we want to lock the scheduler for the hall sensor and motor watchdog errors?

KnockbackNemo (changes requested):
Just left a couple questions/comments

Apps/Src/ReadTritium.c
```diff
@@ -34,7 +38,11 @@ tritium_error_code_t MotorController_getTritiumError(void){
 }
 
 
-
+// Callback for motor watchdog
+static void motorWatchdog(void *tmr, void *p_arg) {
+    // Attempt to restart 3 times, then fail
+    assertTritiumError(T_WATCHDOG_LAST_RESET_ERR);
```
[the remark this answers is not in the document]
Cam0Cow: Huh I didn't realize that this was a tritium defined error. I think it makes sense to make another enum value for it, but still call it a tritium error. Something like T_MOTOR_WATCHDOG = 1 << 31;

Apps/Src/ReadTritium.c
```diff
@@ -98,19 +114,20 @@ void Task_ReadTritium(void *p_arg){
...
 	}
 }
 
 void MotorController_Restart(void){
 	CANDATA_t resetmsg = {0};
 	resetmsg.ID = MOTOR_RESET;
-	CANbus_Send(resetmsg, true, MOTORCAN);
+
+	// Since this runs with the scheduler locked, we can risk pending
+	// on the CAN mailbox semaphore
```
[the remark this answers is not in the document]
Cam0Cow: That's a good point, it might be worth making a BSP-level call instead to get around that

Apps/Src/ReadTritium.c
```diff
@@ -150,7 +167,7 @@ static void assertTritiumError(tritium_error_code_t motor_err){
 	Error_ReadTritium = (error_code_t)motor_err; // Store error code for inspection info
 	if(motor_err == T_NONE) return; // No error, return
 	
-	if(motor_err != T_HALL_SENSOR_ERR){
+	if(motor_err != T_HALL_SENSOR_ERR && motor_err != T_WATCHDOG_LAST_RESET_ERR){
```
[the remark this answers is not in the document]
Cam0Cow: I think it makes sense to only try to restart the motor a set number of times instead of limiting it by type of event. I'll rename some variables to make that clearer

Apps/Src/ReadTritium.c
Cam0Cow: I don't think we need to. That actually simplifies MotorController_Reset, since we can just use a blocking CANbus_Send to send the reset message

Apps/Src/ReadTritium.c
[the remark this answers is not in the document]
Cam0Cow: Actually never mind, see my other comment

Apps/Inc/ReadTritium.h
```diff
@@ -25,6 +25,7 @@ typedef enum{
     T_DESAT_FAULT_ERR = (1<<7), 
     T_MOTOR_OVER_SPEED_ERR = (1<<8), 
     T_INIT_FAIL = (1<<9), //motor controller fails to restart or initialize
+    T_MOTOR_WATCHDOG_TRIP = (1 << 15), // Defined by us, not by tritium
```
NathanielDelgado, line 28: Could we mention in the comment where these are defined? Like say, "These are defined in the tritium motor doc called xxx"?

Apps/Src/ReadTritium.c
```diff
@@ -145,26 +156,26 @@ static inline void handler_Tritium_HallError(void) {
...
  */
 static void assertTritiumError(tritium_error_code_t motor_err){   
-	static uint8_t hall_fault_cnt = 0; //trip counter, doesn't ever reset
+	static uint8_t motor_fault_cnt = 0; //trip counter, doesn't ever reset
 
 	Error_ReadTritium = (error_code_t)motor_err; // Store error code for inspection info
 	if(motor_err == T_NONE) return; // No error, return
 	
-	if(motor_err != T_HALL_SENSOR_ERR){
+	if(motor_err != T_HALL_SENSOR_ERR && motor_err != T_MOTOR_WATCHDOG_TRIP){
```
NathanielDelgado, line 164: Could it be the case that both happen? If that's the case, should it be `!((motor_err & T_HALL_SENSOR_ERR) || (motor_err & T_MOTOR_WATCHDOG_TRIP))`

NathanielDelgado (changes requested):
Looks good. Just two small changes

Apps/Inc/ReadTritium.h
Cam0Cow: Yeah I can add a comment

Apps/Src/ReadTritium.c
Cam0Cow: Good point, I'll fix that

Apps/Src/ReadTritium.c
```diff
@@ -9,19 +9,33 @@
 #include "CAN_Queue.h"
 #include "CANbus.h"
 #include "UpdateDisplay.h"
+#include "os_cfg_app.h"
 #include <string.h>
 
 //status limit flag masks
-#define MOTOR_RESTART_THRESHOLD 3	// Number of times to restart before asserting a nonrecoverable error
+#define MASK_MOTOR_TEMP_LIMIT (1<<6) //check if motor temperature is limiting the motor 6
```
KnockbackNemo, line 16: Just wondering what the 6 in the comment signifies

KnockbackNemo (approved):
Looks good to me!

Apps/Src/ReadTritium.c
Cam0Cow: Honestly no clue, I'll remove it.

Apps/Inc/Tasks.h
```diff
@@ -173,6 +173,29 @@ void EmergencyContactorOpen();
...
+ * displaying a fault screen (if nonrecoverable), jumping to a callback function, and entering an infinite loop. 
+ * Called by task-specific error-assertion functions that are also responsible for setting the error variable.
+ * @param errorLoc the task from which the error originated. Note: should be taken out when last task pointer is integrated
+ * @param errorCode the enum for the specific error that happened
+ * @param errorCallback a callback function to a handler for that specific error, 
+ * @param schedLock whether or not to lock the scheduler to ensure the error is handled immediately
+ * @param nonrecoverable whether or not to kill the motor, display the fault screen, and enter an infinite while loop
+*/
+void assertTaskError(error_code_t errorCode, callback_t errorCallback, error_scheduler_lock_opt_t lockSched, error_recov_opt_t nonrecoverable);
+
```
IshDeshpa, line 198: Why are these registering as changes when the same thing should be on master?

Apps/Src/Tasks.c
```diff
@@ -146,6 +146,64 @@ void EmergencyContactorOpen() {
...
+ * @param errorCallback a callback function to a handler for that specific error, 
+ * @param schedLock whether or not to lock the scheduler to ensure the error is handled immediately
+ * @param nonrecoverable whether or not to kill the motor, display the fault screen, and enter an infinite while loop
+*/
+void assertTaskError(error_code_t errorCode, callback_t errorCallback, error_scheduler_lock_opt_t lockSched, error_recov_opt_t nonrecoverable) {
+    OS_ERR err;
+
+    if (lockSched == OPT_LOCK_SCHED) { // We want this error to be handled immediately without other tasks being able to interrupt
+        OSSchedLock(&err);
+        assertOSError(err);
```
IshDeshpa, line 164: Why is this stuff marked as different from master?

Tests/Test_BasicCAN.c
```diff
@@ -0,0 +1,57 @@
...
+    TaskSwHook_Init();
+
+    OSTaskCreate(
+        (OS_TCB*)&Task1_TCB,
+        (CPU_CHAR*)"Task1",
+        (OS_TASK_PTR)Task1,
+        (void*)NULL,
+        (OS_PRIO)4,
+        (CPU_STK*)Task1_Stk,
+        (CPU_STK_SIZE)DEFAULT_STACK_SIZE/10,
```
IshDeshpa, line 45: Why stack limit so small?

IshDeshpa (changes requested):
Some stuff to clarify

Apps/Src/ReadTritium.c
```diff
@@ -4,18 +4,23 @@
...
 #include "CANbus.h"
 #include "UpdateDisplay.h"
+#include "os_cfg_app.h"
 #include <string.h>
 
 //status limit flag masks
 #define MASK_MOTOR_TEMP_LIMIT (1<<6) //check if motor temperature is limiting the motor 6
 #define MAX_CAN_LEN 8
 #define RESTART_THRESHOLD 3	// Number of times to restart before asserting a nonrecoverable error
+#define MOTOR_TIMEOUT_SECS 1
```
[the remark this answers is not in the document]
Cam0Cow: Wanted the timeout to mean that we've missed a couple of messages. A second seemed reasonable.

Tests/Test_ReadTritium.c
```diff
@@ -1,94 +1,81 @@
...
-        (OS_MSG_QTY)NULL,
+        (OS_PRIO)TASK_READ_TRITIUM_PRIO,
+        (CPU_STK*)ReadTritium_Stk,
+        (CPU_STK_SIZE)TASK_READ_TRITIUM_STACK_SIZE/10,
+        (CPU_STK_SIZE)TASK_READ_TRITIUM_STACK_SIZE,
+        (OS_MSG_QTY)0,
         (OS_TICK)NULL,
         (void*)NULL,
-        (OS_OPT)(OS_OPT_TASK_STK_CLR),
+        (OS_OPT)(OS_OPT_TASK_STK_CLR|OS_OPT_TASK_STK_CHK),
```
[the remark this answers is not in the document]
Cam0Cow: Not really, we don't actually spawn the stats task that checks this.

Tests/Test_BasicCAN.c
Cam0Cow: That doesn't mean it's small, it actually means that the watermark only gets surpassed if we use more than 90% of the stack. Not that it matters, since we don't actually check the watermark limit anywhere.

Tests/Test_ReadTritium.c
```diff
@@ -1,94 +1,81 @@
...
+        (OS_TASK_PTR)Task_ReadTritium,
         (void*)NULL,
-        (OS_PRIO)1,
-        (CPU_STK*)Task1Stk,
-        (CPU_STK_SIZE)128/10,
-        (CPU_STK_SIZE)128,
-        (OS_MSG_QTY)NULL,
+        (OS_PRIO)TASK_READ_TRITIUM_PRIO,
+        (CPU_STK*)ReadTritium_Stk,
+        (CPU_STK_SIZE)TASK_READ_TRITIUM_STACK_SIZE/10,
```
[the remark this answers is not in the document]
Cam0Cow: Same as below.

Apps/Inc/Tasks.h
Cam0Cow: Messed up the merge on that one.

Apps/Src/Tasks.c
Cam0Cow: Same as above, fixing.

Apps/Src/ReadTritium.c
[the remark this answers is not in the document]
IshDeshpa:
```suggestion
#define MOTOR_TIMEOUT_SECS 1 // Timeout for several missed motor messages
```

IshDeshpa (approved):
LGTM.

NathanielDelgado (approved):
Seems fine

Issue #357, opened 2023-09-26
Main.C Initialization Procedure


IshDeshpa:
**Describe the problem this feature will solve**
Main.c is a little bit outdated as of right now and we are unsure if it spawns all necessary tasks. Even if it does, it may have some integration bugs/issues.

**Describe the proposed solution**
Ensure main.c starts all necessary drivers & applications, and test on hardware to see if there are any visible problems in intended functionality.

**Describe possible alternatives**
Run the Controls codebase on an Arduino with one while loop and no RTOS/BSP/Drivers.

Issue #333, opened 2023-07-06
Get rid of the os_error_loc_t enum


KnockbackNemo:
All the assertOSError functions require a location, but we don't actually do anything with it. The only place the enum is actually useful is in assertTaskError and Display_Error.  However, in assertTaskError it was intended to be replaced by a more proper system/enum later, and in Display_Error it will likely be deleted once we complete the issue ticket requesting changes to the display screen message.

Cam0Cow:
> Additionally, when implementing Display_Error, we can grab the last task from the previous tasks queue and send the display the name of the task where the fault happened. This would require additional work on the display end to support being sent a string and displaying it on the fault screen.

Disregard this, the driver does not care which task the fault came from.

PR #312, opened 2023-04-08
Update faultstate

KnockbackNemo:

PR #361, opened 2023-09-30
Update main.c

Cam0Cow:
# Quality Assurance Checklist 

## Requirements 
- [x] Code Build checks pass 
- [x] No merge conflicts 
- Testing
    - [ ] Tested on hardware
- [x] Did you have fun?

## Things to Consider 


Cam0Cow:
Not done yet, still need to remove fault state when #312 gets merged in, and then error_loc_t needs to be removed whenever #333 gets done.
Closes #357

Apps/Src/main.c
```diff
@@ -139,16 +151,16 @@ void Task_Init(void *p_arg){
...
-        (OS_TCB*)&ReadTritium_TCB,
-        (CPU_CHAR*)"ReadTritium",
-        (OS_TASK_PTR)Task_ReadTritium,
+        (OS_TCB*)&SendCarCAN_TCB,
+        (CPU_CHAR*)"SendCarCAN",
+        (OS_TASK_PTR)Task_SendCarCAN,
         (void*)NULL,
-        (OS_PRIO)TASK_READ_TRITIUM_PRIO,
-        (CPU_STK*)ReadTritium_Stk,
+        (OS_PRIO)TASK_SEND_CAR_CAN_PRIO,
```
diyarajon, line 160: In Tasks.h, the PRIOs are not sequential because 7 is skipped. Is this something we want to fix so that it is incremental?

diyarajon: Also, the last three (telemetry, debug dump) are only being used for tests. I know this PR isn't for Tasks.h, but does this need to be addressed?

diyarajon (changes requested):
IGN_CONT_PERIOD on line 21 is not being used so I believe that it can be deleted.

Apps/Src/main.c
```diff
@@ -175,10 +187,5 @@ void Task_Init(void *p_arg){
     );
     assertOSError(err);
 
-
-    while(1){
-        Contactors_Set(MOTOR_CONTACTOR, Minions_Read(IGN_2), true); //turn on the contactor if the ign switch lets us
-        assertOSError(err);
-        OSTimeDlyHMSM(0, 0, 0, IGN_CONT_PERIOD, OS_OPT_TIME_HMSM_NON_STRICT, &err);
-    }
+    OSTaskDel(NULL, &err);
```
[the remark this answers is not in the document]
IshDeshpa: This will delete the currently running task.
image
If the currently running task is deleted then code after this line will not run.

Apps/Src/main.c
IshDeshpa: Will fix.

Apps/Src/main.c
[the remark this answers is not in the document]
diyarajon: Got it. Documentation had mentioned that err needed to be in a different place, so I had assumed it could be checked inside the task but this makes sense. Thanks, for clarifying this!

diyarajon (approved):

NathanielDelgado (dismissed):
Seems fine. Just don't know if we want CommandLine as default in main.c

IshDeshpa:
This is the latest working version from testing on 12/13 with @NathanielDelgado. I merged in master which may have changed some things but I checked to make sure no behavior was changed so it should still function the same.

NathanielDelgado (approved):
lgtm

IshDeshpa (approved):
LGTM

Tests/Test_TrailingZeros.c
```diff
@@ -0,0 +1,34 @@
+/*
+ * C Program to Count the Number of Trailing Zeroes in Integer
+ */
+#include "common.h"
+#include "config.h"
+#include <bsp.h>
+ 
+int main()
```
babusid, line 8: why do we have this

Drivers/Src/CANbus.c
```diff
@@ -15,8 +15,6 @@ static OS_SEM CANBus_ReceiveSem4[NUM_CAN]; // sem4 to count how many msgs in our
 static OS_MUTEX CANbus_TxMutex[NUM_CAN];   // mutex to lock tx line
 static OS_MUTEX CANbus_RxMutex[NUM_CAN];   // mutex to lock Rx line
```
babusid, line 16: a little bit of spacing is nice but whatevs

Drivers/Inc/CANbus.h
```diff
@@ -29,7 +29,7 @@
  */
 typedef enum { 
 	BPS_TRIP						= 0x002,
-	BPS_CONTACTOR					= 0x101,
+	BPS_CONTACTOR					= 0x102,
```
babusid, line 32: why did this can id change / is it documented in the sheet properly

Drivers/Inc/CANConfig.h
```diff
@@ -15,7 +15,7 @@
  * Filter Lists for CarCAN and MotorCAN
 */
 #define NUM_CARCAN_FILTERS 6
-#define NUM_MOTORCAN_FILTERS 0
+#define NUM_MOTORCAN_FILTERS 7
```
babusid, line 18: im guessing this is just the number of IDs to allow in right?

BSP/STM32F413/Makefile
```diff
@@ -77,7 +77,7 @@ SF = st-flash
 CPU = -mcpu=cortex-m4
 
 # float-abi
-FLOAT-ABI = -mfloat-abi=soft
+FLOAT-ABI = -mfloat-abi=hard
```
babusid, line 80: has hardware float been proven to work on this branch? Also, is the default not hardware float, do we actually even need this flag anymore?

Apps/Inc/SendTritium.h
```diff
@@ -12,7 +12,7 @@
 
 #include "common.h"
 
-#define SENDTRITIUM_PRINT_MES
+//#define SENDTRITIUM_PRINT_MES
```
babusid, line 15: just commenting this out is kinda inelegant imo. Could we make this a makefile flag maybe? That way we don't have to change code when we want this, we could just build a different binary.

Apps/Inc/Tasks.h
```diff
@@ -31,10 +31,11 @@
 #define TASK_READ_TRITIUM_PRIO              3
 #define TASK_SEND_TRITIUM_PRIO              4
 #define TASK_READ_CAR_CAN_PRIO              5
-#define TASK_UPDATE_DISPLAY_PRIO            6
-#define TASK_SEND_CAR_CAN_PRIO              7
-#define TASK_DEBUG_DUMP_PRIO                8
-#define TASK_COMMAND_LINE_PRIO              9
+#define TASK_SEND_CAR_CAN_PRIO              6
+#define TASK_PUT_IOSTATE_PRIO               7
+#define TASK_UPDATE_DISPLAY_PRIO            8
```
babusid, line 36: I kind of think updating the display is more important than sending controls telemetry, but sendcan is needed for ignition (?).

I think maybe it should be sendcan=6,display=7,iostate=8 but it likely doesn't really matter, this is kinda nitpicky

Apps/Inc/fifo.h
```diff
@@ -42,6 +42,8 @@
 #ifndef __FIFO_H
 #define __FIFO_H
 #include <stdbool.h>
+#include <string.h>
```
babusid, line 45: are these necessary? Fifo.h has been working for a long time. again, not rly a problem, just curious as to why

Apps/Src/ReadTritium.c
```diff
@@ -36,41 +37,27 @@ static void motorWatchdog(void *tmr, void *p_arg) {
...
-	OSTmrCreate(&MotorWatchdog, "Motor watchdog", MOTOR_TIMEOUT_TICKS, MOTOR_TIMEOUT_TICKS, OS_OPT_TMR_PERIODIC, motorWatchdog, NULL, &err);
-	assertOSError(err);
-	OSTmrStart(&MotorWatchdog, &err);
-	assertOSError(err);
+	static bool watchdogCreated = false;
 
 	while (1){
 		ErrorStatus status = CANbus_Read(&dataBuf, true, MOTORCAN);
 
 		if (status == SUCCESS){
```
babusid, line 50: invert this to be status != SUCCESS and hit an early continue. I think i proposed this in one of my refactor PRs.

babusid:
```
if(status!=SUCCESS) continue;
```

Apps/Src/SendCarCAN.c
```diff
@@ -48,19 +51,27 @@ uint8_t get_SendCarCAN_Q_Space(void) {
...
     CPU_TS ticks;
-
-    OSMutexPend(&CarCAN_Mtx, 0, OS_OPT_PEND_BLOCKING, &ticks, &err);
-    assertOSError(err);
+    bool success = false;
+    
+    static uint8_t carcan_ctr = 0;
     
-    bool success = SendCarCAN_Q_put(&CANFifo, message);
+    if(carcan_ctr > 3){
```
babusid, line 62: wait, does this not just put every 3d message onto the bus? What is the point of this?

babusid: Okay, in order to solve Q saturation issues, the solution is not to just skip 2/3 of all the incoming queue pushes. See if you can throttle the input rate on the sending applications, and more finely control what actually gets skipped. IMO, IOstate data is super low priority, and so is most of the motor's stuff. Add delays / filtering there, and see whether you can handle the traffic.

Apps/Src/SendTritium.c
```diff
@@ -32,7 +32,7 @@
 #define MIN_CRUISE_VELOCITY mpsToRpm(20.0f) // rpm
 #define MAX_GEARSWITCH_VELOCITY mpsToRpm(8.0f) // rpm
 
-#define BRAKE_PEDAL_THRESHOLD 15  // percent
+#define BRAKE_PEDAL_THRESHOLD 50  // percent
```
babusid, line 35: what do these macros mean? None of them have descriptions

Apps/Src/Tasks.c
```diff
@@ -103,17 +105,17 @@ void throwTaskError(error_code_t errorCode, callback_t errorCallback, error_sche
 
             #if DEBUG == 1
                // Print the error that caused this fault
-                printf("\n\rCurrent Error Code: 0x%04x\n\r", errorCode);
+                // printf("\n\rCurrent Error Code: 0x%04x\n\r", errorCode);
```
babusid, line 108: why not just delete this

babusid (changes requested):

babusid (dismissed):
Mostly looks good, but the queue saturation fix in SCC is kind of a patch job, and while it practically likely is stable, theoretically the messages it skips is completely arbitrary. I would investigate an approach with stronger guarantees.

Drivers/Inc/CANConfig.h
IshDeshpa: yup. honestly we could use sizeof for the array instead of this macro.

BSP/STM32F413/Makefile
IshDeshpa: hardware floats aren't still technically being used in the first place since we still use PedalToPercent in SendTritium anyways. but enabling this hasn't caused us any issues thus far.

Apps/Inc/SendTritium.h
IshDeshpa: Just created #384 to address this. Not important for MVP, will be addressed in M2.0.1

Apps/Inc/Tasks.h
IshDeshpa: I see what you mean. Even though ignition sequence hinges upon IOState, it's not as important once the car has started up.

Apps/Inc/fifo.h
IshDeshpa: I thought I removed this. Will fix.

Apps/Src/ReadTritium.c
IshDeshpa: We'll address this in the refactor PR.

diyarajon (changes requested):
Just a couple of comments and questions on the code

Apps/Src/SendCarCAN.c
```diff
@@ -26,7 +29,7 @@ CPU_STK putIOState_Stk[TASK_SEND_CAR_CAN_STACK_SIZE];
 
 //fifo
 #define FIFO_TYPE CANDATA_t
-#define FIFO_SIZE 16
+#define FIFO_SIZE 50
```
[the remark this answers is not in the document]
NathanielDelgado: Just so we can handle bigger bursts of data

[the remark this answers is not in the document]
NathanielDelgado: This might change once integrating with BPS

Apps/Src/SendCarCAN.c
NathanielDelgado: We did this soley for the purposes of testing to see if everything works without needing to set up the software the configure the motor controller to give us less messages less frequently. The counters will be removed and addressed in #386

Apps/Src/SendCarCAN.c
```diff
@@ -97,7 +108,7 @@ void Task_SendCarCAN(void *p_arg){
         (CPU_CHAR*)"PutIOState",
         (OS_TASK_PTR)Task_PutIOState,
         (void*)NULL,
-        (OS_PRIO)TASK_SEND_CAR_CAN_PRIO, // Round-robin with SendCarCAN task
+        (OS_PRIO)TASK_PUT_IOSTATE_PRIO,
```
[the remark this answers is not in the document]
NathanielDelgado: This was done before being tested on hardware. I believe this was done while debugging to see if there was issues with the priorities and it ended up staying like that in what we tested as working. I suppose we could go back but this is also just fine.

Apps/Src/SendCarCAN.c
```diff
@@ -141,7 +150,6 @@ static void putIOState(void){
     // Get minion information
     for(pin_t pin = 0; pin < NUM_PINS; pin++){
         bool pinState = Minions_Read(pin);
-        if(pin == IGN_1 || pin == IGN_2) pinState = !pinState;
```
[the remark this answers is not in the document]
NathanielDelgado: @IshDeshpa Do you know? I guess it doesn't matter much, as long as Data Acq knows the format

Apps/Src/SendTritium.c
NathanielDelgado: They handle what threshold do we consider the reading of the pedal to be to be considered on. This documentation should happen when we go through and document the code more thoroughly

Apps/Src/SendTritium.c
```diff
@@ -668,6 +665,8 @@ void Task_SendTritium(void *p_arg){
         #endif
         state.stateDecider();    // decide what the next state is
 
+        velocitySetpoint = (velocitySetpoint>0)?MAX_VELOCITY:-MAX_VELOCITY;
```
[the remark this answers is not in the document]
NathanielDelgado: Should be handled in the formatting PR

Apps/Src/Tasks.c
NathanielDelgado: I suppose we might want to maybe enable it later when debugging in the future. I would consider this to be apart of the code cleanup PR

BSP/STM32F413/Makefile
NathanielDelgado: I think it is fine to merge it in as is because we don't use float. I do think this was part of the fix with correcting floats

Drivers/Inc/CANConfig.h
NathanielDelgado: I think we can merge it as is and address it later. #387

Drivers/Inc/CANbus.h
NathanielDelgado: It is updated in the sheet. @IshDeshpa Why did this change?

Tests/Test_TrailingZeros.c
NathanielDelgado: @IshDeshpa ?

Apps/Src/SendCarCAN.c
NathanielDelgado: Also, might redesign this in this issue ticket #388

Apps/Src/SendCarCAN.c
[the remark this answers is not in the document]
NathanielDelgado: Nevermind. Round robin is disabled. maybe. This works and it seems fine so I don't care right now

Apps/Src/SendCarCAN.c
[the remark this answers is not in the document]
NathanielDelgado: Nevermind. THis should not have been deleted but this issue should be handled in #374

Drivers/Inc/CANbus.h
NathanielDelgado: software typo. It is correct

Apps/Src/SendCarCAN.c
diyarajon, line 62: This isn't too important, but maybe a macro could be used instead of 3

Apps/Src/main.c
```diff
@@ -56,26 +54,18 @@ int main(void) {
...
-    assertOSError(err);
     
     // Initialize drivers
     Pedals_Init();
-    OSTimeDlyHMSM(0,0,5,0,OS_OPT_TIME_HMSM_STRICT,&err);
-    OSTimeDlyHMSM(0,0,10,0,OS_OPT_TIME_HMSM_STRICT,&err);
     BSP_UART_Init(UART_2);
-    CANbus_Init(CARCAN, (CANId_t*)carCANFilterList, NUM_CARCAN_FILTERS);
+    CANbus_Init(CARCAN, carCANFilterList, NUM_CARCAN_FILTERS);
     CANbus_Init(MOTORCAN, NULL, NUM_MOTORCAN_FILTERS);
```
diyarajon, line 69: Are we using `motorCANFilterList` here?

Drivers/Inc/CANConfig.h
```diff
@@ -15,7 +15,7 @@
  * Filter Lists for CarCAN and MotorCAN
 */
 #define NUM_CARCAN_FILTERS 6
```
diyarajon, line 17: If this is the number of allowed IDs, is this 4 now?

diyarajon (changes requested):
Thanks for explaining the code from my questions last time. I left some suggestions on comments and can filter list. Very very cool cod🐟

KnockbackNemo (changes requested):
I left a few questions, but otherwise it looks pretty good! 
I also wanted to note that since we're leaving some comments to be fixed in other issue tickets ([#387](https://github.com/lhr-solar/Controls/issues/387), [#374 ](https://github.com/lhr-solar/Controls/issues/374), [#386](https://github.com/lhr-solar/Controls/issues/386), [#388 ](https://github.com/lhr-solar/Controls/issues/388)), we probably want to make sure the important ones are addressed for the car (or integration/when needed).

IshDeshpa:
Gonna start addressing these changes.

Apps/Src/SendTritium.c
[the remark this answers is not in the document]
IshDeshpa: I'll add a comment to explain it, but it basically disables regen. The velocity setpoint is always set to the maximum (unreachable) value in whichever direction we are traveling, which ensures that we don't ever enter velocity controlled mode (and stay in current controlled mode) while still being allowed to test regen/cruise mode transition logic since the transition itself isn't explicitly disabled.

Apps/Src/Tasks.c
```diff
@@ -54,12 +54,13 @@ extern const pinInfo_t PININFO_LUT[]; // For GPIO writes. Externed from Minions
  * Error assertion-related functions
 */
 
-void _assertOSError(OS_ERR err)
+void _assertOSError(volatile OS_ERR err)
```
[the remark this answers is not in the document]
IshDeshpa: We made this volatile when we were testing a specific OS fault related bug. At a certain section in ReadTritium, there was an extra assertOSError left in that at certain points was being called on an uninitialized OS_ERR variable. This means it was essentially calling assertOSError on garbage data (since uninitialized variables become whatever used to be on the stack). To solve this issue, we printed out the error value in this function through GDB first, and to do so, I had to make this volatile so that GCC didn't optimize the value out and so we could see it in GDB.

However, now that debugging is done, I'll remove this volatile as well as the print statement below.

Apps/Src/Tasks.c
IshDeshpa: Can we make an issue ticket to make this conditional on a separate macro like PRINT_ERROR or something?

Apps/Src/main.c
IshDeshpa: We disabled motorCANFilterList during testing and re-enabling it now would probably mean we'd have to re-test. I'll make an issue ticket for re-testing this functionality on main when we get back.

Drivers/Inc/CANConfig.h
IshDeshpa: Yes, good catch. I will change this.

Apps/Src/SendCarCAN.c
IshDeshpa: Yes, will create a macro for this.

Apps/Src/SendTritium.c
```diff
@@ -62,13 +62,10 @@ float velocitySetpoint = 0;
 float cruiseVelSetpoint = 0;
 
 // Current observed velocity
-
```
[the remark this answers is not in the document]
IshDeshpa: Can you make an issue ticket for this so that we can solve it in another PR?

Apps/Src/main.c
IshDeshpa: #397

Drivers/Inc/CANConfig.h
IshDeshpa: This will also be addressed in #387.

Apps/Src/SendTritium.c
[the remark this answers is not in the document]
KnockbackNemo: Got it- this issue will be resolved in [#398](https://github.com/lhr-solar/Controls/issues/398)

Apps/Src/SendTritium.c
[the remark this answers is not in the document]
KnockbackNemo: Ah, thanks for explaining it!

Apps/Src/SendCarCAN.c
```diff
@@ -48,19 +50,27 @@ uint8_t get_SendCarCAN_Q_Space(void) {
 
 /**
  * @brief Wrapper to put new message in the CAN queue
- * @param message the CAN message to put in SendCarCAN Queue
 */
 void SendCarCAN_Put(CANDATA_t message){
-    OS_ERR err;
+    OS_ERR err = OS_ERR_NONE;
```
KnockbackNemo, line 55:
```suggestion
    OS_ERR err;
```
I'm not sure which way is better, but we could delete this and put the assertOSError into the conditional (so it isn't asserted if nothing happens) to be more consistent with other parts of the codebase.

Apps/Src/SendCarCAN.c
```diff
@@ -48,19 +50,27 @@ uint8_t get_SendCarCAN_Q_Space(void) {
...
+
+        carcan_ctr = 0;
+    }
+    carcan_ctr++;
 
-    OSMutexPost(&CarCAN_Mtx, OS_OPT_POST_NONE, &err);
-    assertOSError(err);
 
     if(success) OSSemPost(&CarCAN_Sem4, OS_OPT_POST_1, &err);
     assertOSError(err);
```
KnockbackNemo, line 76:
```suggestion
    if(success) {
        OSSemPost(&CarCAN_Sem4, OS_OPT_POST_1, &err);
        assertOSError(err);
    }
```

KnockbackNemo (approved):
I left a suggestion, but it's more related to preference and doesn't need to be followed if you disagree. Otherwise, looks good to me!

Drivers/Src/CANConfig.c
```diff
@@ -35,6 +35,7 @@ const CANLUT_T CANLUT[MAX_CAN_ID] = {
 	[TEMPERATURE] 					= {NOIDX, DOUBLE}, /**     TEMPERATURE                     **/
 	[ODOMETER_AMPHOURS] 			= {NOIDX, DOUBLE}, /**     ODOMETER_AMPHOURS               **/
 	[ARRAY_CONTACTOR_STATE_CHANGE] 	= {NOIDX, BYTE  }, /**     ARRAY_CONTACTOR_STATE_CHANGE    **/
+    [SLIP_SPEED]                    = {NOIDX, DOUBLE},
```
diyarajon, line 38: Will we eventually need to write this into ReadTritium?

diyarajon (approved):
Looks good! I left some questions.

Apps/Src/main.c
```diff
@@ -56,26 +54,18 @@ int main(void) {
...
     while(1);
-
-    return 0;
 }
 
 void Task_Init(void *p_arg){
     OS_ERR err;
 
     // Start systick    
     OS_CPU_SysTickInit(SystemCoreClock / (CPU_INT32U) OSCfg_TickRate_Hz);
```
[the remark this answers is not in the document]
IshDeshpa: I think it might be some sort of documentation interpretation error on VSCode's part, but OS_CPU_SysTickInit definitely takes arguments. Here's the source code from `os_cpu_c.c`:
```
/*
*********************************************************************************************************
*                                         INITIALIZE SYS TICK
*
* Description: Initialize the SysTick.
*
* Arguments  : cnts         Number of SysTick counts between two OS tick interrupts.
*
* Note(s)    : 1) This function MUST be called after OSStart() & after processor initialization.
*********************************************************************************************************
*/

void  OS_CPU_SysTickInit (CPU_INT32U  cnts)
```

Seeing as the argument is the number of SysTick counts between two OS ticks, we can see that SystemCoreClock (frequency of the system's clock, or number of ticks per second) divided by the OS Tick Rate in Hz (frequency of the OS, or number of ticks per second) gives us the number of system ticks between two OS ticks.

Drivers/Src/CANConfig.c
IshDeshpa: ReadTritium reads in every message that the motor sends us, regardless of what it is. The point of the switch case in ReadTritium is that sometimes we may need to do some conditional logic if the message is a certain thing (such as outputting battery SOC to the display). Every message we get is forwarded to SendCarCAN, including Slip Speed.

The CANbus driver uses this LUT when calling CANbus_Read or CANbus_Send, so we had to add an entry for SLIP_SPEED to the lookup table.

Apps/Src/main.c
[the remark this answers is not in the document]
IshDeshpa: Also a tip; in VSCode, if you hit Ctrl-T, you can do a symbol lookup where you can type in any variable name or function name and VSCode will try to find where the definition is in your codebase. That's how I found this function.

(you can also use Ctrl-Shift-F, but that'll find any occurrence across your codebase, not necessarily take you to the definition)

Apps/Src/Tasks.c
NathanielDelgado: Yes

NathanielDelgado (approved):
lgtm

KnockbackNemo (approved):
Looks good!

Apps/Src/main.c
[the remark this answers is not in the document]
diyarajon: Thanks so much for explaining this!

Drivers/Src/CANConfig.c
diyarajon: Got it, thanks!

PR #368, opened 2023-11-11
Canloopback macro

loocyx:
# Quality Assurance Checklist 

## Requirements 
- [ ] Code Build checks pass 
- [x] No merge conflicts 
- Testing
    - [ ] Tested on hardware
- [x] Did you have fun?

## Things to Consider 


IshDeshpa:
Will need to merge master into this

Tests/Test_BSP_CAN_loopback.c
KnockbackNemo, line 1: It might be nice to add a short comment at the top of the file that briefly describes how we should use the test. For instance, it looks like this test is meant to be stepped-through with a debugger like GDB to see that the CAN Mode is being set correctly? Another tip for the comment could be that we need to run "make clean" whenever we change the setting to ensure the code gets re-compiled with the right macro option (at least I did, but if you didn't have to, then you can exclude this part)

KnockbackNemo (changes requested):
Awesome job on your first issue ticket! I left a couple general questions, but it looks pretty good to me :)

BSP/STM32F413/Src/BSP_CAN.c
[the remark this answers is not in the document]
IshDeshpa: Good point. Maybe LOOPBACK could have a value option of Car or Motor or Both, so that we could choose which loopback to enable

BSP/STM32F413/Src/BSP_CAN.c
```diff
@@ -94,7 +94,11 @@ void BSP_CAN1_Init(uint16_t* idWhitelist, uint8_t idWhitelistSize) {
     CAN_InitStruct.CAN_NART = DISABLE;
     CAN_InitStruct.CAN_RFLM = DISABLE;
     CAN_InitStruct.CAN_TXFP = ENABLE;
+    #ifdef LOOPBACK
```
diyarajon, line 97: I think that it would be helpful to add a comment here too, especially after adding the MotorCAN and CarCAN values as Maddie and Ishan have suggested.

diyarajon (changes requested):
Great work

IshDeshpa (approved):
LGTM

NathanielDelgado (approved):
lgtm

PR #419, opened 2024-02-29
PedalsRead returning signed int

### LINKED ISSUE #379: Pedals_Read returns a signed int
**Describe the bug**
Pedals_Read produces a number from 0 to 100 but returns a signed integer. This creates the possibility of a bug. Return type should be uint8_t.

diyarajon:
# Quality Assurance Checklist 
## Requirements 
- [x] Code Build checks pass 
- [x] No merge conflicts 
- Testing
    - [ ] Tested on hardware
- [ ] Did you have fun?

## Things to Consider 


Drivers/Src/Pedals.c
```diff
@@ -37,26 +37,28 @@ void PedalsInit() { BspAdcInit(); }
...
-        (int16_t)BspAdcGetMillivoltage((pedal == kAccelerator) ? kCh10 : kCh11);
-
-    int8_t percentage = 0;
 
-    if (millivolts_pedal >= kLowerBound[pedal]) {
-        percentage = (int8_t)((int32_t)(millivolts_pedal - kLowerBound[pedal]) *
-                              100 / (kUpperBound[pedal] - kLowerBound[pedal]));
-    }
+    int16_t millivolts_pedal =
+        BspAdcGetMillivoltage((pedal == kAccelerator) ? kCh10 : kCh11);
```
IshDeshpa, line 46: Can you also make this function return an unsigned int? Take a look at BSP_ADC.c

IshDeshpa (changes requested):
to test this I would just make sure that when the simboard is plugged in, both BSP_ADC and Pedals_Read return the right value. Write a test file that does this.

Drivers/Src/Pedals.c
diyarajon: Sounds good - it looks like both the getters should be using unsigned ints so I'll change them both.

IshDeshpa (dismissed):
LGTM. @KnockbackNemo can you review?

KnockbackNemo (dismissed):
Looks great to me!

diyarajon:
outdated.

Issue #353, opened 2023-09-22
Limit motor speed in reverse


KnockbackNemo:
Currently there's no software limit to how fast we can go in reverse, but we might want to add one.

This is a proposed solution to the fact that there's no velocity check when going from brake state to forward gear. This means if we go 40 MPH in reverse, move the switch to forward, and then brake, then upon release of the brake, pushing the accelerator pedal results in us (attempting) to go forward.

PR #430, opened 2025-01-18
Send tritium rework

OscarPortillo37:
NOTES: 
- Still need to modify it to build
- Need to add some gear fault functionality (& some gear display issues)
- Need to add deadband for hysteresis (for brake & accel pedal press in cruise)
- Will need to look over SendCarCAN
- Later on, probably, but want to add more dynamic current drawing functionality (e.g. if low on battery, draw less; etc)

# Quality Assurance Checklist 

## Requirements 
- [x] Code Build checks pass 
- [x] No merge conflicts 
- [x] Software feature has documentation for it updated in both ReadTheDocs and the comments
- [x] Software feature has documentation for it updated
- Testing
    - [ ] Tested on hardware
- [x] Did you have fun?

## Things to Consider 

https://drive.google.com/file/d/1_yEBhHfJ_CH7eZtytHqsMtaKGBlM4xWr/view?usp=sharing


jiwoop:
**Overall**
- Nice job integrating brake state into handler functions. Though, for turning brake light on/off, that could be a separate small helper function as it is used in every handlers.
- What do the caller functions (i.e. callForwardDriverHandler()) ultimately do? Currently it seems like it's calling handlers and deciders, could we just call them directly?
- Conditionals in ParkDecider() checks for the negative velocity value (velocityObserved), what was the rationale?
- The code follows the state machine coherently

**Code**
- FSM array size should be modified depending on which file (i.e. MVP should have FSM[3] instead of FSM[9])

diyarajon (changes requested):
Nice work! I left some comments.
Could you update the FSM diagram saying that there's hysteresis and pressed/unpressed threshold implemented?

Apps/Src/SendTritium.c
```diff
@@ -254,36 +236,55 @@ static void readInputs()
...
+        if(accelPedalPercent <= ACCEL_PEDAL_UNPRESSED_THRESHOLD)
+            accelSaturationCt--;
+        
+        accelPressed = true;
+    }
 
-    // if(Minions_Read(CRUZ_EN) && cruiseEnableCounter < DEBOUNCE_PERIOD){cruiseEnableCounter++;}
-    // else if(cruiseEnableCounter > 0){cruiseEnableCounter--;}
+    // Protect from both brake and accel being pressed
+    if (brakePedalPercent == 100 && accelPedalPercent == 100) accelPedalPercent = 0;
```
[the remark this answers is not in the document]
OscarPortillo37: Thanks for catching that. Used 25% as the threshold for accel to be set to 0. I figured I'd be a bit more conservative just in case, but let me know if you think that's fine or if it should be higher/lower.

Apps/Src/SendTritium.c
```diff
@@ -379,200 +386,158 @@ static uint8_t map(uint8_t input, uint8_t in_min, uint8_t in_max, uint8_t out_mi
...
-static void NeutralDriveDecider()
+void ParkDecider()
 {
-    if (brakePedalPercent >= BRAKE_PEDAL_THRESHOLD)
-    {
-        state = FSM[BRAKE_STATE];
-    }
-    else if (gear == FORWARD_GEAR)
+    // Switch for FORWARD_DRIVE/REVERSE_DRIVE if in FORWARD_GEAR/REVERSE_GEAR & speed is less than MAX_GEARSWITCH_VELOCITY
+    if (gear == FORWARD_GEAR && (velocityObserved < 0 ? (velocityObserved >= -MAX_GEARSWITCH_VELOCITY) : (velocityObserved <= MAX_GEARSWITCH_VELOCITY)))
```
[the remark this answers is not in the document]
OscarPortillo37: Changed this per Maddie's suggestion to only check if we exceed the gearswitch velocity in the direction opposite to the direction we want to switch to since we don't have to worry about gearswitch if we're changing from, say, forward -> park -> forward.

Apps/Src/SendTritium.c
```diff
@@ -636,157 +599,44 @@ static void CoastingCruiseDecider()
...
+    }
 }
 
 /**
  * @brief Accelerate Cruise State Decider. Determines transitions out of accelerate
  * cruise state (brake, neutral drive, one pedal, forward drive, record velocity,
  * coasting cruise).
  */
-static void AccelerateCruiseDecider()
+void AccelerateCruiseDecider()
```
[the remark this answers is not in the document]
OscarPortillo37: Talked to Parthiv, and he kind of likes the existing implementation - perhaps we just stick with this for now? I can change it if anything later if we A) go to comp for certain & B) the decided driver doesn't like it.

Apps/Src/SendTritium.c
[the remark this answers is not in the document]
diyarajon: That sounds good.

Apps/Src/SendTritium.c
```diff
@@ -1,84 +1,72 @@
...
-#define PEDAL_MIN 0        // percent
-#define PEDAL_MAX 100      // percent
-#define CURRENT_SP_MIN 0   // percent
-#define CURRENT_SP_MAX 100 // percent
 
-#define GEAR_FAULT_THRESHOLD 3 // number of times gear fault can occur before it is considered a fault
 
 // Inputs
 static bool cruiseEnable = false;
 static bool cruiseSet = false;
```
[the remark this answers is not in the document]
OscarPortillo37: Similar to the Accel Cruise comment, Parthiv likes it, and we'll just tailor to the driver. I have an FSM marked old that supports this in case we want to go back to this, though.

Apps/Src/SendTritium.c
```diff
@@ -292,45 +293,51 @@ static void readInputs()
...
+    // Allow CRUZ_SET on for just one loop of the SendTritium task
+    if(cruiseSet) 
+    {
+        cruiseSet = false;
+        cruiseSetAllow = false;
+    }
+    if (!cruiseSetPrevious && cruiseSetButton && cruiseSetAllow) 
+    {
+        cruiseSet = true;
+    }
```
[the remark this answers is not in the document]
OscarPortillo37: This is new logic that allows us to easily set a new cruiseVelSetpoint while in cruise. It does this by performing the regular rising edge detection, but it also has cruiseSetAllow, which makes sure that we've fully unpressed before we press again, and this allows us to only have cruiseSet set to true for one task cycle.

Apps/Src/SendTritium.c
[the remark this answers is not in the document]
diyarajon: For future reference- the decision was to go with the existing implementation

Apps/Src/SendTritium.c
```diff
@@ -1,84 +1,72 @@
...
 
 // Outputs
-float currentSetpoint = 0;
-float velocitySetpoint = 0;
-float cruiseVelSetpoint = 0;
-float busCurrentSetPoint = 69.0f; // why is this float and not int?
+static float currentSetpoint = 0.0f;
+static float velocitySetpoint = 0.0f;
+static float cruiseVelSetpoint = 0.0f;
+static float busCurrentSetPoint = 69.0f; // why is this float and not int?
```
aidenmadaffri, line 47: Is this a question we can answer or resolve?

Apps/Src/SendTritium.c
```diff
@@ -213,9 +215,8 @@ static void dumpInfo()
     printf("State: %s\n\r", stateName);
     printf("cruiseEnable: %d\n\r", cruiseEnable);
     printf("cruiseSet: %d\n\r", cruiseSet);
-    printf("onePedalEnable: %d\n\r", onePedalEnable);
     printf("brakePedalPercent: %d\n\r", brakePedalPercent);
```
aidenmadaffri, line 218: brakePedalPercent is now a uint8_t so this should be %u

Apps/Src/SendTritium.c
```diff
@@ -232,105 +233,134 @@ static void readInputs()
...
 
-    uint32_t latest_pedal = Pedals_Read(BRAKE);
+    uint8_t latest_pedal = Pedals_Read(BRAKE);
 
     if (brakeSaturationCt < 3)
     {
-        if (latest_pedal == 100)
-            brakeSaturationCt++;
+        if (latest_pedal == 100) 
+            {brakeSaturationCt++;}
```
aidenmadaffri, line 244: Not sure what our current style guidelines are, but I think this if statement structure is different than what is in the rest of the code base.

Apps/Src/SendTritium.c
```diff
@@ -1,84 +1,72 @@
...
 static bool cruiseEnable = false;
 static bool cruiseSet = false;
-static bool onePedalEnable = false;
-static bool regenEnable = false;
 
 static uint8_t brakePedalPercent = 0;
 static uint8_t accelPedalPercent = 0;
 
-static Gear_t gear = NEUTRAL_GEAR;
+static Gear_t gear = PARK_GEAR;
```
aidenmadaffri, line 41: I might have missed something on the mechanical side of things but is our neutral gear now actually a park? I was just under the assumption that everything in the car can still move by itself like neutral in a normal car, like our neutral is really just a switch for us to disable the pedals.

Apps/Src/SendTritium.c
```diff
@@ -350,8 +380,9 @@ static void readInputs()
 static uint8_t map(uint8_t input, uint8_t in_min, uint8_t in_max, uint8_t out_min, uint8_t out_max)
```
aidenmadaffri, line 380: FYI that a PR is about to merged in to make this map function return a float value. Will probably need to be merged into this branch but shouldn't have any conflicts I think though.

aidenmadaffri: Actually now that I look a little further down, it might take some merge resolution where you are using the map function.

Apps/Src/SendTritium.c
```diff
@@ -459,120 +487,72 @@ static void NeutralDriveDecider()
...
+        currentSetpoint = percentToFloat(map(accelPedalPercent, ACCEL_PEDAL_THRESHOLD, PEDAL_MAX, CURRENT_SP_MIN, CURRENT_SP_MAX));
     }
-    velocitySetpoint = -MAX_VELOCITY;
-    currentSetpoint = percentToFloat(map(accelPedalPercent, ACCEL_PEDAL_THRESHOLD, PEDAL_MAX, CURRENT_SP_MIN, CURRENT_SP_MAX));
-    cruiseEnable = false;
-    onePedalEnable = false;
+
+    // Turn brakelight on/off
+    brakeUpdate();
 }
```
aidenmadaffri, line 503: There is an open issue ticket (#353) about limiting our maximum speed in the reverse gear. Not sure if that's something we'd want to address in this PR or not.

Apps/Src/SendTritium.c
```diff
@@ -459,120 +487,72 @@ static void NeutralDriveDecider()
...
-        state = FSM[NEUTRAL_DRIVE];
-    }
-    else if (onePedalEnable)
+    // If you're no longer in FORWARD_GEAR or you're braking, exit cruise
+    if (brakePedalPercent == BRAKE_PRESSED || gear == PARK_GEAR || gear == REVERSE_GEAR)
     {
+        state = FSM[PARK_STATE];
         cruiseEnable = false;
-        state = FSM[ONEPEDAL];
     }
```
aidenmadaffri, line 549: In most cars, pressing the brake while in cruise control will just return you to normal forward gear. It essentially serves as a quick way to deactivate cruise. I think muscle memory might make that something we should try to mimic.

aidenmadaffri: This also applies to the other cruise state deciders.

Apps/Src/SendTritium.c
```diff
@@ -796,9 +646,15 @@ void Task_SendTritium(void *p_arg)
...
+    prevState = FSM[PARK_STATE];
+    UpdateDisplay_SetGear(PARK_GEAR);
+
+    // Initialize Regen & Cruise disabled
+    // NOTE: No regen on Daybreak
+    UpdateDisplay_SetRegenState(DISP_DISABLED);
+    UpdateDisplay_SetCruiseState(DISP_DISABLED);
 
 #ifndef SENDTRITIUM_EXPOSE_VARS
     CANDATA_t driveCmd = {
```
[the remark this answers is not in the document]
aidenmadaffri: Agreed!

Tests/Test_App_SendTritium.c
```diff
@@ -133,465 +154,275 @@ void Task1(void *arg)
...
+    //     (OS_PRIO)TASK_SEND_TRITIUM_PRIO,
+    //     (CPU_STK*)SendTritium_Stk,
+    //     (CPU_STK_SIZE)WATERMARK_STACK_LIMIT/10,
+    //     (CPU_STK_SIZE)TASK_SEND_TRITIUM_STACK_SIZE,
+    //     (OS_MSG_QTY) 0,
+    //     (OS_TICK)NULL,
+    //     (void*)NULL,
+    //     (OS_OPT)(OS_OPT_TASK_STK_CLR),
+    //     (OS_ERR*)&err
+    // );
```
aidenmadaffri, line 173: Will we be able to run the test if it doesn't have an associated task?

aidenmadaffri (commented):
I had a few nitpicky things, but overall this is such a great PR. The FSM is so much cleaner and clearer to understand. This will definitely be much easier to debug as well.

Apps/Src/SendTritium.c
OscarPortillo37: Park is functionally the exact same as neutral, though perhaps the name is a tad misleading. We just call it "Park" (Jiwoo and Arnav came up with the name when we first started correcting the FSM).  I'll add a comment to clarify.

aidenmadaffri: Gotcha makes sense. I guess as the driver I would assume that Park means the drive axle is locked, but the car can def still move with this. I think we should make sure the driver knows

Apps/Src/SendTritium.c
OscarPortillo37: Good catch. Thanks!

Apps/Src/SendTritium.c
OscarPortillo37: Good point. This has been changed & the FSM diagram has been updated.

Apps/Src/SendTritium.c
OscarPortillo37: I talked to Nathaniel, and he mentioned that this shouldn't be much of an issue. Your concern is basically the same as applying a gearswitch threshold (which I originally had), but since our car doesn't use gears in the traditional sense, we should be fine - it's all electrical anyway.

krazykrayon:
For the cruise to park transition, cruiseEnable is set to false to avoid going back in case it was a brake. This transition only relies on a gear change though. Shouldn't this be on the transition from cruise to forward since that one is triggered off of brake?

krazykrayon:
Screenshot 2025-02-08 115149

I might be understanding wrong, but for accelerate cruise, shouldn't the cruiseVelSetpoint be changed to the new velocity regardless of if cruise_set is pressed? Or is the acceleration just temporarily and eventually it goes back to the old velocity?

OscarPortillo37:
> Screenshot 2025-02-08 115149
> 
> I might be understanding wrong, but for accelerate cruise, shouldn't the cruiseVelSetpoint be changed to the new velocity regardless of if cruise_set is pressed? Or is the acceleration just temporarily and eventually it goes back to the old velocity?

The acceleration is temporary unless the driver selects cruise set to explicitly change to the new accelerated speed

Apps/Src/SendTritium.c
OscarPortillo37: I'd been kind of putting off addressing that bit truthfully lol, but after some research, I've found the following:

- For context, generally, bus current is the DC current drawn by the motor controller, which is then stepped up/down into AC current/DC modulated (three-phase DC modulation, six-step, etc) into the required current/voltage that's fed to the motor. In our case, the motor controller takes high voltage DC & converts it into lower voltage AC (per the Wavesculptor Manual).
- The busCurrentSetPoint value should thus be 0.0f-1.0f, representing a percentage with respect to the maximum possible motor controller current draw.
- The max instantaneous current that the motor controller can draw is 122A, so the percentage is calculated with respect to that. According to Parthiv, the battery can support a peak current draw of 64A, and continuous 30A for this gen (max 135A next-gen, not sure what next-gen continuous would be). As such, I've set it the bus current limit to 30/122 = ~0.2459 (or 24.59%)
- Nevertheless, we'll have to test this once the motor controller is fixed (just in case), and we'd need to do further testing to check for an appropriate current limit (might be the same idk) once we incorporate regen in next-gen.

Tests/Test_App_SendTritium.c
OscarPortillo37: I've made it to where we can do so. 

Previously:
- Tests ran the task and had a bunch of macros in the SendTritium header and source file to allow this (which was a bit messy)
- In the actual test, we'd have some rather tedious overhead where we went from brake (which no longer exists under this FSM) & then cycled through a bunch of states to arrive at the desired state you wanted to test.

Now:
- Tests call the handlers and deciders & set parameters more manually, allowing us to both avoid macros and directly enter a state + test it. With this approach, we have the same testing of SendTritium's handlers and deciders in a cleaner way.
- This doesn't lose any of the functionality because we only actually tested handlers and deciders in the previous test - we couldn't do CAN, readInputs, and display stuff (and we used a macro to ignore them anyway)

Let me know if you have any questions or concerns!

aidenmadaffri (dismissed):
Looks good to me!

Apps/Inc/SendTritium.h
```diff
@@ -14,14 +14,47 @@
...
 #define MOTOR_MSG_COUNTER_THRESHOLD (MOTOR_MSG_PERIOD)/(FSM_PERIOD)
 
+#define MAX_VELOCITY 20000.0f // rpm (unobtainable value)
+#define MIN_CRUISE_VELOCITY mpsToRpm(20.0f)    // rpm
+#define MAX_GEARSWITCH_VELOCITY mpsToRpm(8.0f) // rpm
+
+// Used to define threshold for when to start powering the motor for movement
+#define ACCEL_PEDAL_THRESHOLD 15 // percent
+
+#define BRAKE_UNPRESSED 100 // percent - the switch is digital
```
MichaelMohn: This is gonna be analog right

Drivers/Src/Pedals.c
```diff
@@ -41,7 +41,7 @@ void Pedals_Init(){
  * @param   pedal_t, ACCELERATOR or BRAKE as defined in enum
  * @return  percent amount the pedal has been pressed in percentage
  */
-int8_t Pedals_Read(pedal_t pedal){
+uint8_t Pedals_Read(pedal_t pedal){
     if (pedal == BRAKE){
         return (BSP_GPIO_Read_Pin(PORTC, GPIO_Pin_15))?100:0;
```
MichaelMohn: prob dont want this if analog

Apps/Src/SendTritium.c
```diff
@@ -157,12 +135,43 @@ extern const float pedalToPercent[];
 static float percentToFloat(uint8_t percent)
 {
     if (percent > 100)
-    {
         return 1.0f;
-    }
     return pedalToPercent[percent];
```
MichaelMohn: this is quite possibly the most disappointing line of code I have ever reviewed.

JOKINGGGG. I LOVE YOU OSCAR

MichaelMohn: do this: 
return static_cast<float>(percent) / 100.0f;

Apps/Src/SendTritium.c
```diff
@@ -157,12 +135,43 @@ extern const float pedalToPercent[];
...
 }
 
+/**
+ * @brief Updates the brakelight and brake indication on display
+ */
+static void brakeUpdate(){
+    if(brakePedalPercent == BRAKE_PRESSED)
```
MichaelMohn: we gonna need brakes threshold for analog

Apps/Src/SendTritium.c
```diff
@@ -292,45 +313,70 @@ static void readInputs()
...
+    } 
 
-    // Toggle
-    // if(onePedalButton != onePedalPrevious && onePedalPrevious){onePedalEnable = !onePedalEnable;}
-    // if(!regenEnable) onePedalEnable = false;
-    // onePedalPrevious = onePedalButton;
-    onePedalEnable = false;
-    onePedalPrevious = false;
+    // Toggle CRUZ_EN
+    if(cruiseEnableButton != cruiseEnablePrevious && cruiseEnablePrevious) // Falling edge toggle/CRUZ_EN detection
```
MichaelMohn: can u put some parenthesis to clear up what this line means.

MichaelMohn: also some comments lol me and diya are confused.

PR #434, opened 2025-01-27
Refactor to use hardware floating point calculations

### LINKED ISSUE #382: Remove PedalToPercent now that floats have been validated
**Describe the problem this feature will solve**

In #367 , @Cam0Cow fixed the Micrium OS floating point issue, which means we can now use the floating point unit. Even though technically an array access is still a faster access than converting from an integer 0-100 to a float 0.0-1.0, it's probably more understandable and cleaner to sacrifice that _extremely slightly faster_ access for more readable code and less unnecessary fluff and context. See **Additional context** for more info.

**Describe the proposed solution**

Get rid of PedalToPercent, and refactor SendTritium to use float conversions since now we use the FPU in the hardware.

**Describe possible alternatives**
Alternative would be to continue to use the lookup table. This is fine and currently works, so it's up for debate whether it's worth it to switch to using the FPU and getting rid of PedalToPercent because we would need to reverify SendTritium and that takes time. This would be an M2.0.1 issue, so it would be fixed after M2.0 release.

**Additional context**
The Floating Point Unit on the Controls board used to be unusable because of some weird initialization error in the Micrium uCOS-III RTOS. As a result, hardware float operations had to be turned off, but software float operations are far slower. We need to send Current information to our Motor Controller in the form of floats, but we work with it in the form of an integer percentage 0-100, and converting it to a float from 0.0-1.0 would have been slow since it would have been a software float operation.

As a result, [PedalToPercent](./Apps/Src/PedalToPercent.c) is a piece of code that was intended to circumvent the problems created by not having access to the Floating Point Unit on the board. PedalToPercent is an array with indices from 0 to 100. Each position in the array has the float value of the corresponding index. For example, index 45 has a value of 0.45. Since these floats are stored in a big const array, they are compiled into the program at the start, so we can just pull their values without relying on using software float operations. No conversion is being done, rather we are just indexing into an array. All conversions from integer percentages to floats (mainly in [SendTritium](./Apps/Src/SendTritium.c)) are done by merely indexing into PedalToPercent (see the function ```percentToFloat``` in SendTritium).

aidenmadaffri:
Removed PedalToPercent lookup process to now utilize the hardware FPU.

Closes #382.

# Quality Assurance Checklist 

## Requirements 
- [x] Code Build checks pass 
- [x] No merge conflicts 
- Testing
    - [ ] Tested on hardware
- [x] Did you have fun?

## Things to Consider 


diyarajon (changes requested):
Good stuff!
I left a comment. The next step is to write a test file to test on hardware. I would suggest looking at Test_App_FaultState and how the tasks are created there.

Apps/Src/SendTritium.c
```diff
@@ -387,11 +372,7 @@ static void ForwardDriveHandler()
         UpdateDisplay_SetRegenState(DISP_DISABLED);
     }
     velocitySetpoint = MAX_VELOCITY;
-    // printf("accelPedalPercent: %d\n\r", accelPedalPercent);
-    uint8_t after_map = map(accelPedalPercent, ACCEL_PEDAL_THRESHOLD, PEDAL_MAX, CURRENT_SP_MIN, CURRENT_SP_MAX);
-    // printf("map(accelPedalPercent): %d\n\r", after_map);
-    currentSetpoint = percentToFloat(after_map);
-    // print_float("float(map(accelPedalPercent)): %d\n\r", currentSetpoint);
+    currentSetpoint = map(accelPedalPercent, ACCEL_PEDAL_THRESHOLD, PEDAL_MAX, CURRENT_SP_MIN, CURRENT_SP_MAX) / 100.0;
```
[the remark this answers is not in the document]
aidenmadaffri: Integrated into map function!

aidenmadaffri:
> Good stuff! I left a comment. The next step is to write a test file to test on hardware. I would suggest looking at Test_App_FaultState and how the tasks are created there.

Test should be covered in Test_Float.c, Roie used this when documenting and fixing the original FPU issue.

Apps/Src/SendTritium.c
```diff
@@ -369,7 +354,7 @@ static uint8_t map(uint8_t input, uint8_t in_min, uint8_t in_max, uint8_t out_mi
         uint8_t in_range = in_max - in_min;    // Input range
         uint8_t out_range = out_max - out_min; // Output range
         uint8_t offset_out = out_min;
-        return (offset_in * out_range) / in_range + offset_out; // slope = out_range/in_range. y=mx+b so output=slope*offset_in+offset_out
+        return ((offset_in * out_range) / in_range + offset_out) / 100.0; // slope = out_range/in_range. y=mx+b so output=slope*offset_in+offset_out
```
diyarajon, line 357: In pedalsToPercent, the returned value is always fixed to the hundredths place. Is that something we need to enforce because it seems like the previous implementation was assuming that the float always ends in the hundredths place

diyarajon (changes requested):

Apps/Src/SendTritium.c
aidenmadaffri: I think this was just a limitation of the pedalsToPercent method. Any more decimal places would have required a lot more storage. The [Prohelion documentation](https://docs.prohelion.com/Motor_Controllers/WaveSculptor22/User_Manual/Appendix_C.html#data-field) indicates no required precision. It can take any 32-bit float value.

aidenmadaffri:
Test has been written and validated. Map works as expected.

krazykrayon:
Im a little confused how mapToPercent is returning a percent. The y = mx + b equation, x and b are the same units. The slope is unitless. Then y must be in the same units as x and b. Then y / 100 wouldn't be a percentage. I might just be understanding wrong. 

Otherwise, I think it should work assuming I was just being stupid in understanding. I'm still a little confused why the table was being used in the first place and why floating points work now, but it should be good (according to my semi-broken understanding of the code).

aidenmadaffri:
> Im a little confused how mapToPercent is returning a percent. The y = mx + b equation, x and b are the same units. The slope is unitless. Then y must be in the same units as x and b. Then y / 100 wouldn't be a percentage. I might just be understanding wrong.
> 
> Otherwise, I think it should work assuming I was just being stupid in understanding. I'm still a little confused why the table was being used in the first place and why floating points work now, but it should be good (according to my semi-broken understanding of the code).

The values in this example are unitless. The input parameters are integer values from 0-100. They already represent a percentage, they are just in an integer range. The map function scales the input range (for example 15-100) to a given output range (0-100), this generates a scaled percentage value in that output range (in this case 0-100). We this to be in range 0.00-1.00, so we divide by 100.0 to turn it into a true float value.

Apps/Inc/SendTritium.h
```diff
@@ -98,5 +98,17 @@ EXPOSE_SETTER(float, velocitySetpoint)
...
+ * in_min to in_max is mapped to out_min to out_max.
+ * @param input input integer value
+ * @param in_min minimum value of input range
+ * @param in_max maximum value of input range
+ * @param out_min minimum value of output range
+ * @param out_max maximum value of output range
+ * @returns float value from (out_min / 100.0) to (out_max / 100.0)
+ */
+float mapToPercent(uint8_t input, uint8_t in_min, uint8_t in_max, uint8_t out_min, uint8_t out_max);
+
```
OscarPortillo37, line 112: Perhaps this was intentional, but shouldn't this be included within the include guard (before the last #endif)?

OscarPortillo37:
Altogether, this is a great PR! Tests are comprehensive, and this cleans up the pedal-to-percent features a lot!

Apps/Inc/SendTritium.h
aidenmadaffri: Good catch! It should be included.

IshDeshpa (approved):
LGTM

diyarajon (approved):
LGTM!

PR #435, opened 2025-02-01
Ensure ignition switch is switched to Array On before Motor On

### LINKED ISSUE #431: Ignition Cycling
**Describe the bug**
The power board may power cycle and ignition state may not start out at LV ignition state. This caused immediate precharge without visiting the prior states and blew the fuse, as well as blowing the high voltage side of the motor controller

**To Reproduce**
1. Which feature is being tested?
2. Which requirements are being tested?
3. Which test is used?
4. What is the hardware configuration?
5. What combination of inputs cause the error?
6. What does the error look like?
7. Does this issue already exist?

**Expected behavior**

**Screenshots**

**Additional context**

aidenmadaffri:
Closes #431.

# Quality Assurance Checklist 

## Requirements 
- [x] Code Build checks pass 
- [x] No merge conflicts 
- Testing
    - [ ] Tested on hardware
- [x] Did you have fun?

## Things to Consider 
- [x] Even if the above are checked, is this the best way of writing my code? 

### DIFF
--- a/Apps/Src/ReadCarCAN.c
+++ b/Apps/Src/ReadCarCAN.c
@@ -71,6 +71,9 @@ static uint8_t HVPlusMinusMotorOldestMsgIdx = 0;
 static bool arrIgnStatus = false;
 static bool mcIgnStatus = false;
 
+// Array ignition visited bool to ensure ignition switch goes through each state
+static bool arrIgnVisited = false;
+
 // Boolean to indicate precharge status for Array Precharge Bypass Contactor (PBC) and Motor Controller PBC
 static bool arrPBCComplete = false;
 static bool mcPBCComplete = false;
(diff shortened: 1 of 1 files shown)


krazykrayon (commented):
Looks good to me! (Assuming I understand correctly - see below)

Just so I make sure I didn't misunderstand, the switch goes off -> array -> motor, and this code ensures that array is visited before the switch is at motor. And if the car restarts for whatever reason, the code will restart, and the default value is that array visited is false, so then the driver turns the switch back to array, then back to motor to get the code to actually do stuff. If that all is correct, then the code should be good.

aidenmadaffri:
> Looks good to me! (Assuming I understand correctly - see below)
> 
> Just so I make sure I didn't misunderstand, the switch goes off -> array -> motor, and this code ensures that array is visited before the switch is at motor. And if the car restarts for whatever reason, the code will restart, and the default value is that array visited is false, so then the driver turns the switch back to array, then back to motor to get the code to actually do stuff. If that all is correct, then the code should be good.

That is correct!

diyarajon:
resolved by #440

PR #439, opened 2025-03-15
Display Info page redesign + Fault and Evac issues

guytonde:
# Quality Assurance Checklist 

## Requirements 
- [X] Followed Coding Style Guide 
- [X] Presented/discussed in some capacity with others on the Controls team
- [X] Code Build checks pass 
- [ ] No merge conflicts 
- [X] Software feature has documentation for it updated in both ReadTheDocs and the comments
- [X] Software feature has documentation for it updated
- Testing
    - [X] Tested on hardware
- [X] Did you have fun?

## Things to Consider 


Apps/Src/UpdateDisplay.c
```diff
@@ -40,56 +40,30 @@ typedef enum{
...
 	HEAT_SINK_TEMP,
 	GEAR,
 	// Fault code components
 	OS_CODE,
 	FAULT_CODE,
-    NUM_COMPONENTS
 } Component_t;
 
-static uint32_t componentVals[NUM_COMPONENTS] = {0};
+uint32_t DISPLAY_COMP_VALS[NUM_COMPONENTS] = {0};
```
IshDeshpa: for cleanliness i probably wouldn't format the name of this array as all caps so that it's not mistaken for a macro. or if you want to make it a const array it would probably make more sense to make it all caps then.

Tests/Test_Driver_Display.c
IshDeshpa, line 1: there's a large amount of commented stuff in this test now; if it's no longer necessary i would remove the comments.

IshDeshpa (commented):
i started reviewing but shit started getting wonky when i was looking at the display driver (changes were misaligned hella) so i stopped. can you merge in master (and fix conflicts if necessary) so that i can continue my review

Drivers/Src/Display.c
```diff
@@ -5,142 +5,324 @@
...
  */
 
 #include "Display.h"
-#include "bsp.h"   // for writing to UART
 #include "Tasks.h" // for os and fault error codes
+#include "bsp.h"   // for writing to UART
+
+#include "ReadCarCAN.h"
+#include "ReadTritium.h"
+#include "UpdateDisplay.h"
```
Lakshay983, line 17: Why does display need an awayness of ReadCarCAN or ReadTritium? 

UpdateDisplay should just call these functions given ReadCarCAN and ReadTrtium as inputs.

Tests/Test_Driver_Display.c
guytonde: The commented stuff is useful for testing similar things in different ways (using display driver vs no driver for sanity checking). Removing all the comments compromises nothing but for comp it might be a nice to have?

Apps/Src/UpdateDisplay.c
guytonde: Bet ok, my intention was for it to be like a global variable but not a const (as it needs to have its values updated). I guess it could be made into a const but it would be sketchy as it violates assumptions (and probably at least throws a warning).

I can change it to be `g_display_comp_vals`

Apps/Inc/ReadCarCAN.h
```diff
@@ -21,11 +21,11 @@
...
-	READCARCAN_ERR_NONE,
-	READCARCAN_ERR_CHARGE_DISABLE,		   // Received a charge disable msg
-	READCARCAN_ERR_MISSED_MSG,			   // Didn't receive a BPS charge msg in time
-	READCARCAN_ERR_DISABLE_CONTACTORS_MSG, // Ignition is turned to neither (off due to LV) or both at the same time (impossible) are on at
-	READCARCAN_ERR_BPS_TRIP				   // Received a BPS trip msg (0 or 1)
+	READCARCAN_ERR_NONE = 0x0000,
+	READCARCAN_ERR_CHARGE_DISABLE = 0x000F,		    // Received a charge disable msg
+	READCARCAN_ERR_MISSED_MSG = 0x00F0,			    // Didn't receive a BPS charge msg in time
+	READCARCAN_ERR_DISABLE_CONTACTORS_MSG = 0x0F00, // Ignition is turned to neither (off due to LV) or both at the same time (impossible) are on at
+	READCARCAN_ERR_BPS_TRIP = 0xF000				// Received a BPS trip msg (0 or 1)
```
diyarajon: could you make it 0xaaaa, 0xbbbb, etc because it would be easier to identify. or make them super distinct

diyarajon (commented):
let's test on new leader, but may need to change the parent branch or whatever it's called to feature/NewPinout

Apps/Src/UpdateDisplay.c
```diff
@@ -173,141 +147,141 @@ UpdateDisplayError_t UpdateDisplay_SetPage(Page_t page){
...
+	return UPDATEDISPLAY_ERR_NONE;
 }
 
 UpdateDisplayError_t UpdateDisplay_SetArray(bool state){
-	componentVals[ARRAY] = state;
-    
-    return UPDATEDISPLAY_ERR_NONE;
+	DISPLAY_COMP_VALS[ARRAY] = state;
+	
+	return UPDATEDISPLAY_ERR_NONE;
```
Lakshay983: Can you write functions for displaying the Contactor states. Mostly important for Motor and Array precharge contactors

Apps/Src/UpdateDisplay.c
```diff
@@ -40,56 +40,30 @@ typedef enum{
...
+	PACK_CURRENT,
+	PACK_TEMP,
 	MC_BUS_VOLTAGE,
 	MC_BUS_CURRENT,
 	HEAT_SINK_TEMP,
 	GEAR,
 	// Fault code components
 	OS_CODE,
 	FAULT_CODE,
-    NUM_COMPONENTS
```
Lakshay983, line 53: Why'd you get rid of this and hard code it in Display.h?

Drivers/Src/Display.c
```diff
@@ -5,142 +5,324 @@
...
+        .compOrCmd = (char*) DISPLAY_COMP_STR[13], // "pc"
+        .attr = "val",
+        .op = "=",
+        .numArgs = 1,
+        .argTypes = {INT_ARG},
+        {
+            {.num = DISPLAY_COMP_VALS[13]}
+        }
+    };
+    Display_Send(packcurr_cmd);
```
Lakshay983: i've seen this same slice of code a few times. Can this not be turned into a generic function call (with preferably not magic number sizes, you could prob don some funny preprocessor things).

Apps/Src/UpdateDisplay.c
```diff
@@ -173,141 +147,141 @@ UpdateDisplayError_t UpdateDisplay_SetPage(Page_t page){
...
  * @brief Loops through the display queue and sends all messages
  */
 void Task_UpdateDisplay(void *p_arg) {
-    OS_ERR err;
-    while (1) {
+	OS_ERR err;
+	while (1) {
 		for(Component_t comp = ARRAY; comp <= GEAR; comp++){
 			if (comp != REGEN_ST && comp != CRUISE_ST){
 				UpdateDisplay_SetComponent(comp);
```
Lakshay983: imo get the UpdateDisplayError_t returns from UpdateDisplay_SetComponent and if the error code above that threshold u set in Display.c then do wtv display fault stuff u have.

Apps/Src/UpdateDisplay.c
```diff
@@ -173,141 +147,141 @@ UpdateDisplayError_t UpdateDisplay_SetPage(Page_t page){
 		}
 	};
 
-    Display_Send(pgCmd);
+	Display_Send(pgCmd);
 	
 	return UPDATEDISPLAY_ERR_NONE;
 }
 
 /* WRAPPERS */
```
Lakshay983, line 90: For wrapper functions can you error bound them.

For example, if you input a SOC of 110. Display some hella fucked number like -1 so we can tell immediately

SOC is a bad example since you can just visually look at  it, but for like battery current it can be a bit harder to tell. Putting -1 is a good way to know smth is messed up likely on how the value is read into the code.

Lakshay983 (changes requested):
lgtm other than what i've attatched.

I'd like to see a pic or vid of what the display looks like irl before I approve since it's hard to tell by just looking at the code.

Apps/Src/UpdateDisplay.c
guytonde: because of not needing to include the enum in Display.h/c. This number is useful to have in Display.c but is defined in UpdateDisplay.c.

Apps/Src/UpdateDisplay.c
guytonde: it isn't possible to display a -1 as an int, i would have to add a separate text item with txt='-' and make it visible/invisible to get a -1. It's probably fine IMO. If we need, we can add more things to the display easily later on.

Apps/Src/UpdateDisplay.c
guytonde: There is no display fault. It doesn't actually do anything. We have no way of verifying that there was a fault. It's mainly symbolic 💀

Drivers/Src/Display.c
guytonde: It only needs these for the error enums. I tried moving to like task.h and some stuff broke, so I just include them here. It's not that big a deal since nothing else is actually used. But lmk if you had some other solution in mind that doesn't involve a full rewrite.

guytonde:
There is no special evac page anymore in favor of displaying the fault page with evac instruction (so we can get detailed fault info). There have been some component changes in the display. The blink component has the functionality to either be visible or not, but I did not see any code for turning on the blinkers so this will need to be fixed once merged with NewPinout.

Apps/Inc/ReadCarCAN.h
guytonde: yep, added this in latest commit

Apps/Inc/UpdateDisplay.h
```diff
@@ -95,19 +95,19 @@ UpdateDisplayError_t UpdateDisplay_SetVelocity(uint32_t mphTenths);
...
- * @brief Sets the motor contactor indicator state on the display
- * @param state motor contactor on (true) or off (false)
- * @returns UpdateDisplayError_t
- */
-UpdateDisplayError_t UpdateDisplay_SetMotor(bool state);
+// /**
+//  * @brief Sets the motor contactor indicator state on the display
+//  * @param state motor contactor on (true) or off (false)
+//  * @returns UpdateDisplayError_t
+//  */
```
IshDeshpa: why add comments here on a multiline comment

IshDeshpa: just remove this stuff if you want it removed

Apps/Src/ReadCarCAN.c
```diff
@@ -81,14 +81,21 @@ static uint32_t SOC = 0;
 static uint32_t SBPV = 0;
 
 // Error assertion function prototype
-static void assertReadCarCANError(ReadCarCAN_error_code_t rcc_err);
+// static void assertReadCarCANError(ReadCarCAN_error_code_t rcc_err);
```
IshDeshpa, line 84: del

Apps/Src/ReadCarCAN.c
```diff
@@ -255,7 +258,7 @@ void attemptTurnArrayPBCOn(void)
     if (arrPBCComplete && chargeEnable)
     {
         Contactors_Set(ARRAY_PRECHARGE_BYPASS_CONTACTOR, ON, true); // Turn on
-        UpdateDisplay_SetArray(true);
+        // UpdateDisplay_SetArray(true);
```
IshDeshpa, line 261: No motor/array contactor indicators on new nextion screen?

Apps/Src/ReadTritium.c
```diff
@@ -30,7 +30,7 @@ CANDATA_t motorstatusmsg = {0};
 static OS_TMR MotorWatchdog;
 
 // Function prototypes
-static void assertTritiumError(tritium_error_code_t motor_err);
+// static void assertTritiumError(tritium_error_code_t motor_err);
```
IshDeshpa, line 33: del

Apps/Src/UpdateDisplay.c
```diff
@@ -173,141 +129,135 @@ UpdateDisplayError_t UpdateDisplay_SetPage(Page_t page){
...
 
-UpdateDisplayError_t UpdateDisplay_SetGear(TriState_t gear){
-    componentVals[GEAR] = gear;
-	
-    return UPDATEDISPLAY_ERR_NONE;
-}
+// UpdateDisplayError_t UpdateDisplay_SetMotor(bool state) {
+// 	g_display_comp_vals[DISP_MOTOR] = state;
+// 	return UPDATEDISPLAY_ERR_NONE;
+// }
```
IshDeshpa: del

Tests/Test_App_UpdateDisplay.c
```diff
@@ -6,74 +6,84 @@
...
-void testBoolComp(UpdateDisplayError_t(*function)(bool)){
+void delay(){
     OS_ERR e;
+    OSTimeDlyHMSM(0, 0, 0, 750, OS_OPT_TIME_HMSM_STRICT, &e);
+}
+
+void delay_short(){
+    OS_ERR e;
+    OSTimeDlyHMSM(0, 0, 0, 50, OS_OPT_TIME_HMSM_STRICT, &e);
+}
```
IshDeshpa: why even have a helper function for these

IshDeshpa: at least make it inline static ig

Tests/Test_Lights.c
IshDeshpa, line 1: Why is this even in this PR

Tests/Test_Unveiling.c
IshDeshpa, line 1: Why in this PR

Apps/Src/ReadCarCAN.c
```diff
@@ -539,10 +535,11 @@ void Task_ReadCarCAN(void *p_arg)
  * Stores the error code and calls assertTaskError with the appropriate parameters and callback handler
  * @param  rcc_err error code to specify the issue encountered
  */
-static void assertReadCarCANError(ReadCarCAN_error_code_t rcc_err)
+void assertReadCarCANError(ReadCarCAN_error_code_t rcc_err)
```
IshDeshpa, line 538: why not make static?

Apps/Inc/Tasks.h
```diff
@@ -52,9 +52,30 @@
...
+// uOS_ERR_XXXX
+#define ERR_MSG_OFFSET 9
+#define SET_ERR_MSG_HEX(prefix, arr, err) do { \
+    arr[0] = '\"'; \
+    strncpy(arr + 1, prefix, 4); \
+    strncpy(arr + 4, "_ERR_", 6); \
+    snprintf(arr + ERR_MSG_OFFSET, 5, "%04X", (err) & 0xFFFF); \
+    arr[ERR_MSG_OFFSET + 4] = '\"'; \
+    arr[ERR_MSG_OFFSET + 5] = '\0'; \
+} while(0)
```
IshDeshpa: just make this an inline function in Tasks.h so that it has type checking but will still be optimized

Apps/Src/ReadTritium.c
```diff
@@ -156,12 +156,15 @@ static inline void handler_ReadTritium_HallError(void)
  * or locking the scheduler and entering a nonrecoverable fault (all other cases)
  * @param   motor_err Bitmap with motor error codes to check
  */
-static void assertTritiumError(tritium_error_code_t motor_err)
+void assertTritiumError(tritium_error_code_t motor_err)
```
IshDeshpa, line 159: why not static

IshDeshpa (changes requested):
Just small things. Lot of commented out code that should just be removed to avoid confusion. Also the new fault stuff has been tested right

Apps/Src/ReadCarCAN.c
guytonde: there will be indicators for each contactor now, instead of just a generic array enable. The contactor state will be fetched on every display update. Similar to how it worked with the array and motor icons but without a component vals array in the middle.

Apps/Src/ReadTritium.c
guytonde: for the tests, i dont think itll blow anything up

Apps/Inc/Tasks.h
guytonde: done, its an inline function in the latest commit

Apps/Src/ReadCarCAN.c
guytonde: wont blow anything up and i can use in tests

Apps/Inc/UpdateDisplay.h
guytonde: removed

Apps/Src/ReadTritium.c
guytonde: deleted

Tests/Test_Lights.c
guytonde: idk, it doesnt hurt tho

Tests/Test_Unveiling.c
guytonde: no clue, idk, it doesnt hurt tho

PR #447, opened 2025-04-19
Contactor rewrite

Lakshay983:
# Quality Assurance Checklist 

## Requirements 
- [ ] Code Build checks pass 
- [ ] No merge conflicts 
- Testing
    - [ ] Tested on hardware
- [ ] Did you have fun?

## Things to Consider 


Apps/Src/ReadCarCAN.c
```diff
@@ -525,6 +525,23 @@ void Task_ReadCarCAN(void *p_arg)
...
+            Contactors_Set(ARRAY_PRECHARGE_BYPASS_CONTACTOR, ARRAY_PRECHARGE_ACTUAL_VALUE(dataBuf.data), true);
+
+            // Update Motor Precharge sense state
+            Contactors_Set(MOTOR_CONTROLLER_PRECHARGE_BYPASS_CONTACTOR, MOTOR_PRECHARGE_ACTUAL_VALUE(dataBuf.data), true);
+
+            // TODO: Update display
+
+            // TODO: Fault if needed
+                        
+            break;
```
diyarajon, line 540: do we need to include precharge timeout?

BSP/STM32F413/Src/BSP_GPIO.c
```diff
@@ -37,6 +37,31 @@ void BSP_GPIO_Init(port_t port, uint16_t mask, direction_t direction, bool pull_
...
+	RCC_AHB1PeriphClockCmd(1 << port, ENABLE);
+	// Configure the pins to be generic GPIO
+    GPIO_InitStruct.GPIO_Pin   = mask;
+    GPIO_InitStruct.GPIO_Mode  = direction ? GPIO_Mode_OUT : GPIO_Mode_IN;
+    GPIO_InitStruct.GPIO_OType = GPIO_OType_PP;
+    GPIO_InitStruct.GPIO_Speed = GPIO_Low_Speed;
+	GPIO_InitStruct.GPIO_PuPd  = (pull_up)?GPIO_PuPd_UP:GPIO_PuPd_NOPULL;
+
+	// Compute the offset for the port handle from the port passed in
+    GPIO_TypeDef *portHandle = GPIO_GetPort(port);
```
IshDeshpa, line 60: indent

Apps/Src/ReadCarCAN.c
```diff
@@ -525,6 +525,23 @@ void Task_ReadCarCAN(void *p_arg)
...
+        case CONTACTOR_SENSE:
+        {
+            // Update Motor Contactor sense state
+            Contactors_Set(MOTOR_CONTROLLER_CONTACTOR, MOTOR_SENSE_ACTUAL_VALUE(dataBuf.data), true);
+
+            // Update Array Precharge sense state
+            Contactors_Set(ARRAY_PRECHARGE_BYPASS_CONTACTOR, ARRAY_PRECHARGE_ACTUAL_VALUE(dataBuf.data), true);
+
+            // Update Motor Precharge sense state
+            Contactors_Set(MOTOR_CONTROLLER_PRECHARGE_BYPASS_CONTACTOR, MOTOR_PRECHARGE_ACTUAL_VALUE(dataBuf.data), true);
```
IshDeshpa, line 534: check this out, seems funky

Tests/Test_Driver_Contactors.c
```diff
@@ -1,42 +1,95 @@
...
+            Status_Leds_Write(MOTOR_CONTROLLER_FAULT_LED, false);
+            Status_Leds_Write(MOTOR_PRECHARGE_CONTACTOR_LED, true);
+            Status_Leds_Write(ARRAY_PRECHARGE_CONTACTOR_LED, false);          
+        }
+        if(Contactors_Get(ARRAY_PRECHARGE_BYPASS_CONTACTOR)){
+            Status_Leds_Write(MOTOR_CONTROLLER_FAULT_LED, false);
+            Status_Leds_Write(MOTOR_PRECHARGE_CONTACTOR_LED, false);
+            Status_Leds_Write(ARRAY_PRECHARGE_CONTACTOR_LED, true);          
+        }
+        Status_Leds_Write(OS_FAULT_LED, ON);
```
IshDeshpa: Read sense values and confirm

Drivers/Inc/Contactors.h
```diff
@@ -16,8 +16,26 @@
 #include "BSP_GPIO.h"
 #include "stm32f4xx_gpio.h"
 
+#define CONTACTOR_SENSE_DELAY 1
+
+
+// Masks for Contactor driver board's can message
+#define ARRAY_PRECHARGE_SENSE_FAULT(dataBuf)   ((dataBuf[1] >> 0) & 0x01) // Bit 8: Array precharge sense fault happened
+#define ARRAY_PRECHARGE_EXPECTED_VALUE(dataBuf) ((dataBuf[0] >> 7) & 0x01) // Bit 7: Expected array precharge sense value
```
IshDeshpa:
```suggestion
#define ARRAY_PRECHARGE_EXPECTED_VALUE(dataBuf) ((bool)(dataBuf[0] & 0x80)) // Bit 7: Expected array precharge sense value
```
something like this would return a bool

IshDeshpa: don't rly matter just suggestion

Tests/Test_Driver_IgnitionRead.c
```diff
@@ -12,26 +12,26 @@ int main(void){
             case IGN_OFF:
                 Status_Leds_Write(OS_FAULT_LED, ON);
                 Status_Leds_Write(CRUISE_IND_LED, OFF);
-                Status_Leds_Write(MOTOR_CONTACTOR_LED, OFF);
-                Status_Leds_Write(CONTROLS_FAULT_LED, OFF);                
+                Status_Leds_Write(CONTROLS_FAULT_LED, OFF); 
+                Status_Leds_Write(MOTOR_CONTROLLER_FAULT_LED, OFF);
```
IshDeshpa, line 16: why the order change?

IshDeshpa (changes requested):
idk

Apps/Src/ReadCarCAN.c
Lakshay983: Gonna change this up in #448

BSP/STM32F413/Src/BSP_GPIO.c
Lakshay983: wOrKs On My CoMpUtEr

BSP/Inc/BSP_GPIO.h
```diff
@@ -80,6 +80,16 @@ uint8_t BSP_GPIO_Get_State(port_t port, uint16_t pin);
...
 
+/**
+ * @brief   Initializes a GPIO port
+ * @param   port - port to initialize
+ * @param	mask - pins
+ * @param	direction - input or output 
+ * @param	pull_up - pulled up or not
+ * @return  None
+ */ 
+ void BSP_GPIO_Init_PullUp(port_t port, uint16_t mask, direction_t direction, bool pull_up);
```
KnockbackNemo, line 91: Are there any benefits to having a separate BSP_GPIO_Init_PullUp function instead of modifying the current BSP_GPIO_Init function to allow a pull-up resistor option?

KnockbackNemo: Or would it make sense to rename the other function BSP_GPIO_Init_PullDown for symmetry?

Drivers/Src/Contactors.c
```diff
@@ -65,19 +77,19 @@ void Contactors_Init() {
...
-            state = BSP_GPIO_Get_State(MOTOR_PRCHG_BYPASS_PORT, MOTOR_PRCHG_BYPASS);
             break;
         default:
             break;
     }
-    return (state == ON) ? ON : OFF;
+    return contactorState[contactor];
 }
 
 /**
```
KnockbackNemo, line 101: Might be helpful to update the comments since MOTOR_CONTROLLER_CONTACTOR is now the only option

Drivers/Src/Contactors.c
```diff
@@ -103,7 +115,15 @@ ErrorStatus Contactors_Set(contactor_t contactor, bool state, bool blocking) {
...
     setContactor(contactor, state);
-    bool ret = (bool)Contactors_Get(contactor);
+
+    // the motor contactor is the only contactor that controls has a sense pin for
+    if(blocking && contactor == MOTOR_CONTROLLER_CONTACTOR) {
+        // Need to add delay for pin to settle
+        OSTimeDlyHMSM(0, 0, 0, CONTACTOR_SENSE_DELAY, OS_OPT_TIME_HMSM_STRICT, &err);
+    }
+
+    // TODO: add delay between sense reads
```
KnockbackNemo: Does this still need to be added for the non-blocking case?

Tests/Test_Driver_Contactors.c
```diff
@@ -1,42 +1,95 @@
...
+        }
+        if(Contactors_Get(MOTOR_CONTROLLER_PRECHARGE_BYPASS_CONTACTOR)){
+            Status_Leds_Write(MOTOR_CONTROLLER_FAULT_LED, false);
+            Status_Leds_Write(MOTOR_PRECHARGE_CONTACTOR_LED, true);
+            Status_Leds_Write(ARRAY_PRECHARGE_CONTACTOR_LED, false);          
+        }
+        if(Contactors_Get(ARRAY_PRECHARGE_BYPASS_CONTACTOR)){
+            Status_Leds_Write(MOTOR_CONTROLLER_FAULT_LED, false);
+            Status_Leds_Write(MOTOR_PRECHARGE_CONTACTOR_LED, false);
+            Status_Leds_Write(ARRAY_PRECHARGE_CONTACTOR_LED, true);
```
KnockbackNemo: I'm guessing you've already tested this and it works, but does this overwrite the LEDs if more than once contactor is closed at a time?

Tests/Test_Driver_Contactors.c
```diff
@@ -1,42 +1,95 @@
...
-        bool set = Contactors_Get(ARRAY_CONTACTOR);
-        printf("Turning contactor %s\r\n", set ? "off" : "on");
-        Contactors_Set(ARRAY_CONTACTOR, !set, true);
+void Task_Test_Contactors(){
+
+    OS_ERR err;
+    OSTimeDlyHMSM(0, 0, 5, 0, OS_OPT_TIME_HMSM_STRICT, &err);
+    CANDATA_t recv;  
+    recv.idx = 0;
+    recv.ID = CONTACTOR_SENSE;
```
KnockbackNemo, line 51: Just curious- what is the purpose of setting the ID of a message you're planning to receive?

BSP/Inc/BSP_GPIO.h
Lakshay983: was mostly me being lazy, i didn't want to edit every BSP_GPIO init function to account for pullups and pulldowns since BSP_GPIO just had a true false for pulldowns

Tests/Test_Driver_Contactors.c
Lakshay983: u right, i haven't had the chance to send this with multiple contactors yet. Imma change this

Drivers/Src/Contactors.c
Lakshay983: I'll update the comment, it can return the status of any of the Contactor. MOTOR_CONTROLLER_CONTACTOR  is just the only one that's updated everyime you call Contactors_Get

Tests/Test_Driver_Contactors.c
Lakshay983: force of habit from how BPS's CANBus works, ig it'd be overridden.

Drivers/Inc/StatusLeds.h
guytonde, line 1: Why delete the MOTOR_CONTACTOR_LED?

Drivers/Src/Contactors.c
guytonde, line 1: Consider adding input validations in the contactors get and set. Also is there a reason you mutex protected the set but not get?

Tests/Test_Driver_Contactors.c
```diff
@@ -1,42 +1,101 @@
...
 int main(){
+    Status_Leds_Init();
+    Contactors_Init();
+    CANbus_Init(CARCAN, (CANId_t *) carCANFilterList, NUM_CARCAN_FILTERS);
     OS_ERR err;
     OSInit(&err);
     assertOSError(err);
 
+    CPU_Init();
+    OS_CPU_SysTickInit(SystemCoreClock / (CPU_INT32U) OSCfg_TickRate_Hz);
```
guytonde, line 93: OS_CPU_SysTickInit must be called after OS Start (so in a task) as per the comment in os_cpu_c.c

Drivers/Inc/StatusLeds.h
Lakshay983: it was a funny oversight on my part for the Status LEDs driver. The motor contactor led is an LED hooked up directly to the thing that drives the Contactor, so it's rlly smth you drive for an LED since it'd just drive the Contactor

Drivers/Src/Contactors.c
Lakshay983: 1) what kind of input validation?
2) I didn't rlly think it'd matter for the get if they were behind. No interrupts use the Contactors so it's not like I'll randomly context switch inside Contactor_Get since there are no scheduling points.

Tests/Test_Driver_Contactors.c
Lakshay983: poop

Tests/Test_Driver_Contactors.c
```diff
@@ -1,42 +1,101 @@
...
+        Status_Leds_Write(MOTOR_CONTACTOR_STATUS_LED, set == ON ? ON : OFF);
+        CANbus_Read(&recv, true, CARCAN);
+         // Update Array Precharge sense state
+        Contactors_Set(ARRAY_PRECHARGE_BYPASS_CONTACTOR, ARRAY_PRECHARGE_ACTUAL_VALUE(recv.data), true);
+
+        // Update Motor Precharge sense state
+        Contactors_Set(MOTOR_CONTROLLER_PRECHARGE_BYPASS_CONTACTOR, MOTOR_PRECHARGE_ACTUAL_VALUE(recv.data), true);
+
+        // clear Contactor status LEDs
+        Status_Leds_Write(MOTOR_CONTROLLER_FAULT_LED, true);
```
diyarajon:
```suggestion
```

Drivers/Src/Contactors.c
```diff
@@ -61,23 +73,22 @@ void Contactors_Init() {
...
  * @param   contactor the contactor
- *              (MOTOR_CONTROLLER_PRECHARGE_BYPASS_CONTACTOR/ARRAY_PRECHARGE_BYPASS_CONTACTOR)
  * @return  The contactor's state (ON/OFF)
  */ 
 bool Contactors_Get(contactor_t contactor) {
-    State state = OFF;
     switch (contactor) {
-        
+        case MOTOR_CONTROLLER_CONTACTOR:
+            contactorState[MOTOR_CONTROLLER_CONTACTOR] = BSP_GPIO_Get_State(MOTOR_CONTACTOR_PORT, MOTOR_CONTACTOR) == 0 ? ON : OFF;
```
diyarajon: should this be switched? or is this negative logic

diyarajon (changes requested):

Drivers/Src/Contactors.c
Lakshay983: yeah it's negative logic

Apps/Src/ReadCarCAN.c
Lakshay983: precharge timeout is in #448

Tests/Test_Driver_Contactors.c
Lakshay983: what's the change?

Apps/Src/ReadCarCAN.c
IshDeshpa, line 1: Why change ReadCarCAN in this PR if you have another open for that

BSP/Inc/BSP_GPIO.h
IshDeshpa: find and replace is awesome

BSP/STM32F413/Src/BSP_GPIO.c
```diff
@@ -37,6 +37,31 @@ void BSP_GPIO_Init(port_t port, uint16_t mask, direction_t direction, bool pull_
...
 
+/**
+ * @brief   Initializes a GPIO port
+ * @param   port - port to initialize
+ * @param	mask - pins
+ * @param	direction - input or output 
+ * @param	pull_up - pulled up or not
+ * @return  None
+ */ 
+void BSP_GPIO_Init_PullUp(port_t port, uint16_t mask, direction_t direction, bool pull_up){
```
IshDeshpa, line 48: u could get rid of this and find and replace existing GPIO inits to support pullup

Drivers/Src/Contactors.c
```diff
@@ -23,11 +27,17 @@ static OS_MUTEX contactorsMutex;
...
             break;
-        case MOTOR_CONTROLLER_PRECHARGE_BYPASS_CONTACTOR :
-            BSP_GPIO_Write_Pin(MOTOR_PRCHG_BYPASS_PORT, MOTOR_PRCHG_BYPASS, state);
+
+        // Precharge Contactors are updated by ReadCarCAN.c
+        case MOTOR_CONTROLLER_PRECHARGE_BYPASS_CONTACTOR:
+            contactorState[MOTOR_CONTROLLER_PRECHARGE_BYPASS_CONTACTOR] = state;
+            break;
+        case ARRAY_PRECHARGE_BYPASS_CONTACTOR:
+            contactorState[ARRAY_PRECHARGE_BYPASS_CONTACTOR] = state;
```
IshDeshpa, line 40: Little worried about sending/receiving from the active precharge board. what happens if active prchg hard faults and we think the contactor is on when it's actually off or vice versa. imo there should be some sort of ack to confirm that contactor state has actually changed but that could be another pr.

Drivers/Src/Contactors.c
```diff
@@ -61,23 +73,22 @@ void Contactors_Init() {
...
-            state = BSP_GPIO_Get_State(ARRAY_PRCHG_BYPASS_PORT, ARRAY_PRCHG_BYPASS);
             break;
         case MOTOR_CONTROLLER_PRECHARGE_BYPASS_CONTACTOR :
-            state = BSP_GPIO_Get_State(MOTOR_PRCHG_BYPASS_PORT, MOTOR_PRCHG_BYPASS);
             break;
         default:
             break;
     }
-    return (state == ON) ? ON : OFF;
+    return contactorState[contactor];
```
IshDeshpa, line 98: if the MOTOR_CONTROLLER_CONTACTOR is the only one you want diff behavior on why even have a switch case rather than an if statement at the beginning and a return

Drivers/Src/Contactors.c
```diff
@@ -103,7 +114,14 @@ ErrorStatus Contactors_Set(contactor_t contactor, bool state, bool blocking) {
 
     // change contactor to match state and make sure it worked
     setContactor(contactor, state);
-    bool ret = (bool)Contactors_Get(contactor);
+
+    // the motor contactor is the only contactor that controls has a sense pin for
+    if(blocking && contactor == MOTOR_CONTROLLER_CONTACTOR) {
```
IshDeshpa: == takes precedence over && but i'd still add parentheses here:
```suggestion
    if(blocking && (contactor == MOTOR_CONTROLLER_CONTACTOR)) {
```

Tests/Test_Driver_Contactors.c
IshDeshpa: I think the comment says clear but the line says true; if you're trying to clear the LED should it not be false

IshDeshpa: also it's the motor controller fault led not the contactor status led

Tests/Test_Driver_Contactors.c
```diff
@@ -1,42 +1,101 @@
...
+        if(Contactors_Get(MOTOR_CONTROLLER_CONTACTOR)){
+            Status_Leds_Write(MOTOR_CONTROLLER_FAULT_LED, true);
+        }
+        if(Contactors_Get(MOTOR_CONTROLLER_PRECHARGE_BYPASS_CONTACTOR)){
+            Status_Leds_Write(MOTOR_PRECHARGE_CONTACTOR_LED, true);
+        }
+        if(Contactors_Get(ARRAY_PRECHARGE_BYPASS_CONTACTOR)){
+            Status_Leds_Write(ARRAY_PRECHARGE_CONTACTOR_LED, true);          
+        }
+        Status_Leds_Write(TEST_CONTACTOR_TASK_HEARTBEAT, ON);
```
IshDeshpa, line 80: why do you use true some places and ON in other places. i'd say use true everywhere

IshDeshpa (commented):
reviewed

Drivers/Src/Contactors.c
Lakshay983: That should hopefully be solved by the active precharge can timer in #448 

Getting the states is rlly only for showing on the display anyway.

Drivers/Src/Contactors.c
Lakshay983: in case i want to do any other things in there later

Tests/Test_Driver_Contactors.c
Lakshay983: I like ON if we ever need to add support for negative logic LEDs

Drivers/Src/Contactors.c
```diff
@@ -61,23 +70,32 @@ void Contactors_Init() {
...
-        
+        case MOTOR_CONTROLLER_CONTACTOR:
+            // Read the GPIO and update the contactor state array in a critical section
+            OS_ERR err = 0;
+            CPU_TS ts = 0;
+            OSMutexPend(&contactorsMutex, 0, blocking ? OS_OPT_PEND_BLOCKING : OS_OPT_PEND_NON_BLOCKING, &ts, &err);
+            assertOSError(err);
+            contactorState[MOTOR_CONTROLLER_CONTACTOR] = BSP_GPIO_Get_State(MOTOR_C_SENSE_PORT, MOTOR_C_SENSE) == 0 
+                                                         ? true : false;
+            OSMutexPost(&contactorsMutex, OS_OPT_POST_NONE, &err);
```
IshDeshpa, line 86: why do we need mutex protection here

guytonde: I did it to be on the safe side of race conditions in the case of the moco contactor (aka motor enable) because that's the only contactor where we can change its value in the contactorStates array in both the set and the get. Since multiple things can call set and get, just thought we could use the synchronization here

IshDeshpa: fair

IshDeshpa (approved):
ship

PR #448, opened 2025-04-20
ReadCarCAN Rewrite for Daybreak 2025

Lakshay983:
# Quality Assurance Checklist 

## Requirements 
- [ ] Code Build checks pass 
- [ ] No merge conflicts 
- Testing
    - [ ] Tested on hardware
- [ ] Did you have fun?

## Things to Consider 


Apps/Src/Tasks.c
```diff
@@ -101,8 +104,16 @@ void throwTaskError(error_code_t errorCode, callback_t errorCallback, error_sche
     
 
     if (nonrecoverable == OPT_NONRECOV) { // Enter an infinite while loop
+        volatile static int faultLoopCount = 0;
         while(1) {
-
+            faultLoopCount++;
+
+            // periodically toggle Controls Fault LED
```
IshDeshpa: why periodically toggle rather than setting it

Apps/Src/ReadCarCAN.c
```diff
@@ -377,34 +338,41 @@ static void handler_ReadCarCAN_contactorsDisable(void)
...
-        }
-    }
-    else
-    {
-        UpdateDisplay_SetArray(false);
-        UpdateDisplay_SetMotor(false);
-    }
 }
 
+static void check_MotorControllerContactor(void){
```
IshDeshpa: would this just be checking if sense == value; and shouldn't this happen in contactors driver

Apps/Src/main.c
```diff
@@ -219,24 +206,24 @@ void Task_Init(void *p_arg){
...
+    Contactors_EmergencyDisable();
     while(1){}
 }
 
 void UsageFault_Handler(){
     __disable_irq();
-    EmergencyContactorOpen();
+    Contactors_EmergencyDisable();
     while(1){}
 }
```
IshDeshpa: are these meant to be defined here or were these just left over from some debug session. because you could just put this in the DefaultHandler (esp if it's all the same behavior)

IshDeshpa (commented):
left some comments

Apps/Src/Tasks.c
Lakshay983: cuz blinky mean thingy runny

Apps/Src/ReadCarCAN.c
Lakshay983: Was also thinking about adding additional checks like if BPS says battery is safe, ignition state, etc

Basically all the things that should be true for the motor contactor to be on

Apps/Src/main.c
Lakshay983: default handler as in the Tasks.c one? i don't remember what the default handler was. I figured might as well have a function to disable the contactors in the Contactor driver so all Contactor calls can use the static setContactor function instead of how it was before where the same code to control Contactors was split across different files.

PR #451, opened 2025-05-02
Send tritium rework MVP

OscarPortillo37:
# Quality Assurance Checklist 

## Requirements 
- [x] Code Build checks pass 
- [x] No merge conflicts 
- [x] Software feature has documentation for it updated in both ReadTheDocs and the comments
- [x] Software feature has documentation for it updated
- Testing
    - [ ] Tested on hardware
- [x] Did you have fun?

## Things to Consider 


Apps/Src/SendTritium.c
```diff
@@ -776,58 +188,75 @@ static void BrakeDecider()
...
+        if(parkReset) 
         {
-            memcpy(&driveCmd.data[4], &currentSetpoint, sizeof(float));
-            memcpy(&driveCmd.data[0], &velocitySetpoint, sizeof(float));
-            CANbus_Send(driveCmd, CAN_NON_BLOCKING, MOTORCAN);
-            motorMsgCounter = 0;
+            // If the car just turned on and isn't in park, require a reset to neutral before normal gear behavior resumes.
+            if(gear == PARK_GEAR) parkReset = false;
+            velocitySetpoint = MAX_VELOCITY;
+            currentSetpoint = 0.0f;
```
Lakshay983: add gear reset logic to the dashboard driver

Apps/Src/SendTritium.c
```diff
@@ -776,58 +188,75 @@ static void BrakeDecider()
...
+                    velocitySetpoint = MAX_VELOCITY;
+                    currentSetpoint = 0.0f;
+                    break;
+            }
         }
-#endif
+
+        memcpy(&driveCmd.data[4], &currentSetpoint, sizeof(float));
+        memcpy(&driveCmd.data[0], &velocitySetpoint, sizeof(float));
+        CANbus_Send(driveCmd, CAN_NON_BLOCKING, MOTORCAN);
```
Lakshay983: if ur doing a nonblocking send, u should handle the case where it falls, and if it's continually messing up then fault.

Apps/Src/SendTritium.c
```diff
@@ -837,3 +266,20 @@ void Task_SendTritium(void *p_arg)
         }
     }
 }
+
+static void assertSendTritiumError(SendTritium_error_code_t sterr)
+{
+    Error_SendTritium = (error_code_t) sterr;
+
+    switch(sterr)
+    {
```
Lakshay983: add default case

Drivers/Src/Dashboard.c
```diff
@@ -1,23 +1,40 @@
...
+}
+
+switch_state_t getSwitchState(dash_pin_t pin){
     switch(pin){
-        case(GEAR):
-            if(BSP_GPIO_Read_Pin(FORWARD_PORT, FORWARD)) {return FWD;}
-            else if(BSP_GPIO_Read_Pin(REVERSE_PORT, REVERSE)) {return REV;}
-            else {return NUE;}
+        case(DASH_CRUZ_SET):
+            return BSP_GPIO_Read_Pin(CRUISE_SET_PORT, CRUISE_SET) ? DASH_SWITCH_ON : DASH_SWITCH_OFF;
```
diyarajon: I'd suggest removing this for now and `switch state t`

Drivers/Src/Pedals.c
```diff
@@ -14,12 +14,12 @@
...
 static const int16_t LowerBound[NUMBER_OF_PEDALS] = {
     0, // Accelerator lower bound
-    2220, // Brake lower bound
+    500, // Brake lower bound
 };
 
 static const int16_t UpperBound[NUMBER_OF_PEDALS] = {
     3150, // Accelerator upper bound
-    3205, // Brake upper bound
+    2600, // Brake upper bound
```
diyarajon: could you # define this so it's easier access when we wanna change the value

Drivers/Src/Display.c
```diff
@@ -244,6 +246,18 @@ DisplayError_t Display_Error() {
         Display_Send(disp_flt_cmd);
         strncpy(ErrMsg_UpdateDisplay, DISP_NA_STR_LITERAL, ERR_CODE_LEN);
         memset(&Error_UpdateDisplay, 0, sizeof(error_code_t));
+    } else if (Error_SendTritium != SENDTRITIUM_ERR_NONE) {
```
guytonde: I say put send tritium right after read tritium since its just as mission critical.

Drivers/Src/Pedals.c
```diff
@@ -14,12 +14,12 @@
 // Refine in testing
 static const int16_t LowerBound[NUMBER_OF_PEDALS] = {
     0, // Accelerator lower bound
-    2220, // Brake lower bound
+    500, // Brake lower bound
```
guytonde, line 19: Is there a reason you've changed these values?

Apps/Inc/UpdateDisplay.h
```diff
@@ -49,7 +49,7 @@ typedef enum {
 #define DISP_ACTIVE STATE_2	// Actively being used right now
 
 // For gear changes
-#define DISP_NEUTRAL STATE_0
+#define DISP_PARK STATE_0
```
guytonde: Why 😭 ? We have no park gear; park gear is supp off switch 💀

Apps/Inc/UpdateDisplay.h
```diff
@@ -123,7 +123,7 @@ UpdateDisplayError_t UpdateDisplay_SetMCVoltage(uint32_t volts);
 
 UpdateDisplayError_t UpdateDisplay_SetMCCurrent(int32_t val);
 
-UpdateDisplayError_t UpdateDisplay_SetBrake(bool state);
+UpdateDisplayError_t UpdateDisplay_SetBrake(uint8_t percent);
```
guytonde: Display as it is right now cannot support anything other than a boolean here for brake because it is swtiching the visibility of a component on/off. A non bool value will have no effect. We could display the brake percent separately though if needed but that functionality is not there atm. 

Consider reverting unless u had reasoning for changing this that I'm not getting.

Apps/Src/UpdateDisplay.c
```diff
@@ -162,8 +162,8 @@ UpdateDisplayError_t UpdateDisplay_SetMCCurrent(int32_t val) {
 	return UPDATEDISPLAY_ERR_NONE;
 }
 
-UpdateDisplayError_t UpdateDisplay_SetBrake(bool state) {
-	g_display_comp_vals[DISP_BRAKE] = (uint32_t)(state);
+UpdateDisplayError_t UpdateDisplay_SetBrake(uint8_t percent) {
```
guytonde: See other comment abt brake and display

Apps/Src/ReadCarCAN.c
```diff
@@ -80,9 +80,28 @@ static bool mcPBCComplete = false;
...
 static uint32_t SBPV = 0;
 
+// Boolean to indicate internally whether bps has been checked (for Daybreak => HV+ & HV- have already been turned on
+// to avoid extra CAN messages)
+// NOTE: For nextgen, BPS status (contactors, checked, trip, etc) will be done more compactly in a packed BPS message
+static bool bps_checked = false;
+
+// Synchronization-protected event flag group signaling BPS_SAFE, if BPS
+// has been checked, & motor ready to run status
+OS_FLAG_GRP BPS_Motor_Status_Flags;
```
guytonde: Consider putting the flags in Tasks.c/h. As it is right now one would need to include ReadCarCAN.h which shouldn't be the case if I just wanted to check the state of motor safety (i.e. in SendTritium.c).

Apps/Src/SendTritium.c
```diff
@@ -1,274 +1,117 @@
...
  */
 
 #include "Pedals.h"
 #include "ReadCarCAN.h"
 #include "Minions.h"
-#include "SendTritium.h"
+#include "Dashboard.h"
 #include "ReadTritium.h"
 #include "SendCarCAN.h"
+#include "ReadCarCAN.h"
```
guytonde: Take out the send/read carcan header includes after moving the flag group to tasks.c. 
Motor can should not know of car can imo.

Apps/Src/SendTritium.c
```diff
@@ -1,274 +1,117 @@
...
-static bool onePedalButton = false;
-static bool onePedalPrevious = false;
-
-static bool cruiseEnableButton = false;
-static bool cruiseEnablePrevious = false;
-
-// FSM
-static TritiumState_t prevState; // Previous state
-static TritiumState_t state;     // Current state
+static float velocityObserved = 0.0f;
```
guytonde: Why have this and not just have a getter function for the Motor_Velocity float that's in read tritium?

Apps/Src/ReadCarCAN.c
IshDeshpa, line 1: If readcarcan is being changed a lot too just note there'll be a merge conflict to fix on that end. not big deal though

Apps/Src/SendTritium.c
IshDeshpa, line 1: good

Apps/Src/main.c
```diff
@@ -123,6 +124,7 @@ void Task_Init(void *p_arg){
     // Initialize applications
     UpdateDisplay_Init();
     SendCarCAN_Init();
+    ReadCarCAN_Init();
```
IshDeshpa: why in this PR? shouldn't this be in ReadCarCAN

IshDeshpa: actually maybe it was just for testing so disregard. but note that stuff like this will cause merge conflicts (again, not a bad thing and practically unavoidable but something y'all will have to fix when ReadCarCAN gets merged in)

Drivers/Src/Pedals.c
IshDeshpa, line 19: are these constants from testing, or are they just for some little potentiometer

BSP/STM32F413/Makefile
```diff
@@ -159,7 +159,7 @@ vpath %.c $(sort $(dir $(C_SOURCES)))
 OBJECTS += $(addprefix $(BUILD_DIR)/,$(notdir $(ASM_SOURCES:.s=.o)))
 vpath %.s $(sort $(dir $(ASM_SOURCES)))
 
-$(BUILD_DIR)/%.o: %.c Makefile | $(BUILD_DIR) 
+$(BUILD_DIR)/%.o: %.c Makefile | $(BUILD_DIR)
```
IshDeshpa: what actually changed here...

IshDeshpa (approved):
Good job removing an embarrasing amount of technical debt. Ping me for reapproval if you change anything.

IshDeshpa:
Oh also please merge in master and resolve all ur conflicts

Apps/Inc/ReadCarCAN.h
```diff
@@ -16,6 +16,20 @@
...
+#define MOTOR_ERR 1 << 2
+#define MOTOR_SAFE_TO_RUN 1 << 3
+
+// Synchronization-protected event flag group signaling BPS_SAFE, if BPS
+// has been checked, & motor ready to run status
+extern OS_FLAG_GRP BPS_Motor_Status_Flags;
+
+/**
+ * @brief Initialize ReadCarCAN
+*/
```
Lakshay983: why not put in Tasks.h if shared with other tasks?

diyarajon (changes requested):

Drivers/Src/Pedals.c
diyarajon: they'll need to be changed

Apps/Inc/SendTritium.h
```diff
@@ -11,91 +11,63 @@
 #define __SENDTRITIUM_H
 
 #include "common.h"
+#include "Dashboard.h"
 
 //#define SENDTRITIUM_PRINT_MES
 
+// Macros
```
Lakshay983: macros 👀

Apps/Src/ReadCarCAN.c
```diff
@@ -284,6 +303,9 @@ void turnMotorControllerPBCOff(void)
 {
     Contactors_Set(MOTOR_CONTROLLER_PRECHARGE_BYPASS_CONTACTOR, OFF, true);
     // UpdateDisplay_SetMotor(false);
+    OS_ERR err;
+    OSFlagPost(&BPS_Motor_Status_Flags, MOTOR_SAFE_TO_RUN, OS_OPT_POST_FLAG_CLR, &err);
+    assertOSError(err);
 }
```
Lakshay983: iirc @KnockbackNemo we burned charge enable

Apps/Src/SendCarCAN.c
```diff
@@ -163,6 +164,27 @@ static void putIOState(void){
...
+    CPU_TS ticks;
+    OS_FLAGS res_set = OSFlagPend(&BPS_Motor_Status_Flags, BPS_SAFE | BPS_CHECKED, 0, OS_OPT_PEND_FLAG_SET_ALL | OS_OPT_PEND_NON_BLOCKING, &ticks, &err);
+    assertOSError(err);
+    OS_FLAGS res_clr = OSFlagPend(&BPS_Motor_Status_Flags, MOTOR_ERR, 0, OS_OPT_PEND_FLAG_CLR_ALL | OS_OPT_PEND_NON_BLOCKING, &ticks, &err);
+    assertOSError(err);
+
+    // IF BPS Safe & in motor ignition rotary switch position & motor controller precharge bypass contactor is closed, mark motor ready to run
+    // NOTE: BPS Safe means HV+ & HV- are closed
+    // TODO: Modify this as needed when merging (to use the PR w/ motor contactor stuff -> also make MOTOR_SAFE_TO_RUN false when moco contactor turned off)
+    if(res_set && res_clr && Minions_Read(IGN_2) && BSP_GPIO_Read_Pin(MOTOR_C_SENSE_PORT, MOTOR_C_SENSE) && Contactors_Get(MOTOR_CONTROLLER_PRECHARGE_BYPASS_CONTACTOR)) {
```
guytonde: Why use minions and not Ignition driver for ign2

guytonde (changes requested):
Good stuff, left some comments here n there. prepare for big changes as stuff from other pr's get merged in tho.

Drivers/Src/Pedals.c
Lakshay983: I like the array. Makes indexing in nice. just make it not hard coded position wise, ie if suddenly brake isn't position 1 then this gets messed up.

You can do that with funny compile time macros

Tests/Test_Driver_Dashboard.c
```diff
@@ -4,10 +4,28 @@
...
     Status_Leds_Init();
     dashboardInit();
-    switch_state state = SWITCH_ERROR;
+    gear_t gear = DASH_GEAR_FAULT_ERROR;
     //uint8_t dashStatus[NUM_DASH_PINS];
+
+
+    //// Testing forced gear reset to neutral
+    // Move to non-neutral start point
+    printf("ACTION: Please move the gear switch to FWD gear\n\r");
```
Lakshay983: Printf no workie btw. I'd remove this section of the test and just use the LEDs

This'd be good for like regression or automated testing but we don't rlly do that currently

Drivers/Src/Pedals.c
Lakshay983: https://utsvt.slack.com/archives/C013F4XGG58/p1746652350524419?thread_ts=1746639015.607549&channel=C013F4XGG58&message_ts=1746652350.524419
@IshDeshpa 
@diyarajon 
@guytonde

BSP/STM32F413/Makefile
Lakshay983: Why change?

Apps/Src/SendTritium.c
```diff
@@ -1,274 +1,117 @@
...
-    case ONEPEDAL:
-        strcpy(nameStr, "ONEPEDAL");
-        break;
-    case ACCELERATE_CRUISE:
-        strcpy(nameStr, "ACCELERATE_CRUISE");
-        break;
-    default:
-        strcpy(nameStr, "UNKNOWN");
-        break;
+        Minions_Write(BRAKELIGHT, true);
```
Lakshay983: minions is dead 🦀

Apps/Src/SendTritium.c
```diff
@@ -1,274 +1,117 @@
...
+        UpdateDisplay_SetBrake(true);
+    } 
+    else if(brakePedalPercent <= BRAKE_UNPRESSED_THRESHOLD)
```
Lakshay983: how about

Minions_Write(brakePedalPercent > BRAKE_UNPRESSED_THRESHOLD ? true : false);
UpdateDisplay_SetBrake(brakePedalPercent > BRAKE_UNPRESSED_THRESHOLD ? true : false)

Apps/Src/SendTritium.c
```diff
@@ -1,274 +1,117 @@
...
-    if (gearFault != 1)
+    // Check for gear fault
+    if (gear == DASH_GEAR_FAULT_ERROR)
     {
         // Fault behavior
         if (gearFaultCnt > GEAR_FAULT_THRESHOLD)
-            state = FSM[NEUTRAL_DRIVE];
+            assertSendTritiumError(SENDTRITIUM_ERR_GEAR_FAULT);
         else
             gearFaultCnt++;
```
Lakshay983, line 117: why global?

Apps/Src/SendTritium.c
```diff
@@ -776,58 +169,66 @@ static void BrakeDecider()
...
+        CANbus_Send(powerCmd, CAN_BLOCKING, MOTORCAN); 
+        readInputs(); // read inputs from the system
+        updateDisplayState();
+
+        // Update velocitySetpoint & currentSetpoint based on gear/state
+
+        switch(gear) {
+            case DASH_FWD:
+                velocitySetpoint = MAX_VELOCITY;
+                currentSetpoint = (brakePedalPercent >= BRAKE_PRESSED_THRESHOLD) ? 0 : mapToPercent(accelPedalPercent, ACCEL_PEDAL_THRESHOLD, PEDAL_MAX, CURRENT_SP_MIN, CURRENT_SP_MAX);
```
Lakshay983: add some hysterisis with these brakePedalPercent thingies

Drivers/Src/Dashboard.c
```diff
@@ -1,23 +1,40 @@
 #include "Dashboard.h"
 
-switch_state getDashState(dashPin_t pin){
+// Boolean used to ensure that if car turns on in non-park, it'll be overriden to park 
+// until the switch is moved to park; then, it'll follow the specified gear state afterward.
+static bool neutralReset = true;
```
Lakshay983: keep static inside the getGear function

Drivers/Src/Dashboard.c
```diff
@@ -1,23 +1,40 @@
...
+
+gear_t getGear(void) {
```
Lakshay983, line 7: i'm iffy on this logic tbh, will take another look when I'm less sleep deprived

Apps/Src/SendTritium.c
```diff
@@ -776,58 +169,66 @@ static void BrakeDecider()
...
+    // Initialize display
+    UpdateDisplay_SetGear(DASH_NEU);
+    UpdateDisplay_SetRegenState(DISP_DISABLED); // Not on Daybreak
+    UpdateDisplay_SetCruiseState(DISP_DISABLED); // Probably not on Daybreak
+    UpdateDisplay_SetAccel(accelPedalPercent); 
+    UpdateDisplay_SetBrake(brakePedalPercent);
+
+    // Wait for motor ready to run
+    OSFlagPend(&BPS_Motor_Status_Flags, BPS_SAFE | BPS_CHECKED | MOTOR_SAFE_TO_RUN, 0, OS_OPT_PEND_FLAG_SET_ALL | OS_OPT_PEND_BLOCKING, &ticks, &err);
+    assertOSError(err);
```
Lakshay983: so display won't show gear states unless motor is safe to run?

maybe move the gear display code to UpdateDisplay then

Apps/Src/SendTritium.c
```diff
@@ -776,58 +169,66 @@ static void BrakeDecider()
...
-        CANbus_Send(powerCmd, CAN_BLOCKING, MOTORCAN); //<-------
-        state.stateHandler();                          // do what the current state does
-#ifndef SENDTRITIUM_EXPOSE_VARS
-        readInputs(); // read inputs from the system
-        UpdateDisplay_SetAccel(accelPedalPercent);
-#endif
-        prevState = state;
-        state.stateDecider(); // decide what the next state is
+        // Check that motor is ready to run
+        OSFlagPend(&BPS_Motor_Status_Flags, BPS_SAFE | BPS_CHECKED | MOTOR_SAFE_TO_RUN, 0, OS_OPT_PEND_FLAG_SET_ALL | OS_OPT_PEND_BLOCKING, &ticks, &err);
```
Lakshay983: i'm iffy on always pending on the motor every while loop iteration. 

if motor becomes unsafe to run we need to put it into neutral here, we can't assume the motor will just stop running, you need to actively turn it off

Apps/Src/SendTritium.c
```diff
@@ -776,58 +169,66 @@ static void BrakeDecider()
...
 
+    // Initialize display
+    UpdateDisplay_SetGear(DASH_NEU);
+    UpdateDisplay_SetRegenState(DISP_DISABLED); // Not on Daybreak
+    UpdateDisplay_SetCruiseState(DISP_DISABLED); // Probably not on Daybreak
+    UpdateDisplay_SetAccel(accelPedalPercent); 
+    UpdateDisplay_SetBrake(brakePedalPercent);
+
+    // Wait for motor ready to run
+    OSFlagPend(&BPS_Motor_Status_Flags, BPS_SAFE | BPS_CHECKED | MOTOR_SAFE_TO_RUN, 0, OS_OPT_PEND_FLAG_SET_ALL | OS_OPT_PEND_BLOCKING, &ticks, &err);
```
Lakshay983: should use OS_FLAG_WAIT_SET_ALL

Lakshay983 (changes requested):
looks really good so far

i need to think more about the implications of the OS flags and the gear implementation, but here's some other comments for now

Lakshay983:
also make sure to merge feature/NewPinout and fix merge conflicts :)

Apps/Src/main.c
OscarPortillo37: Fair enough, it was just included here to initialize the BPS & Motor flags, but I'll remove this & put the flags in Tasks.h/.c, as per other comments on this review

Drivers/Src/Pedals.c
OscarPortillo37: See Lakshay's comments here https://github.com/lhr-solar/Controls/pull/451#discussion_r2094737519

Apps/Src/ReadCarCAN.c
OscarPortillo37: Good idea, thank you! Done.

Apps/Src/SendTritium.c
OscarPortillo37: Also done, thank you!

Apps/Inc/ReadCarCAN.h
OscarPortillo37: Good idea, thank you! Done.

BSP/STM32F413/Makefile
OscarPortillo37: My bad, it's been removed

OscarPortillo37: My bad, it's been removed

Apps/Inc/UpdateDisplay.h
OscarPortillo37: Fixed, sorry I thought I'd removed all traces of park lol

Drivers/Src/Dashboard.c
OscarPortillo37: I was just keeping Mika's prev code for dash driver for that. I can comment it out, but I don't really think it matters enough to fully remove it (since it's not being utilized atm & it doesn't hurt to keep)

Drivers/Src/Display.c
OscarPortillo37: Sounds good, I had just put it last (excluding Display) bc I didn't want my stuff to take "priority" over something else

Apps/Inc/UpdateDisplay.h
OscarPortillo37: I see, thank you! It's been reverted

Apps/Src/SendCarCAN.c
OscarPortillo37: Done, thank you! I'd missed that

Apps/Src/SendTritium.c
OscarPortillo37: Fair enough, this was being used for cruise before that was cut (for easy access), but that's a good point - helps reduce redundancy

Drivers/Src/Dashboard.c
OscarPortillo37: cool, pls lmk if it looks good

Apps/Src/SendTritium.c
OscarPortillo37: maybe i'm being stupid, but is there another driver I'll be writing to for this brakelight stuff (a lights driver I'm not seeing or a CAN message I'm missing) rather than using minions? I'm just using minions atm despite being deprecated b/c I'm not sure what else to write to (or what CAN msg to send)

Apps/Src/SendTritium.c
OscarPortillo37: muy sexy 🌶️

Apps/Src/SendTritium.c
OscarPortillo37: this is probably a rlly stupid questioin, but wdym? assertSendTritiumError is a static function

Apps/Src/SendTritium.c
OscarPortillo37: sry if i'm being slow, but that doesn't seem to exist?

Apps/Src/SendTritium.c
OscarPortillo37: Makes sense...lowkey, both driveCmd & powerCmd are different (non-blocking & blocking, respectively). i think I just kept that the same from prev SendTritium & didn't question it (under the assumption that if it ain't broke, don't fix it). 

i feel like driveCmd & powerCmd should both have the same blocking/non-blockingness, though. do you think I should make them both non-blocking so that the rest of the tritium logic executes w/out stall?

Tests/Test_Driver_Dashboard.c
OscarPortillo37: sadge, i've changed it; thank you laksee

Apps/Src/SendTritium.c
Lakshay983: Just make em all blocking tbh. If you're blocking with any safety critical messages, I'd add some form of timer that triggers if you've not sent a message in x amount of time.

For example, in ReadCarCAN we have a timer that resets every time BPS gets a message and if that timer goes off then BPS may be messed up so Controls faults.

What you'd do is have a timer that you start after canbus send, and then reset it right after canbus send. This is another reason why I'm somewhat against pending on flags every iteration of the while loop since when the task doesn't run we don't know if it's cuz the bits aren't sent or we're can blocking

If we had more time we'd implement a full watchdog that monitors every task to ensure it's executing (smth to keep in mind for next gen).

Apps/Src/ReadCarCAN.c
KnockbackNemo: Yes, no more charge enable

Apps/Src/SendTritium.c
OscarPortillo37: Oops, I must've forgotten it when I was editing it, thanks!

Apps/Src/SendTritium.c
OscarPortillo37: I'll just edit IOState to send brakelight off of what SendTritium has as its current brake status. Brakelight stuff is being sent over CAN through IOState, right?

Apps/Src/ReadCarCAN.c
OscarPortillo37: What PR is that in? I've merged everything I can so far & don't see that

Apps/Inc/IOState.h
```diff
@@ -13,6 +13,9 @@
 #define SWITCH_BITMAP_IGN_2_MOTOR(value)      ((value & 0x01) << 1) // Bit 1: IGN_2_Motor
 #define SWITCH_BITMAP_IGN_1_ARRAY(value)      ((value & 0x01) << 0) // Bit 0: IGN_1_Array
 
+#define CANBUS_MOTOR_SAFE_TO_RUN 1
+#define CANBUS_MOTOR_NOT_SAFE_TO_RUN !CANBUS_MOTOR_SAFE_TO_RUN
```
IshDeshpa: put parentheses around this macro so that order of operations doesn't accidentally screw us

Apps/Inc/SendTritium.h
```diff
@@ -11,90 +11,59 @@
...
-    RECORD_VELOCITY,
-    POWERED_CRUISE,
-    COASTING_CRUISE,
-    BRAKE_STATE,
-    ONEPEDAL,
-    ACCELERATE_CRUISE
-} TritiumStateName_t;
+// Motor Controller current values. Current is in Amps (A)
+#define MAX_MOCO_BATTERY_CURRENT 64.0f  // NOTE: Provided only for reference. This 64A max for daybreak, anticipated to be 135 for next-gen
+#define CONT_MOCO_BATTERY_CURRENT 30.0f // Continuous
```
IshDeshpa, line 36: put the comments above the macro; not sure if they will accidentally get copied with the macro or not

Apps/Inc/Tasks.h
```diff
@@ -79,17 +96,36 @@ typedef uint16_t error_code_t;
...
+#undef ERR_MSG_OFFSET
 
 extern const char *DISP_ERRMSG_NA;
 extern const char *DISP_EVACMSG_DEFAULT;
 extern const char *DISP_EVACMSG_REQ;
 
+/**
+ * @brief Can be used by tasks to add a delay in ms
+ */
+inline void delay_ms(uint32_t ms) {
```
IshDeshpa, line 162: nice

Drivers/Src/Contactors.c
```diff
@@ -76,8 +76,8 @@ void Contactors_Init() {
 bool Contactors_Get(contactor_t contactor, bool blocking) {
     switch (contactor) {
         case MOTOR_CONTROLLER_CONTACTOR:
-            contactorState[MOTOR_CONTROLLER_CONTACTOR] = BSP_GPIO_Read_Pin(MOTOR_C_SENSE_PORT, MOTOR_C_SENSE) == 0 
-                                                         ? ON : OFF;
+            // contactorState[MOTOR_CONTROLLER_CONTACTOR] = BSP_GPIO_Read_Pin(MOTOR_C_SENSE_PORT, MOTOR_C_SENSE) == 0
```
IshDeshpa, line 79: why

Drivers/Src/Dashboard.c
```diff
@@ -1,19 +1,60 @@
...
+        delay_ms(NEUTRAL_DEBOUNCE_DLY_MS);
+    }
+
+    // Until we've actually seen neutral, manually override to neutral
+    if (neutralReset) {
+        // only clear neutralReset once we see the switches truly in neutral
+        if (neu) {
+            neutralReset = false;
+        }
+        return DASH_NEU;
```
IshDeshpa, line 34: i don't think you need to return this here as it will be handled by the statement below

Apps/Src/ReadTritium.c
```diff
@@ -173,6 +178,8 @@ void assertTritiumError(tritium_error_code_t motor_err)
 	if (motor_err != T_HALL_SENSOR_ERR && motor_err != T_MOTOR_WATCHDOG_TRIP)
 	{
 		// Assert a nonrecoverable error with no callback function- nonrecoverable will kill the motor and infinite loop
+		// OSFlagPost(&BPS_Motor_Status_Flags, MOTOR_SAFE_TO_RUN, OS_OPT_POST_FLAG_CLR, &err);
```
IshDeshpa: Why no more flag posts

IshDeshpa (approved):

Apps/Inc/Tasks.h
```diff
@@ -79,17 +96,36 @@ typedef uint16_t error_code_t;
...
     arr[0] = '\"';
-    memcpy(arr + 1, prefix, 4);
-    memcpy(arr + 4, "_ERR_", 6);
-    snprintf(arr + ERR_MSG_OFFSET, 5, "%04X", err & 0xFFFF);
-    arr[ERR_MSG_OFFSET + 4] = '\"';
-    arr[ERR_MSG_OFFSET + 5] = '\0';
+    memcpy(arr + 1, prefix, 9);
+    snprintf(arr + ERR_MSG_OFFSET, 6, "_%04X", err & 0xFFFF);
+    arr[ERR_MSG_OFFSET + 5] = '\"';
+    arr[ERR_MSG_OFFSET + 6] = '\0';
```
diyarajon: would it be worth making these macros and doing macro math for the numbers? i doubt we will need extra stuff in the msg. unless Akshay has changed it in his PR.

Drivers/Src/Contactors.c
Lakshay983: this is over CAN now

Apps/Src/ReadTritium.c
Lakshay983: we do it in throw task error now

Apps/Src/ReadCarCAN.c
```diff
@@ -18,9 +18,11 @@
...
 
-#define BPS_CAN_WATCHDOG
+// #define BPS_CAN_WATCHDOG
 // #define PRECHARGE_CAN_WATCHDOG
 
+#define MOTOR_PRECHARGE_ON_COUNT_THRESHOLD 5
+
 
 // Timer delay constants
 #define CAN_WATCH_TMR_DLY_MS 1000u                                                             // 500 ms
```
OscarPortillo37: nitpick, but is this supposed to be 500ms, or is the comment outdated?

Apps/Src/ReadCarCAN.c
```diff
@@ -84,35 +97,45 @@ static bool check_MotorControllerContactor(void){
...
+    // assertOSError(err);
 }
 
+static void setMotorControllerContactor(bool state, bool blocking){
+    Contactors_Set(MOTOR_CONTROLLER_CONTACTOR, state, blocking);
+
+    // If the motor contactor is turned off, we should not be running the motor controller
+    if(state == OFF){
+        OS_ERR err;
+        OSFlagPost(&BPS_Motor_Status_Flags, MOTOR_SAFE_TO_RUN, OS_OPT_POST_FLAG_SET, &err);
```
OscarPortillo37: Should be OS_OPT_POST_FLAG_CLR, right?

Apps/Src/ReadCarCAN.c
```diff
@@ -226,19 +265,47 @@ void Task_ReadCarCAN(void *p_arg)
...
+                    // If the motor precharge contactor has been on for enough iterations, we can consider it safe to run
+                    OS_ERR err;
+                    OSFlagPost(&BPS_Motor_Status_Flags, MOTOR_SAFE_TO_RUN, OS_OPT_POST_FLAG_SET, &err);
+                    assertOSError(err);
+                }
+            }
+            else{
+                motorPrechargeOnCount = 0; // Reset the counter if both motor contactors are not on
+            }
+
```
OscarPortillo37: I realize we clear MOTOR_SAFE_TO_RUN if BPS isn't safe, but wouldn't we still wanna double check BPS_SAFE & BPS_CHECKED here (also o.w. BPS_CHECKED never get used)

OscarPortillo37:
This looks way cleaner & better, thank you Lakshay:)

Drivers/Src/Dashboard.c
OscarPortillo37: Seems to me that it is necessary; if we don't override the return value here, it could return, say, DASH_FWD if we started in FWD gear, thereby not performing the desired override into neutral when starting up the car

Apps/Src/ReadCarCAN.c
Lakshay983: nah, it's just for how many times we want to read the message before considering the motor as safe

Apps/Src/ReadCarCAN.c
Lakshay983: yup

Apps/Inc/IOState.h
```diff
@@ -13,6 +13,9 @@
 #define SWITCH_BITMAP_IGN_2_MOTOR(value)      ((value & 0x01) << 1) // Bit 1: IGN_2_Motor
 #define SWITCH_BITMAP_IGN_1_ARRAY(value)      ((value & 0x01) << 0) // Bit 0: IGN_1_Array
 
+#define CANBUS_MOTOR_SAFE_TO_RUN 1
+#define CANBUS_MOTOR_NOT_SAFE_TO_RUN (!CANBUS_MOTOR_SAFE_TO_RUN)
```
diyarajon: 🫨

Apps/Inc/SendTritium.h
```diff
@@ -11,90 +11,59 @@
...
 #define MOTOR_MSG_PERIOD 100 // in ms
 #define FSM_PERIOD 100 // in ms
 #define DEBOUNCE_PERIOD 2 // in units of FSM_PERIOD
-#define MOTOR_MSG_COUNTER_THRESHOLD (MOTOR_MSG_PERIOD)/(FSM_PERIOD)
 
-#define FOREACH_Gear(GEAR) \
-        GEAR(FORWARD_GEAR),   \
-        GEAR(NEUTRAL_GEAR),  \
-        GEAR(REVERSE_GEAR),   \
+#define MAX_VELOCITY 20000.0f // rpm (unobtainable value)
```
diyarajon, line 26: This *might* need to match the stuff on the config file. I really don't think it needs to but it might be conflicting. Nothing on the documentation about it so it's a guess that we've never tried out

Apps/Src/SendTritium.c
```diff
@@ -1,326 +1,124 @@
...
-float busCurrentSetPoint = 69.0f; // why is this float and not int?
-
-// Current observed velocity
-static float velocityObserved = 0;
-
-// Counter for sending setpoints to motor
-static uint8_t motorMsgCounter = 0;
+static float currentSetpoint = 0.0f;
+static float velocitySetpoint = 0.0f;
+static float busCurrentSetPoint = CONT_MOCO_BATTERY_CURRENT / MAX_MOCO_CURRENT; // why is this float and not int?
```
diyarajon: Does this need parenthesis in case it does weird macro math

Apps/Src/SendTritium.c
```diff
@@ -1,326 +1,124 @@
...
-    // regenEnable = ChargeEnable_Get();
-    regenEnable = false;
-
-    // Update buttons
-    // if(Minions_Read(REGEN_SW) && onePedalCounter < DEBOUNCE_PERIOD){onePedalCounter++;}
-    // else if(onePedalCounter > 0){onePedalCounter--;}
-
-    // if(Minions_Read(CRUZ_EN) && cruiseEnableCounter < DEBOUNCE_PERIOD){cruiseEnableCounter++;}
-    // else if(cruiseEnableCounter > 0){cruiseEnableCounter--;}
+    // Brake hysteresis
```
diyarajon, line 102: Do we need to add brake hysteresis? I guess it depends on what ends up happening with brake pedal

Apps/Src/SendTritium.c
diyarajon: Also if you end up doing powered run and it's hitting a weird threshold, this might be one of the issues because it's never been tested before.

diyarajon (commented):
sorry i reviewed this so late

Apps/Inc/Tasks.h
guytonde: yes, this is being removed; see #457

Apps/Inc/SendTritium.h
```diff
@@ -11,90 +11,59 @@
...
+ * Error types
+ * 
+ */
+typedef enum
+{
+    SENDTRITIUM_ERR_NONE,
+    SENDTRITIUM_ERR_GEAR_FAULT,     // Received multiple or no gear inputs (e.g. FOR_SW, REV_SW)
+} SendTritium_error_code_t;
 
 #ifdef SENDTRITIUM_EXPOSE_VARS
```
KnockbackNemo, line 56: SENDTRITIUM_EXPOSE_VARS is sort of a legacy thing and doesn't appear to compile anymore since we're externing static variables. Making the variables non-static is an option, or we could also remove the macros (both PRINT_MES and EXPOSE_VARS) entirely since there are getters and setters, and neither macro looks to be doing much in the current version

Apps/Src/SendTritium.c
KnockbackNemo: For the float vs int thing, it's probably because the motor controller expects a float between 0 and 1 for the current set point (https://docs.prohelion.com/Driver_Control/Communications_Protocol/index.html?h=float#data-field on units and %)

Apps/Src/SendTritium.c
```diff
@@ -1,326 +1,124 @@
...
-    printf("brakePedalPercent: %d\n\r", brakePedalPercent);
-    printf("accelPedalPercent: %d\n\r", accelPedalPercent);
-    printf("gear: %d\n\r", (uint8_t)gear);
-    print_float("currentSetpoint: ", currentSetpoint);
-    print_float("velocitySetpoint: ", velocitySetpoint);
-    print_float("velocityObserved: ", velocityObserved);
-    printf("-------------------\n\r");
+    UpdateDisplay_SetRegenState(DISP_DISABLED); // Not on Daybreak
+    UpdateDisplay_SetCruiseState(DISP_DISABLED); // Probably not on Daybreak
+    UpdateDisplay_SetAccel(accelPedalPercent);
```
KnockbackNemo: Is accelPedalPercent updated twice with the same info?

Apps/Src/ReadCarCAN.c
```diff
@@ -226,19 +269,48 @@ void Task_ReadCarCAN(void *p_arg)
...
+            static volatile uint8_t motorPrechargeOnCount = 0;
             #ifdef PRECHARGE_CAN_WATCHDOG
             OSTmrStart(&prechargeCanWatchTimer, &err); // Restart CAN Watchdog timer for Active Precharge Contactor msg
             assertOSError(err);
             #endif
 
-            // Update Motor Contactor sense state
-            Contactors_Set(MOTOR_CONTROLLER_CONTACTOR, MOTOR_SENSE_ACTUAL_VALUE(dataBuf.data), true);
+            // More things involved with setting the motor controller contactor, so use this function instead
+            setMotorControllerContactor(MOTOR_SENSE_ACTUAL_VALUE(dataBuf.data), true);
```
KnockbackNemo: Is the motor sense bit of the CAN message now being used to tell Controls whether or not to turn on the motor controller contactor?

Apps/Src/ReadCarCAN.c
```diff
@@ -84,35 +97,46 @@ static bool check_MotorControllerContactor(void){
...
+    // OSFlagPost(&BPS_Motor_Status_Flags, BPS_SAFE | MOTOR_SAFE_TO_RUN, OS_OPT_POST_FLAG_CLR, &err);
+    // assertOSError(err);
 }
 
+static void setMotorControllerContactor(bool state, bool blocking){
+    Contactors_Set(MOTOR_CONTROLLER_CONTACTOR, state, blocking);
+
+    // If the motor contactor is turned off, we should not be running the motor controller
+    if(state == OFF){
+        MotorStatus_ModifyBits(MOTOR_SAFE_TO_RUN, false, false);
```
KnockbackNemo: I might be being silly, but is this supposed to be clearing the flag?

Apps/Src/Tasks.c
```diff
@@ -49,9 +49,7 @@ CPU_STK DebugDump_Stk[TASK_DEBUG_DUMP_STACK_SIZE];
 CPU_STK CommandLine_Stk[TASK_COMMAND_LINE_STACK_SIZE];
 CPU_STK IOState_Stk[TASK_IO_STATE_STACK_SIZE];
 
-#define DISP_NA_STR_LITERAL "\"N/A\""
-#define DISP_EVAC_NONREQ_STR_LITERAL "\"V('u')V\""
-#define DISP_EVAC_REQ_STR_LITERAL "\"REQUIRED!!!\""
+
 
 // Controls Fault Message bits
 #define ANY_CONTROLS_FAULT_BIT 1        // 1 if any of the other bits are true
```
KnockbackNemo: Maybe we should add a fault bit for Send_Tritium and let it set the ANY_CONTROLS_FAULT_BIT as well?

Apps/Src/Tasks.c
```diff
@@ -201,7 +198,20 @@ void throwTaskError(error_code_t errorCode, callback_t errorCallback, error_sche
...
+    motormsg.data[0] |= 0x2; // Bit 1 of motor message 
+
     CANbus_Send_Faultstate(faultmsg, CARCAN);
+    CANbus_Send_Faultstate(motormsg, CARCAN);
+
+    // set motor and array positions off
+    CANDATA_t iostatemsg = {0};
+    iostatemsg.ID = IO_STATE;
+    iostatemsg.data[0] |= SWITCH_BITMAP_IGN_1_ARRAY(0);
+    iostatemsg.data[0] |= SWITCH_BITMAP_IGN_2_MOTOR(0);
```
KnockbackNemo, line 260: If we have a recoverable error where we need to reset the ignition position to off without power-cycling the board, will the ign_has_been_reset part of ignition mess with resetting the ignition to off? I think this looks okay and couldn't think of a way it would really cause an issue, but I wanted to ask to make sure.

Drivers/Src/Dashboard.c
KnockbackNemo: The logic looks good to me- it just seems to work different from before. We previously required the car to go through neutral when switching between reverse and forward, but if I'm understanding correctly, this logic lets us go between states without having a momentary neutral period in the middle. The part I'm less sure about is if we need/want to go straight from forward to reverse, though depending on the gear switch (I forgot what it looks like), we might end up going through neutral anyways if it separates the forward and reverse states.

KnockbackNemo: Also would going accidentally switching from forward to reverse while already moving quickly be a concern?

KnockbackNemo (changes requested):
Also sorry for slow review, but this is looking good and much simpler! Thanks for digging into the flags and finding ways to make the code better!

Apps/Src/Tasks.c
Lakshay983: recoverable errors are dead 🦀

Apps/Src/ReadCarCAN.c
Lakshay983: Controls doesn't turn on or off the moco now, Controls sends IGN state and precharge sets motor on. This is mostly cuz i didn't want to make the cables 😬

Apps/Src/SendTritium.c
Lakshay983: oops

Apps/Inc/Tasks.h
```diff
@@ -28,91 +25,136 @@
...
 /**
  * BPS & Motor Status Event Flag Definitions
  */
 
-#define BPS_SAFE 1 << 0
-#define BPS_CHECKED 1 << 1
-#define MOTOR_SAFE_TO_RUN 1 << 2
+#define BPS_SAFE            1 << 0
+#define BPS_CHECKED         1 << 1
+#define MOTOR_SAFE_TO_RUN   1 << 2
```
IshDeshpa, line 124: looks like someone ran a linter on this code

Apps/Src/ReadCarCAN.c
```diff
@@ -194,10 +168,12 @@ void Task_ReadCarCAN(void *p_arg) {
             }
 
             case BPS_CONTACTOR: {
-#ifdef BPS_CAN_WATCHDOG
+            #ifdef BPS_CAN_WATCHDOG
+                // Restart CAN Watchdog timer for BPS Contactor Status msg
+                OS_ERR err;
```
IshDeshpa: There's already an err variable defined above

Apps/Src/ReadCarCAN.c
```diff
@@ -265,25 +241,30 @@ void Task_ReadCarCAN(void *p_arg) {
...
-                        // OS_ERR err;
-                        // OSFlagPost(&BPS_Motor_Status_Flags, MOTOR_SAFE_TO_RUN, OS_OPT_POST_FLAG_SET, &err);
-                        // assertOSError(err);
+                        motorPrechargeOnCount = 0; // Reset the counter so we don't overflow and/or constantly post to the flag
                     }
+                    // Leave the motor safe to run bit as if it was set
+                }
+                else{
+                    MotorStatus_ModifyBits(MOTOR_SAFE_TO_RUN, false, !OS_FLAG_SCHED_POINT);
+                    motorPrechargeOnCount = 0;
```
IshDeshpa, line 266: Won't this keep resetting motorPrechargeOnCount? It's in the else condition so whenever motorPrechargeOnCount < threshold, it will be reset to 0, meaning it never makes progress

Apps/Src/SendTritium.c
```diff
@@ -198,11 +198,16 @@ void Task_SendTritium(void *p_arg) {
...
         // All bits are set
         if (err == OS_ERR_NONE) {
             // CAN message for setpoint of bus current percent
             memcpy(&powerCmd.data[4], &busCurrentSetPoint, sizeof(float));
-            CANbus_Send(powerCmd, CAN_BLOCKING, MOTORCAN);
+
+            // If we're using the profinity software don't set the power here
+            #ifndef USING_PROFINITY
+                CANbus_Send(powerCmd, CAN_BLOCKING, MOTORCAN);
+            #endif
```
IshDeshpa: you can surround the memcpy here too

Scripts/read_car_can_sim.py
```diff
@@ -26,7 +26,7 @@
 
 #com_num = input('Enter COM number: ')
 #port_name = input('Enter port name: ')
-port_name = "/dev/ttyUSB1" ### CHANGE THIS TO MATCH YOUR CANDAPTER PORT ###
+port_name = "/dev/ttyUSB0" #input('Enter port name (e.g., /dev/ttyUSB0): ')
```
IshDeshpa, line 29: why no use input here

IshDeshpa (changes requested):
image

Apps/Src/SendTritium.c
Lakshay983: me want to forward the message on carcan

Apps/Inc/Tasks.h
Lakshay983: @guytonde

